PR #27, opened 2020-05-08
Replace addEventListener with window.addEventListener

dstockwell:
Avoid a compilation error with the Closure compiler.


philipwalton:
I thought Closure Compiler could be configured to accept global `addEventListener()` via externs? It seems strange that Closure Compiler complains about `addEventListener()` but not `setTimeout()`. (I understand one is an "own" property of window and one is up the prototype chain, but still...)

Anyway, happy to merge this if that means it'll be easier for teams using Closure Compiler to integrate, but my goal is to keep the library as small as possible (and removing implied globals is an easy win, if that' workable).

dstockwell:
Ahh yes! I can add extra externs to avoid this, thanks!

PR #28, opened 2020-05-09
Send metrics as Google Tag Manager events

SimenHansen:
Most Googel Analytics implementations are in my experience working in the field done via Google Tag Manager. Added example for sending metrics via GTM, to then be picked up by Google Analytics or any other endpoint you may want to recieve data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ getLCP(sendToGoogleAnalytics);
 
 ### Send the results to Google Tag Manager
 
-The following example measures each of the Core Web Vitals metrics and sends them as seperate dataLayer-events to be used by Google Tag Manager. With the `web-vitals` trigger you send the metrics to any tag inside your account.
+The following example measures each of the Core Web Vitals metrics and sends them as seperate `dataLayer-events` to be used by Google Tag Manager. With the `web-vitals` trigger you send the metrics to any tag inside your account.
 
 ```js
 import {getCLS, getFID, getLCP} from 'web-vitals';
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@
   - [Report only the delta of changes](#report-only-the-delta-of-changes)
   - [Send the results to an analytics endpoint](#send-the-results-to-an-analytics-endpoint)
   - [Send the results to Google Analytics](#send-the-results-to-google-analytics)
+  - [Send the results to Google Tag Manager](#send-the-results-to-google-tag-manager)
   - [Load `web-vitals` from a CDN](#load-web-vitals-from-a-cdn)
 - [API](#api)
   - [Types](#types)
@@ -187,6 +188,33 @@ function sendToGoogleAnalytics({name, delta, id}) {
 getCLS(sendToGoogleAnalytics);
 getFID(sendToGoogleAnalytics);
 getLCP(sendToGoogleAnalytics);
+```
+
+### Send the results to Google Tag Manager
+
+The following example measures each of the Core Web Vitals metrics and sends them as seperate dataLayer-events to be used by Google Tag Manager. With the `web-vitals` trigger you send the metrics to any tag inside your account.
+
+```js
+import {getCLS, getFID, getLCP} from 'web-vitals';
+
+function sendToGoogleAnalytics({name, delta, id}) {
+  // Assumes the global `dataLayer` array exists, see:
+  // https://developers.google.com/tag-manager/devguide
+  dataLayer.push({
+      event: 'web-vitals',
+      event_category: 'Web Vitals',
+      event_action: name,
+      // Google Analytics metrics must be integers, so the value is rounded.
+      // For CLS the value is first multiplied by 1000 for greater precision
+      // (note: increase the multiplier for greater precision if needed).
+      event_value: Math.round(name === 'CLS' ? delta * 1000 : delta),
+      // The `id` value will be unique to the current page load. When sending
+      // multiple values from the same page (e.g. for CLS), Google Analytics can
+      // compute a total by grouping on this ID (note: requires `eventLabel` to
+      // be a dimension in your report).
+      event_label: id
+
+  });
 ```
 
 ### Load `web-vitals` from a CDN


SimenHansen:
@googlebot I signed it!

README.md
```diff
@@ -186,6 +186,32 @@ getFID(sendToGoogleAnalytics);
...
+function sendToGoogleAnalytics({name, delta, id}) {
+  // Assumes the global `dataLayer` array exists, see:
+  // https://developers.google.com/tag-manager/devguide
+  dataLayer.push({
+      event: 'web-vitals',
+      event_category: 'Web Vitals',
+      // Google Analytics metrics must be integers, so the value is rounded.
+      // For CLS the value is first multiplied by 1000 for greater precision
+      // (note: increase the multiplier for greater precision if needed).
+      event_value: Math.round(name === 'CLS' ? delta * 1000 : delta),
```
Zizzamia, line 205: Yeah, I found myself multiply by 100 in my codebase to have something like a score from 0 to 100. Have you seen many results with three decimals points?

SimenHansen: I havn't production tested it yet, so not sure. So I took this x1000 approach because it was suggested in both analytics.js and gtag.js sections of the Readme

SimenHansen: And at least when running the test-servers part of the repo I get number in the 0.00x range for CLS. 100x might be a better find in the real world environments - I have to test that out.

philipwalton: Yeah, I chose 1000x for the Google Analytics examples because I wanted to have enough granularity for several CLS beacons to be sent for the same page without introducing too much of a rounding error. If someone were to only do 100x, then the recommended threshold would be <=10, and there are only 9 integer values between 0 and 10 (which didn't feel like enough).

Also, I didn't want to give the impression that CLS was a percentage (we've have confusion around that before).

README.md
```diff
@@ -186,6 +186,32 @@ getFID(sendToGoogleAnalytics);
...
+The following example measures each of the Core Web Vitals metrics and sends them as seperate dataLayer-events to be used by Google Tag Manager. With the `web-vitals` trigger you send the metrics to any tag inside your account.
+
+```js
+import {getCLS, getFID, getLCP} from 'web-vitals';
+
+function sendToGoogleAnalytics({name, delta, id}) {
+  // Assumes the global `dataLayer` array exists, see:
+  // https://developers.google.com/tag-manager/devguide
+  dataLayer.push({
+      event: 'web-vitals',
```
philipwalton, line 200: Do you know what this value is used for? Is it sent to Google Analytics or just used internally by GTM?

README.md
```diff
@@ -186,6 +186,32 @@ getFID(sendToGoogleAnalytics);
...
 ```
 
+### Send the results to Google Tag Manager
+
+The following example measures each of the Core Web Vitals metrics and sends them as seperate dataLayer-events to be used by Google Tag Manager. With the `web-vitals` trigger you send the metrics to any tag inside your account.
+
+```js
+import {getCLS, getFID, getLCP} from 'web-vitals';
+
+function sendToGoogleAnalytics({name, delta, id}) {
```
philipwalton, line 196: I don't see the `name` value being used anywhere. For consistency with the other examples, it should be set as the event action.

README.md
```diff
@@ -186,6 +186,32 @@ getFID(sendToGoogleAnalytics);
...
+      // Google Analytics metrics must be integers, so the value is rounded.
+      // For CLS the value is first multiplied by 1000 for greater precision
+      // (note: increase the multiplier for greater precision if needed).
+      event_value: Math.round(name === 'CLS' ? delta * 1000 : delta),
+      // The `id` value will be unique to the current page load. When sending
+      // multiple values from the same page (e.g. for CLS), Google Analytics can
+      // compute a total by grouping on this ID (note: requires `eventLabel` to
+      // be a dimension in your report).
+      event_label: id
+
```
philipwalton, line 211: Do you need to set the event's non-interaction flag when using GTM? I assume you still do?

README.md
SimenHansen: This key is what you build a trigger on in GTM. The value just have to match whatever you set in the trigger on the GTM side.

README.md
SimenHansen: Yes you still do. But GTM sets this in the Tag Configuration with a dropdown. No need to send it explicit as a value.

README.md
SimenHansen: Yep, I agree. Fixed that one.

README.md
SimenHansen: Not sent anywhere unsless you want to.

README.md
```diff
@@ -189,6 +190,33 @@ getFID(sendToGoogleAnalytics);
 getLCP(sendToGoogleAnalytics);
 ```
 
+### Send the results to Google Tag Manager
+
+The following example measures each of the Core Web Vitals metrics and sends them as seperate dataLayer-events to be used by Google Tag Manager. With the `web-vitals` trigger you send the metrics to any tag inside your account.
```
philipwalton, line 195: Can you put "dataLayer-events" in backticks?

README.md
philipwalton: I see, so then the you'd set up a trigger in your analytics tag in GTM to listen for the `web-vitals` event, and then configure how you want to send the data to Google Analytics?

Do the `event_*` properties automatically map, or would a user have to set that up themselves? (Sorry, never used GTM).

SimenHansen: Let me give you a quick GTM workflow for setting this up:

1. **Create a GA Event Tag**
chrome_2020-05-10_23-01-28

2. **Map Event Tracking Parameters**
Here you need to manually map each "dataLayer"-variable seperate. So there is actually no automatic mapping here.
bilde

3. **Define and set a trigger**
Here GTM reads from the `event` key defined at the top of the dataLayer.push()-object.
chrome_2020-05-10_23-07-22

4. **Pull it all together**
chrome_2020-05-10_23-08-39

And it will look something like this. Every time a new dataLayer.push() is run, it will send an event to Google Analytics. 

You can of course set more complex triggers, say if you only want to send the events on certain pages of your website or any other criteria that you might think of. I'm thinking of running this only on product-pages for a ecommerce client of mine as a start.

README.md
SimenHansen: Sure. Done.

README.md
philipwalton: Gotcha, thanks for clarifying. Do you know of any good documentation that we could link to for someone who's a GTM user but might not be familiar with how to set this up?

SimenHansen: The official documentation is rather lacking in the particualar area. 
I linked the dev-doc in the Readme for the `dataLayer` part of the story. 

The configuration part you have [the offical Tag Manager docs](https://support.google.com/tagmanager/answer/6106716?hl=en&ref_topic=6333310l). But its probably better to link to something from [analyticsmania.com](https://www.analyticsmania.com/post/google-tag-manager-event-tracking/#section3)

philipwalton: Thanks, I'll merge this and then figure how where to link to in a separate PR.

SimenHansen: I could probably write up the walk-through comment as a gist or something, for a simple step-by-step guide, and we could link to in the docs.

philipwalton: Sounds good to me. Longer term I'd like to expand that docs to individual guides that show how to do all of these types of things. And I'd also like to open it up to contributions from other analytics providers.

If you write up a gist that'd be a good starting point for the time being.

nahtnam: Hello, I had to look up this thread to figure out how to set up GTM. Why not create a new file and slap this comment in there: https://github.com/GoogleChrome/web-vitals/pull/28#discussion_r422701126

philipwalton: Added in https://github.com/GoogleChrome/web-vitals/commit/1aa8f7a198aba63039b9c579e0a928c5dee68f23

madilk: Thanks for this, Simen. In addition to sending it to GA as Event, FID and LCP can be sent as Timing events to GA via GTM.

philipwalton: @madilk I'd recommend not sending these to GA as timing hits, as those are quite restricted (e.g. are sampled at 1%, can't be used in segments, etc.)

madilk: @philipwalton Perhaps, it might be better to send as Events and Timing hits to give the flexibility later? Usually, we set the siteSpeedSampleRate field in GA hits to be much higher % than the default ~1%. Also, I wasn't sure about the timing segment and checked it out. Timing hit values can be used in custom segments within GA. Attaching a screenshot here.
timing

philipwalton: Interesting, maybe the sampling rules for timing hits has changed since I last tried them, but regardless the 1% sampling is the biggest blocker.

philipwalton: And to clarify, the 1% sampling rate is applied on both the client and the server, so changing the 1% to 100% will not actually get you 100%. See: https://developers.google.com/analytics/devguides/collection/analyticsjs/user-timings#sampling_considerations

bertrand-riviere: hello @SimenHansen,
thanks a lot for the GTM workflow, it is very useful!
now that i can track the events in analytics, i'd like to go one step further, and get a quick visual repartition of "good"/"ok"/"poor" performances..
by any chance, would you have any idea of how this could be achieved using data studio?
let's say to produce graphs like the ones in CrUX Dashboard:
image

because having avg. values in analytics is better than nothing but does not help to get the full picture

PR #39, opened 2020-05-17
Update link anchor

roderickhsiao:
Minor change for page anchor, 

changed from `#reporting-only-the-delta-of-changes` to `#report-only-the-delta-of-changes`


philipwalton:
Thanks!

PR #54, opened 2020-06-17
docs(readme): integrations section

adamhenson:
# Summary

This change introduces an "integrations" section of the documentation. Welcoming feedback or change requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,7 +253,7 @@ addEventListener('DOMContentLoaded', function() {
 
 _**Note:** it's safe to use module scripts in legacy browsers because unknown script types are ignored._
 
-#### Integrations
+### Integrations
 
 - [Automated Lighthouse Check](https://www.automated-lighthouse-check.com/web-vitals) provides a REST API endpoint for collecting Web Vitals data and a web dashboard for historic visualization. You can filter metrics by browser, device and OS.
 
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
   - [Send the results to Google Analytics](#send-the-results-to-google-analytics)
   - [Send the results to Google Tag Manager](#send-the-results-to-google-tag-manager)
   - [Load `web-vitals` from a CDN](#load-web-vitals-from-a-cdn)
-  - [Integrations](#integrations)
+- [Integrations](#integrations)
 - [API](#api)
   - [Types](#types)
   - [Functions](#functions)
@@ -253,7 +253,7 @@ addEventListener('DOMContentLoaded', function() {
 
 _**Note:** it's safe to use module scripts in legacy browsers because unknown script types are ignored._
 
-### Integrations
+## Integrations
 
 - [Automated Lighthouse Check](https://www.automated-lighthouse-check.com/web-vitals) provides a REST API endpoint for collecting Web Vitals data and a web dashboard for historic visualization. You can filter metrics by browser, device and OS.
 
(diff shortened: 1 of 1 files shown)


philipwalton:
@adamhenson thanks for the suggestion, but I'm a bit hesitant to add a PR for an integrations section because there are already a lot of integrations with Web Vitals and it would be hard to capture them all in a single README.

Regarding this specific tool, I do have a couple of suggestions though:

- It's called "Automated Lighthouse Check" but (at least upon initial inspection) it doesn't seem to integrate with Lighthouse at all. Keep in mind that Lighthouse is a lab tool and this library tracks Web Vitals metrics in the field. We don't want to imply to developers that these terms our synonymous.
- The code sample you provide on [this page](https://www.foo.software/web-vitals-how-to-track-and-analyze-web-vitals/) checks to see if the metric is final before sending it. That's not a good idea as you'll likely lose a lot of CLS data if you wait until the metric is final (you should be reporting it any time the visibility state changes to hidden).
- The example uses `fetch()`, but it'd be better to use `sendBeacon()` (or `fetch()` with the [keep-alive](https://www.chromestatus.com/features/5760375567941632) flag). Using fetch as is might block other network activity.

adamhenson:
@philipwalton thanks for the feedback in the 2nd and 3rd bullet. The first bullet is quite incorrect however. In fact Automated Lighthouse Check is listed as an integration in the Lighthouse GitHub project. I thought it would make sense to mirror it in this project (by having an “integrations” section). I’m disappointed in your decision to dismiss this PR. I understand your perspective, but by having this type of section it adds a small bit of support for all the dedicated developers that are supporting you. It would be a nice way to give back. As an engineer representing Google I’d recommend being a bit more thoughtful to support the community with decisions like these.

paulirish:
(chiming in just as a watcher of this repo)

FWIW I agree that it doesn't make sense to add an integrations section to this project's readme.  IMO that kinda content isn't too typical for OSS projects, we just happen to do it in the lighthouse readme.   

And let's be real, adding a link to your site **_above_** the API documentation is a ballsy move.

>  it adds a small bit of support for all the dedicated developers that are supporting you. It would be a nice way to give back

speaking in general as an open source maintainer.. i think there's a big distinction between a project's users and its supporters. i wouldn't say most users of my projects are "supporting" the project. and as the projects grow large, it gets considerably harder to scale one's time and attention to 1) deal with all filed issues and 2) do so thoughtfully.  

IMO philip's response was hugely thoughtful, even pointing out two suggestions. And he's fully justified in this decision.

adamhenson:
Fair enough, I appreciate having more than one opinion. I wasn't sure where to place the integrations section but was hoping to get feedback about it. I understand the perspectives here and can empathize. I am deeply familiar and respect your work... going way back. Hope this didn't come across as unappreciative. And I'm content knowing that you at least entertained my perspective and that we had more than one core contributor opinion on this. I'm going to sit down now... before I'm banished from the JS community 😆

adamhenson:
I just meant "support" in the sense that I'm a fan and follower and one of the zillions of developers that click that star. I'm one of the developers paying up the big bucks to go to the next conference you speak at. I'm the one linking to your projects driving traffic to your work... to build your brand and Google's. I'm the one talking about that article you just wrote. I was fishing for some fan appreciation, but I get it... I understand where ya'll stand.

PR #84, opened 2020-11-10
[WIP] Release v1

### LINKED ISSUE #29: Is CLS actually DCLS?
Judging by the implementation, it looks like the "CLS" that web-vitals reports is actually [DCLS](https://github.com/WICG/layout-instability#cumulative-scores). Is that right?

### LINKED ISSUE #55: Feature Request: Add `removeEventListener`/`unobserve` equivalent
Currently every invocation creates a new PerformanceObserver that never `unobserve`s so there's no way to remove a handler. It'd be nice to be able to remove old listeners.

EDIT: This affects the web vitals extension which leaks listeners on every activation of tab. I'm fixing over there, but this seems generally useful too 😃

### LINKED ISSUE #75: Watching for 'scroll' event may end LCP before a valid browser entry is created
**TLDR; I believe `web-vitals` marks `isFinal` on all types of scroll events when only some scroll events stop `largest-contentful-paint` browser entries.**

### Background
Thanks for open sourcing this! 

When we started trying to use `getLCP` in our code at page start, the reporter would never fire. We noticed that our component library's positioning changes caused a scroll event before any meaningful render, that the web-vitals' input handler [would catch](https://github.com/GoogleChrome/web-vitals/blob/56c736b7c4e80f295bc8a98017671c95231fa225/src/lib/whenInput.ts#L22). The page hadn't rendered anything, so there weren't any LCP entries for web-vitals to report to our callback. All well and fine, we were causing a scroll event, and the docs describe marking `isFinal` on scroll. 

However, in our case, the Web Chrome Vitals Extension later displayed an LCP value. The extension would call `getLCP` on [tab-complete](https://github.com/GoogleChrome/web-vitals-extension/blob/fc7c50d079ef41db229057391bcde214664d6cd0/src/bg/background.js#L62), well after the scroll that stopped our web-vitals instance and read LCP entries just fine. 

Let me know if you want a repro. The code is still internal, so it may take some time to get a fiddle up.

I get the impression this library watches for scroll because it reads entries created by the browser, and the browser stops recording on scroll. Through my use case, though, I think I've found that this library stops on _any_ scroll event when the browser only stops creating the `largest-contentful-paint` entries on _some_ scroll events.

### LCP logic in Chromium
Searching for LCP and scroll logic in Chromium, I think I might have found the corresponding files. 

Many types of scrolls I think all create `Event.type = 'scroll':
https://github.com/chromium/chromium/blob/a4f27fdc4913d800752bd1cb9533d31357163d65/third_party/blink/renderer/core/scroll/scroll_types.h#L58

But LCP only stops recording on two types of scroll events:
https://github.com/chromium/chromium/blob/55f44515cd0b9e7739b434d1c62f4b7e321cd530/third_party/blink/renderer/core/paint/paint_timing_detector.cc#L188

### Remediation?
I admittedly don't understand what `kCompositorScroll` is, but the library could listen to key up, key down, and wheel device events instead of `scroll` as a proxy for 'kUserScroll'. 

If this is a valid bug and has a straightforward fix, let me know, I'm happy to open contribute.

### LINKED ISSUE #82: The isFinal flag is confusing
The initial intent of the `isFinal` flag was allow consumers of this library to know when LCP was finalized (since the [Largest Contentful Paint API](https://wicg.github.io/largest-contentful-paint/) does not currently expose this information). However, given #75, the heuristics will become significantly less accurate if we're not tracking scrolls.

The `isFinal` flag was also a bit weird because it only really applied to LCP. CLS is not final until the page has unloaded, and FCP, FID, and TTFB are always final after the first entry is dispatched.

I've also seen examples in the wild where developers were misunderstanding `isFinal` and waiting until it was `true` before sending the data to their analytics provider. That is definitely *not* what `isFinal` is for, and something needs to be done to avoid that confusion.

### LINKED ISSUE #83: The metrics should account for bfcache restores to match CrUX
Now that [back/forward cache](https://www.chromestatus.com/feature/5815270035685376) (bfcache) is supported in Chrome in 86, the Chrome User Experience report (CrUX) is going to start including bfcache versions of these metrics in its dataset.

To match CrUX, this library should be updated to report the relevant metrics on bfcache restores.

philipwalton:
The branch in this PR will be used to track changes for the [v1 release](https://github.com/GoogleChrome/web-vitals/milestone/1).

Fixes #29, fixes #55, fixes #75, fixes #82, #fixes #83


README.md
```diff
@@ -252,6 +263,86 @@ addEventListener('DOMContentLoaded', function() {
...
+### Which bundle is right for you?
+
+Most developers will generally want to use the "full" bundle (either the ES module or UMD version, depending on your build system), as it's the easiest to use out of the box and integrate into existing build tools.
+
+However, there are a few good reasons to consider using the "base" version along with the `polyfill.js` script. For example:
+
+- FID can be measured in all browsers.
+- FCP, FID, and LCP will be more accurate in some cases (since the polyfill detects the page's initial `visibilityState` earlier).
+
+Also, the minimal polyfill to support back/forward cache restores that is found in the "full" version is largely the same as the code used in the `polyfill.js` script—just split out. This means that using the polyfill gets you wider browser support and more accurate results with almost no increased code cost (just increased implementation complexity).
```
housseindjirdeh, line 321: This confused me a bit :)

There's a built-in polyfill in the "full" version that is pretty much the same as _polyfill.js_, but using "base" + _polyfill.js_ provides better browser support and accuracy? Maybe add a one-liner to explain why?

README.md
```diff
@@ -252,6 +263,86 @@ addEventListener('DOMContentLoaded', function() {
...
+
+```diff
+- import {getLCP, getFID, getCLS} from 'web-vitals';
++ import {getLCP, getFID, getCLS} from 'web-vitals/base';
+```
+
+All other usage instructions (as well as the public API) are the same in both versions.
+
+_**Note:** while it's certainly possible to copy and paste the code in `polyfill.js` directly into your templates (for step #1 above), it's better to automate this within your build process—otherwise you risk the polyfill and base scripts getting out of sync when new versions are released._
+
```
housseindjirdeh, line 345: I wonder if this section could/should be moved up to the `Usage` section? For example, something like:

----

### Usage

There are two ways to use the `web-vitals` library on your site.

**1. Use the "full" version**

```js
import {getLCP, getFID, getCLS} from 'web-vitals';

...
```

**2. Use the "base" version with `polyfill.js`**

In your application code, import the "base" build rather than the "full" build. To do this, change any `import` statements to reference `web-vitals/base` rather than `web-vitals`:

```diff
- import {getLCP, getFID, getCLS} from 'web-vitals';
+ import {getLCP, getFID, getCLS} from 'web-vitals/base';
```

Then inline the code from `dist/polyfill.js` into the `<head>` of your pages. 

```html
<!DOCTYPE html>
<html>
  <head>
    <script>
      // Inline code from `dist/polyfill.js` here
    </script>
  </head>
  <body>
    ...
  </body>
</html>
```

See <a href="#which-bundle-is-right-for-you">Which bundle is right for you?</a> for the differences between the two options.

housseindjirdeh: Feel free to ignore of course if you would rather not move the section up

README.md
housseindjirdeh: Now that I understand, I think the term "full" is what makes it confusing. It makes it seem like it has everything (back-forward cache restores **and** FID polyfill)

Changing it to something else should make things clearer 🚀

README.md
```diff
@@ -239,18 +317,101 @@ addEventListener('DOMContentLoaded', function() {
...
+      See <a href="#how-to-use-the-polyfill">how to use the polyfill</a> for more details.
+    </td>
+  </tr>
+</table>
+
+### Which bundle is right for you?
+
+Most developers will generally want to use the "standard" bundle (either the ES module or UMD version, depending on your build system), as it's the easiest to use out of the box and integrate into existing build tools.
+
+However, there are a few good reasons to consider using the "polyfill" version, for example:
```
housseindjirdeh, line 403:
```suggestion
However, there are a few good reasons to consider using the "base+polyfill" version, for example:
```

PR #86, opened 2020-11-10
Remove the isFinal flag from the Metric interface

### LINKED ISSUE #82: The isFinal flag is confusing

philipwalton:
This PR fixes #82 by removing the `isFinal` flag from the `Metric` interface.

The primary implication of this change is now when calling any of the metric functions with the `reportAllChanges` argument set to `true`, the callback will _only_ be invoked if the metric value changes—i.e. the callback won't be invoked when the page is backgrounded or unloading, nor (in the case of `getLCP()`) will the callback be invoked after user input.

Prior to this change a callback could be invoked without its value changing, but the `isFinal` flag would flip from `false` to `true`.

cc: @addyosmani as this will affect the Web Vitals Extension.

### DIFF
--- a/src/getCLS.ts
+++ b/src/getCLS.ts
@@ -43,14 +43,10 @@ export const getCLS = (onReport: ReportHandler, reportAllChanges = false) => {
 
   const po = observe('layout-shift', entryHandler as PerformanceEntryHandler);
   if (po) {
-    report = bindReporter(onReport, metric, po, reportAllChanges);
+    report = bindReporter(onReport, metric, reportAllChanges);
 
-    onHidden(({isUnloading}) => {
+    onHidden(() => {
       po.takeRecords().map(entryHandler as PerformanceEntryHandler);
-
-      if (isUnloading) {
-        metric.isFinal = true;
-      }
       report();
     });
   }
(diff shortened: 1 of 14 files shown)


rviscomi:
Is this a breaking change, and if so are you planning to release it with a new major version number?

philipwalton:
Yeah, this is a breaking change. Sorry if it's not clear from this PR, but the linked issue is part of the [v1 milestone](https://github.com/GoogleChrome/web-vitals/milestone/1).

PR #92, opened 2020-11-20
Add TTLB

pan-alexey:
Knowing the value of TTFB and TTLB, we can get another important metric - the time it takes for the server to send the entire content of the resource requested by the client. That is the difference between TTLB and TTFB.

### DIFF
--- a/src/getTTLB.ts
+++ b/src/getTTLB.ts
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


pan-alexey:
@googlebot I fixed it.

pan-alexey:
@googlebot I fixed it.

pan-alexey:
@googlebot I fixed it.

philipwalton:
Thanks for the suggestion, but the intent of this library is not to include every performance metric that might be useful, it's to include all the field-measurable Web Vitals metrics. Time-to-last-byte is not a Web Vitals metric, so I don't think it makes sense to include it.

If you do want to track TTLB, there's [an example in the README](https://github.com/GoogleChrome/web-vitals#getttfb) of how to use the `entries` object passed to the `getTTFB()` callback, which—while perhaps less intuitive than a dedicated function—avoids the duplication of refetching the navigation entry, so that's what I'd recommend. (It's also [what I do](https://github.com/philipwalton/blog/blob/v3.1.0/assets/javascript/log.js#L266-L286) to track workerStart, responseEnd, DCL, and load on my blog).

PR #110, opened 2021-01-06
Add support for Set in place of WeakSet

### LINKED ISSUE #104: Internet Explorer 11 Support - WeakSet is undefined
Hi,

According to the README file, this library should support IE9+ without any error.
I just updated to v1.0.1 and I see that you are using WeakSet which is not supported in old browsers.

https://github.com/GoogleChrome/web-vitals/blob/bebca81aa66041ffdfacb5ae65990d1acede07f9/src/lib/finalMetrics.ts

omrilotan:
Fixes #104 

IE11 does not support WeakSet but does support Set. It is still maintained until Aug 2021.
Since the Weakset is created on import, even when the module is loaded and not executed a breakage is caused.

A plain Set implements the necessary API required by this module.

Error: `ReferenceError: 'WeakSet' is undefined`
User agent string: `Mozilla/5.0 (Windows NT 10.0; WOW64; Trident/7.0; rv:11.0) like Gecko` (IE11, Win10)

### DIFF
--- a/src/lib/finalMetrics.ts
+++ b/src/lib/finalMetrics.ts
@@ -16,5 +16,6 @@
 
 import {Metric} from '../types.js';
 
-
-export const finalMetrics: WeakSet<Metric> = new WeakSet();
+export const finalMetrics: WeakSet<Metric>|Set<Metric> = 'WeakSet' in window
+  ? new WeakSet()
+  : new Set();
--- a/src/lib/finalMetrics.ts
+++ b/src/lib/finalMetrics.ts
@@ -16,6 +16,6 @@
 
 import {Metric} from '../types.js';
 
-export const finalMetrics: WeakSet<Metric>|Set<Metric> = 'WeakSet' in window
+export const finalMetrics: WeakSet<Metric>|Set<Metric> = typeof WeakSet === 'function'
   ? new WeakSet()
   : new Set();


omrilotan:
@googlebot I signed it!

src/lib/finalMetrics.ts
```diff
@@ -16,5 +16,6 @@
 
 import {Metric} from '../types.js';
 
-
-export const finalMetrics: WeakSet<Metric> = new WeakSet();
+export const finalMetrics: WeakSet<Metric>|Set<Metric> = 'WeakSet' in window
```
philipwalton, line 19: Can you update this to be:

```js
export const finalMetrics: WeakSet<Metric> | Set<Metric> =
    typeof 'WeakSet' !== undefined ? new WeakSet() : new Set();
```

For folks using this library on the server for SSR, `window` may not be defined.

omrilotan: Changed to `typeof WeakSet` (which will result in either "function"  or "undefined")

philipwalton:
Thanks!

Leo555:
IE10  does not support Set。

omrilotan:
> IE10 does not support Set。

IE10 support has ended on January 31, 2020

PR #111, opened 2021-01-11
tsetse conformance

dstockwell:
* missing return values (map vs forEach)


philipwalton:
Thanks, I'd originally used `map` instead of `forEach` to save a few bytes, but happy to make this change if it's causing tools to complain.

philipwalton:
Is a new release required here, or is pulling from the main branch sufficient?

dstockwell:
> Is a new release required here, or is pulling from the main branch sufficient?

Thanks, I can pull from the main branch for now.

PR #112, opened 2021-01-12
Move value below label for examples in README

ben-larson:
Hey!  The order of these in the docs tripped me up.  In GTM the label field comes before the value field.  It's a small thing but I think this makes it a little easier to implement.  🙂 

image


philipwalton:
I'm fine with this change, though I'd prefer to have it be consistent across all examples. Can you make that update as well?

ben-larson:
@philipwalton Thanks, done!  I think I got all the examples.

philipwalton (approved):

philipwalton:
Thanks!

Issue #193, opened 2021-11-09
Disconnect performance observer if the page wasn't hidden prior to the update


monis0395:
I was going through the code and observed that for few of the metrics we `Only report if the page wasn't hidden prior to the metric.`
- [FID](https://github.com/GoogleChrome/web-vitals/blob/main/src/getFID.ts#L34)
- [FCP](https://github.com/GoogleChrome/web-vitals/blob/main/src/getFCP.ts#L37)
- [LCP](https://github.com/GoogleChrome/web-vitals/blob/main/src/getLCP.ts#L40)

Now, if the page is ever hidden and the metric is not yet reported, then we would never update the values for any of the above metrics because `visibilityWatcher.firstHiddenTime` would always be lower because it tracks `hidden` time.

Thus I was thinking (CMIIW) if a page is hidden, we can disconnect performance observer for FID & LCP. We are already doing it for [FCP](https://github.com/GoogleChrome/web-vitals/blob/main/src/getFCP.ts#L33) for a different reason so no change there.

Also is there a reason for [always reporting LCP ](https://github.com/GoogleChrome/web-vitals/blob/main/src/getLCP.ts#L45) i.e. even if the value remains [unchanged](https://github.com/GoogleChrome/web-vitals/blob/main/src/getLCP.ts#L40) ?

philipwalton:
> Now, if the page is ever hidden and the metric is not yet reported, then we would never update the values for any of the above metrics because `visibilityWatcher.firstHiddenTime` would always be lower because it tracks `hidden` time.
> 
> Thus I was thinking (CMIIW) if a page is hidden, we can disconnect performance observer for FID & LCP. We are already doing it for [FCP](https://github.com/GoogleChrome/web-vitals/blob/main/src/getFCP.ts#L33) for a different reason so no change there.

You're correct that we could safely disconnect the observer in cases where the page was hidden prior to the relevant event.

I originally considered this and decided against it due to the fact that it would increase code size and I was trying to keep the library as small as possible. I also talked to the engineering teams about the performance impact of keeping an observer open without disconnecting it, and they said it was likely negligible.

That being said, I'd be open to PRs to add this if it doesn't increase the code size too much.

[...]

PR #196, opened 2021-12-10
Disconnect performance observer when page is already hidden

monis0395:
Added a check in onHidden.ts, to check if page is already hidden when onHidden() is called. This would allow listeners like FID, LCP to disconnect performance observer whenever the page is hidden

### DIFF
--- a/src/lib/onHidden.ts
+++ b/src/lib/onHidden.ts
@@ -14,10 +14,9 @@
  * limitations under the License.
  */
 
+type Event = { type?: string, timeStamp?: number };
 
-export interface OnHiddenCallback {
-  (event: Event): void;
-}
+export type OnHiddenCallback = (event: Event) => void;
 
 
 export const onHidden = (cb: OnHiddenCallback, once?: boolean) => {
@@ -34,4 +33,7 @@ export const onHidden = (cb: OnHiddenCallback, once?: boolean) => {
   // Some browsers have buggy implementations of visibilitychange,
   // so we use pagehide in addition, just to be safe.
   addEventListener('pagehide', onHiddenOrPageHide, true);
+
+  // if the document is already hidden
+  onHiddenOrPageHide({});
 };
--- a/src/lib/getVisibilityWatcher.ts
+++ b/src/lib/getVisibilityWatcher.ts
@@ -26,7 +26,7 @@ const initHiddenTime = () => {
 const trackChanges = () => {
   // Update the time if/when the document becomes hidden.
   onHidden(({timeStamp}) => {
-    firstHiddenTime = timeStamp
+    firstHiddenTime = timeStamp || initHiddenTime();
   }, true);
 };
 


monis0395:
This will resolve part 1 of https://github.com/GoogleChrome/web-vitals/issues/193

vanderhoop:
Any movement on this? It would seem this change would avoid unrepresentative metrics being reported.

monis0395:
@philipwalton can u please review this PR ?

if @vanderhoop is talking about the case where page was already hidden when the module loaded then become visible `getLCP` would be reporting false metrics when `reportAllChanges` is `true`?

This PR:#196 or PR:#197 should fix that!

src/lib/onHidden.ts
```diff
@@ -34,4 +33,7 @@ export const onHidden = (cb: OnHiddenCallback, once?: boolean) => {
   // Some browsers have buggy implementations of visibilitychange,
   // so we use pagehide in addition, just to be safe.
   addEventListener('pagehide', onHiddenOrPageHide, true);
+
+  // if the document is already hidden
+  onHiddenOrPageHide({});
```
philipwalton, line 38: Adding this will cause problem for other functions that use `onHidden`. We do not want to call the callback immediately—only when the page is actually hidden.

I see that your changes to the `visibilityWatcher.ts` file account for calling it right away, but other modules import this function as well, and it won't work in many of those cases.

philipwalton:
Sorry for the late reply, @monis0395. I took a look at your proposed changes, and (unless I'm missing something), I don't think this addresses the original issue in #193.

The point you brought up there was that Performance Observer objects might be created in situation where they don't have to be (because the page is already hidden). In those cases I think the best solution is to not create them in the first place rather than creating them and immediately disconnecting them.

philipwalton:
> Any movement on this? It would seem this change would avoid unrepresentative metrics being reported.

@vanderhoop this PR and the linked issue do not change what data the library reports, it only changes whether or not a Performance Observer object is created. Please file an issue if you're seeing cases where LCP is reported for pages loaded in the background.

vanderhoop:
> @vanderhoop this PR and the linked issue do not change what data the library reports, it only changes whether or not a Performance Observer object is created. Please file an issue if you're seeing cases where LCP is reported for pages loaded in the background.

Hey @philipwalton, thanks for the quick response 🙏. My coworker, @dvndrsn, has cooked up a great reproduction video for the odd behavior we're seeing for LCP reporting for tabs that have been backgrounded. I'll work with him to scrub it and submit an issue.

monis0395:
> The point you brought up there was that Performance Observer objects might be created in situation where they don't have to be (because the page is already hidden). 

Yes! You are correct about this

> In those cases I think the best solution is to not create them in the first place rather than creating them and immediately disconnecting them.

I am immediately disconnecting Performance Observer(`po`) because `onBFCacheRestore` is only attached when `po` is present. Thus I thought of immediately disconnecting them. I am not fully aware when should we attach `onBFCacheRestore`

If u have any approaches, I'll be glad to implement them

PS: https://github.com/GoogleChrome/web-vitals/pull/197#issuecomment-1007002482 this was quite informative. 
I did not know that when the document visibility changes from hidden -> visible, no LCP value would be set on the metric

PPS: sorry for the late reply

Kcwonder197 (approved):

philipwalton:
Closing as this is no longer doable since the code has changed since this PR was filed.

PR #197, opened 2021-12-10
Report LCP only if the value changes

monis0395:
Report LCP only if the value changes. To keep things consistent with other metrics


monis0395:
This will resolve part 2 of https://github.com/GoogleChrome/web-vitals/issues/193

philipwalton (approved):

philipwalton:
Thanks! Note to anyone reading this that this change doesn't affect the behavior of the library or what will be reported.

Before this change, when the `report()` function was called unconditionally it would not actually invoke the `onReport` callback in cases where the document was hidden because no LCP value would be set on the metric. This change just makes the code for LCP consistent with other metrics (e.g. FCP and FID).

PR #221, opened 2022-04-25
Finish adding getINP()

### LINKED ISSUE #208: Request for support: responsiveness metric
A note to remember that if the test is conclusive, it could be relevant to add to this library the support of the [responsivement metric](https://web.dev/responsiveness/).

philipwalton:
Fixes #208.

This PR builds on top of what @mmocny added in #213 with the following:

- Updates the `observe` module logic to support processing a list of entries rather than each individually.
- Moves the `interactionCount` logic into its own module and rewrites it as a "polyfill" so it won't run if the browser natively supports `interactionCount` (hopefully coming to Chrome soon).
- Sets the `Metric.entries` value to all reported entries from the INP interaction, rather than just the longest. (As discussed we should extend this to all entries in the frame, but we probably need a dedicated API for this first.)
- Updates the `durationThreshold` to 50 as a compromise between performance and more accurate duration values at lower percentiles.
- Adds bfcache support.
- Adds tests

### DIFF
--- a/src/getINP.ts
+++ b/src/getINP.ts
@@ -32,13 +32,25 @@ interface Interaction {
 // interaction latencies should only consider the current navigation.
 let prevInteractionCount = 0;
 
+/**
+ * Returns the interaction count since the last bfcache restore (or for the
+ * full page lifecycle if there were no bfcache restores).
+ */
+const getInteractionCountForNavigation = () => {
+  return getInteractionCount() - prevInteractionCount;
+}
+
 // To prevent unnecessary memory usage on pages with lots of interactions,
 // store at most 10 of the longest interactions to consider as INP candidates.
 const MAX_INTERACTIONS_TO_CONSIDER = 10;
 
 // A list of longest interactions on the page (by latency) sorted so the
 // longest one is first. The list is as most MAX_INTERACTIONS_TO_CONSIDER long.
-let longestInteractions: Interaction[] = [];
+let longestInteractionList: Interaction[] = [];
+
+// A mapping of longest interactions by their interaction ID.
+// This is used for faster lookup.
+const longestInteractionMap: {[interactionId: string]: Interaction} = {};
 
 /**
  * Takes a performance entry and adds it to the list of worst interactions
@@ -47,29 +59,37 @@ let longestInteractions: Interaction[] = [];
  * and entries list is updated as needed.
  */
 const processEntry = (entry: PerformanceEventTiming) => {
-  // Only process the entry if it's possibly one of the ten longest.
-  if (longestInteractions.length < MAX_INTERACTIONS_TO_CONSIDER ||
-      entry.duration > longestInteractions[longestInteractions.length - 1].latency) {
+  // The least-long of the 10 longest interactions.
+  const minLongestInteraction =
+      longestInteractionList[longestInteractionList.length - 1]
 
-    const existingInteractionIndex =
-        longestInteractions.findIndex((i) => entry.interactionId == i.id);
+  const existingInteraction = longestInteractionMap[entry.interactionId!];
 
+  // Only process the entry if it's possibly one of the ten longest,
+  // or if it's part of an existing interaction.
+  if (existingInteraction ||
+      longestInteractionList.length < MAX_INTERACTIONS_TO_CONSIDER ||
+      entry.duration > minLongestInteraction.latency) {
     // If the interaction already exists, update it. Otherwise create one.
-    if (existingInteractionIndex >= 0) {
-      const interaction = longestInteractions[existingInteractionIndex];
-      interaction.latency = Math.max(interaction.latency, entry.duration);
-      interaction.entries.push(entry);
+    if (existingInteraction) {
+      existingInteraction.entries.push(entry);
+      existingInteraction.latency =
+          Math.max(existingInteraction.latency, entry.duration);
     } else {
-      longestInteractions.push({
+      const interaction = {
         id: entry.interactionId!,
         latency: entry.duration,
         entries: [entry],
-      });
+      }
+      longestInteractionMap[interaction.id] = interaction;
+      longestInteractionList.push(interaction);
     }
 
     // Sort the entries by latency (descending) and keep only the top ten.
-    longestInteractions.sort((a, b) => b.latency - a.latency);
-    longestInteractions.splice(MAX_INTERACTIONS_TO_CONSIDER);
+    longestInteractionList.sort((a, b) => b.latency - a.latency);
+    longestInteractionList.splice(MAX_INTERACTIONS_TO_CONSIDER).forEach((i) => {
+      delete longestInteractionMap[i.id];
+    });
   }
 }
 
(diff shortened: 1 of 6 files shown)


src/getINP.ts
```diff
@@ -14,118 +14,127 @@
...
+      report();
+    }
   };
 
-  const po = observe('event', entryHandler as PerformanceEntryHandler);
+  const po = observe('event', handleEntries, {
+    // The use of 50 here as a balance between not wanting the callback to
+    // run for too many already fast events (one frame or less), but also
+    // get enough fidelity below the recommended "good" threshold of 200.
+    durationThreshold: 50,
```
mmocny, line 114: Event Timing is rounded to the nearest 8ms.  That makes the duration a nice round number (either 48ms or 56ms). 
 That means that this 50ms durationThreshold is effectively a 56ms duration threshold, which we may want to be explicit about.

TL;DR I think this is a good threshold.  But here are some nuances:

* At 60hz, 56ms is 3.5 frames, which means it can effectively get presented at the same time as other entries with 64ms of latency.
* Conversely, if you were to change to 48ms durationThreshold, you really should probably use 40ms to include entries that take 2.5 frames @ 60hz.
* If you wanted to group entries by frame (i.e. by presentation time) you would need to take startTime + duration and then group entries which are within 8ms windows.  A single shared presentation time can include entries with a wide range of durations.  An entry with less than `durationThreshold` duration can have large processing time.  (This is one reason why we may want a dedicated api)

src/getINP.ts
```diff
@@ -14,118 +14,127 @@
...
+
   report = bindReporter(onReport, metric, reportAllChanges);
 
   if (po) {
     onHidden(() => {
-      po.takeRecords().map(entryHandler as PerformanceEntryHandler);
+      handleEntries(po.takeRecords());
+
+      // If the interaction count shows that there were interactions but
+      // none were captured by the PerformanceObserver, report a latency of 0.
```
mmocny, line 124: Nice!  This cannot happen with the existing polyfill but could with the real version later.

src/getINP.ts
```diff
@@ -14,118 +14,127 @@
...
+      if (metric.value < 0 &&
+          getInteractionCount() - prevInteractionCount > 0) {
```
mmocny, line 126: This seems potentially brittle.  Could you create a simple helper called getInteractionCountForCurrentNavigation() or something, or is this specifically inlined to reduce code size?

src/lib/polyfills/interactionCountPolyfill.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+let po: PerformanceObserver | undefined;
+
+/**
+ * Returns the `interactionCount` value using the native API (if available)
+ * or the polyfill estimate in this module.
+ */
+export const getInteractionCount = () => {
+  return po ? interactionCountEstimate : performance.interactionCount || 0;
```
mmocny, line 50: Why `|| 0` at the end?  Defensive?

mmocny (commented):
Changes look solid, nice!

Most comments are nits, but I think a few may be actual bugs.

I didn't review or try the unit tests.

src/getINP.ts
```diff
@@ -14,118 +14,127 @@
...
+/**
+ * Takes a performance entry and adds it to the list of worst interactions
+ * if its duration is long enough to make it among the worst. If the
+ * entry is part of an existing interaction, it is merged and the latency
+ * and entries list is updated as needed.
+ */
+const processEntry = (entry: PerformanceEventTiming) => {
+  // Only process the entry if it's possibly one of the ten longest.
+  if (longestInteractions.length < MAX_INTERACTIONS_TO_CONSIDER ||
+      entry.duration > longestInteractions[longestInteractions.length - 1].latency) {
```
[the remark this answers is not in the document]
philipwalton: Ahh, good catch. I think we want to keep the optimization (since it contains a `findIndex()` and `sort()`) but add an additional check to see if the entry is part of an existing interaction. I think that covers all cases, right?

src/getINP.ts
```diff
@@ -14,118 +14,127 @@
...
-  const existing = largestINPEntries.findIndex((other) => entry.interactionId == other.interactionId);
-  if (existing >= 0) {
-    // Only replace if this one is actually longer
-    if (entry.duration > largestINPEntries[existing].duration) {
-      largestINPEntries[existing] = entry;
+    // If the interaction already exists, update it. Otherwise create one.
+    if (existingInteractionIndex >= 0) {
+      const interaction = longestInteractions[existingInteractionIndex];
+      interaction.latency = Math.max(interaction.latency, entry.duration);
+      interaction.entries.push(entry);
```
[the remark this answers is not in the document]
philipwalton: That was an open question I had (what to sort they by). I think (?) we want to keep them in the order they were dispatched and leave any sorting up to the debug code, but let me know if you disagree and think it's important to have the largest duration first.

src/getINP.ts
```diff
@@ -14,118 +14,127 @@
...
+  latency: number;
+  entries: PerformanceEventTiming[];
 }
 
-function estimateInteractionCount(): number {
-  return (maxKnownInteractionId > 0) ? ((maxKnownInteractionId - minKnownInteractionId) / 7) + 1 : 0;
-}
+// Used to store the interaction count after a bfcache restore, since p98
+// interaction latencies should only consider the current navigation.
+let prevInteractionCount = 0;
```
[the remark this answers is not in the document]
philipwalton: > However, I wonder if there are alternative designs that are navigation-id aware?

Yep, had the same thought. I was trying to explain this point when we spoke on Thursday, but clearly I didn't do a good enough job :)

src/getINP.ts
```diff
@@ -14,118 +14,127 @@
...
+      metric.value = inp.latency;
+      metric.entries = inp.entries;
+      report();
+    }
   };
 
-  const po = observe('event', entryHandler as PerformanceEntryHandler);
+  const po = observe('event', handleEntries, {
+    // The use of 50 here as a balance between not wanting the callback to
+    // run for too many already fast events (one frame or less), but also
```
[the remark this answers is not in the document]
philipwalton: Yeah, I agree, will update.

src/getINP.ts
philipwalton: It can happen with the existing polyfill because I use `durationThreshold:0` for that.

philipwalton: It can happen with the existing polyfill because I use `durationThreshold:0` for that.

src/getINP.ts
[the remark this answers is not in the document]
mmocny: Alright, fair enough.

While I think making the first entry the largest would be reasonable given the current implementation -- my ultimate preference is to report all entries that present the same frame such that we can create a summary of where time was spent overall.  (Or even just report the summary as part of the Metric return value).

So leaving it up to debug code sgtm.

src/getINP.ts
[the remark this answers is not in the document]
mmocny: I think so!

src/getINP.ts
philipwalton: I can add that. I agree it would be less brittle.

src/lib/polyfills/interactionCountPolyfill.ts
philipwalton: Yeah. I think there could be a case where the PO errors and the browser doesn't support `interactionCount`. It's unlikely but just in case.

src/getINP.ts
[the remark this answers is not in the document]
philipwalton: I very much agree with that ultimate goal. Let's keep it simple for now and re-evaluate once we have more options there or when we plan to release this as stable.

src/getINP.ts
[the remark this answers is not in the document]
philipwalton: Done.

src/getINP.ts
philipwalton: Done.

src/getINP.ts
philipwalton: Done.

src/getINP.ts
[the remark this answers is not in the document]
philipwalton: Done.

PR #225, opened 2022-05-11
Add a config object param to all metric functions

### LINKED ISSUE #224: Support additional configuration options for each function
For the `onINP()` function, we'd like to allow users to pass their own `durationThreshold` option, but currently there's no good way to do that. If we change the 3rd param from a `reportAllChanges: boolean` to a configuration object, then we could support this as well as future configuration.

philipwalton:
Fixes #224 by adding a optional, configuration object as the second parameter for all metric function.

This PR also adds the ability to set the `durationThreshold` parameter when measuring INP, which is useful if you to get more granular debug data on interactions that are already relatively fast (the default `durationThreshold` is `40`).

Here's an example of how you can set the lowest supported `durationThreshold` value supported in browsers.

```js
import {onINP} from 'web-vitals';

onINP(console.log, {
  durationThreshold: 16,
  reportAllChanges: true,
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ _**Important!** users who want to load version 3 beta from the unpkg CDN should 
 
 ### Basic usage
 
-Each of the Web Vitals metrics is exposed as a single function that takes an `onReport` callback. This callback will be called any time the metric value is available and ready to be reported.
+Each of the Web Vitals metrics is exposed as a single function that takes a `callback` function that will be called any time the metric value is available and ready to be reported.
 
 The following example measures each of the Core Web Vitals metrics and logs the result to the console once its value is ready to report.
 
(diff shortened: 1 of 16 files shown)


philipwalton:
Yeah, it potentially could. The main unknown is what the performance impact is of setting to a lower threshold actually is. @mmocny is planning to run some benchmarks.

rviscomi:
Is there an issue I can subscribe to where that's being tracked?

mmocny:
https://bugs.chromium.org/p/chromium/issues/detail?id=1321807

Not sure how highly to prioritize that work.  Let us know demand in the bug :)

rviscomi:
Thanks. The reporting lag in the extension did throw me off, but I'll let you know if there's additional developer feedback to make this a higher priority.

We can also make the duration threshold configurable in the extension options, so developers can choose to trade performance for better reporting.

mmocny:
Reporting lag I think is likely to come from something other than slowness
of the PerformanceObserver API itself.  When responsiveness sucks, it's
likely due to main blockage, which itself blocks the INP snippet.

But yes, the default 40ms right now means sometimes it feels like its not
working.

Philip previously suggested one simple fix is to use the FID event for the
initial INP value, for cases where INP<durationThreshold.

-Michal

Issue #228, opened 2022-05-12
Consider expose FID `duration` as first `INP` score, for cases when its under `durationThreshold`


mmocny:
`onINP` relies on event timing `durationThreshold` and provides a default of 40ms (configurable down to 16ms).

There are cases where we report `onFID` but not `onINP` because of this threshold.  `first-input` event timing actually does already expose the `duration` value so we can report the full latency exactly as INP would measure it, and not just the input delay.

Would help address https://github.com/GoogleChrome/web-vitals-extension/issues/103

mmocny:
> first-input entries do not have an interactionId

Filed crbug.com/1325826

> If we report the first-input entry it may give the false impression that INP was actually that low.

It is that low, at the time of the first interaction.  I guess you mean that the *second* interaction, if it was <durationThreshold, may affect INP in ways we don't catch with onINP.  Fair, but tbh, that doesn't convince me we shouldn't record the first-input.

> only benefits the extension (which can do this itself)

I do agree the current demand is just the extension, and we can fix there.  However, if any RUM providers start measuring INP, the fact that INP is an OPTIONAL metric, recording NULL in cases where there is an interaction will have a larger measurement effect than recording FID.duration instead of some slightly larger INP.

> a really simple solution to this problem is to just to the following

I like it!

> Tangentially, I wish we had made the first-input behavior part of the Event Timing API 

I wasn't here for the history, but my understanding is that first-input shipped before event timing and we don't want to remove anything ever.  So it wasn't so much that we added `first-input`, as we added `event`.

> includeFirstInteraction: true

Interesting.  Personally, I don't think we need this property.  I agree that it would be nice if event timing could expose the first-input even if it was under 16ms.  Perhaps we could change the durationThreshold minimum of 0 to allow actual 0 for the first-input only?

That wouldn't support the case you show above (asking for 40ms min and first-input) but can handle that in either of the above strategies you list.

---

Net/net, I think you are right that we don't need to add this to web-vitals.js.  Even though I worry about the default behaviour, there are simple workarounds that can be used.  We may be able to just fix the spec to make it work better by default.

mmocny:
One more consideration:

```
let po = new PerformanceObserver(list => {
  for (let entry of list.getEntries()) {
    if (entry.entryType != 'first-input' || !entry.interactionId)
      return;

[...]

PR #231, opened 2022-05-24
Fix 228: Change onINP to also use FID `duration` when under `durationTheshold`

mmocny:
This is a **PROTOTYPE** for adding support for registering multiple observer types with the same PO observer -- specifically to use both `first-input` and `event` timing together, with the same PerformanceObserver / handler.

I have not tested this thoroughly, just checking to see how it would look for now and to continue conversation.

Context:
* https://github.com/GoogleChrome/web-vitals/issues/228#issuecomment-1128147203
* Chromium is [fixing the fact that FID does not have interactionID](https://crbug.com/1325826)

### DIFF
--- a/src/lib/observe.ts
+++ b/src/lib/observe.ts
@@ -30,19 +30,21 @@ interface PerformanceEntriesHandler {
  * try/catch to avoid errors in unsupporting browsers.
  */
 export const observe = (
-  type: string,
+  types: string[],
   callback: PerformanceEntriesHandler,
   opts?: PerformanceObserverInit,
 ): PerformanceObserver | undefined => {
   try {
-    if (PerformanceObserver.supportedEntryTypes.includes(type)) {
+    if (types.every(type => PerformanceObserver.supportedEntryTypes.includes(type))) {
       const po: PerformanceObserver = new PerformanceObserver((list) => {
         callback(list.getEntries());
       });
-      po.observe(Object.assign({
-        type,
-        buffered: true,
-      }, opts || {}) as PerformanceObserverInit);
+      for (let type of types) {
+        po.observe(Object.assign({
+          type,
+          buffered: true,
+        }, opts || {}) as PerformanceObserverInit);
+      }
       return po;
     }
   } catch (e) {
--- a/src/onINP.ts
+++ b/src/onINP.ts
@@ -116,7 +116,7 @@ export const onINP = (onReport: ReportCallback, opts?: ReportOpts) => {
 
   const handleEntries = (entries: Metric['entries']) => {
     (entries as PerformanceEventTiming[]).forEach((entry) => {
-      if (entry.interactionId) {
+      if (entry.entryType == 'first-input' || entry.interactionId) {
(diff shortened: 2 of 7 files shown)


src/lib/observe.ts
```diff
@@ -30,19 +30,21 @@ interface PerformanceEntriesHandler {
  * try/catch to avoid errors in unsupporting browsers.
  */
 export const observe = (
-  type: string,
+  types: string[],
```
mmocny, line 33: I considered making this `string || string[]` and adjusting this implementation to support either option... decided it was better to make a single choice but totally understand alterantive.

src/lib/observe.ts
```diff
@@ -30,19 +30,21 @@ interface PerformanceEntriesHandler {
...
       });
-      po.observe(Object.assign({
-        type,
-        buffered: true,
-      }, opts || {}) as PerformanceObserverInit);
+      for (let type of types) {
+        po.observe(Object.assign({
+          type,
+          buffered: true,
+        }, opts || {}) as PerformanceObserverInit);
```
mmocny, line 46: Right now, the same options are passed for all registered types.  This is fine for this specific case.

But in general we may want to have `[string, PerformanceObserverInit][]` (aka array of tuple) instead

src/onINP.ts
```diff
@@ -116,7 +116,7 @@ export const onINP = (onReport: ReportCallback, opts?: ReportOpts) => {
 
   const handleEntries = (entries: Metric['entries']) => {
     (entries as PerformanceEventTiming[]).forEach((entry) => {
-      if (entry.interactionId) {
+      if (entry.entryType == 'first-input' || entry.interactionId) {
```
mmocny, line 119: This is only needed until `first-input` gets a proper interactionID.

For now, this leads to the "10 worst" list being 1 larger.  But since web-vitals.js doesn't use that to count number of interactions, it doesn't really matter.  The only time it matters is if the `first-input` duration becomes the 10th largest and ends up evicting another interaction from the list, and then you also have 500 interactions.

philipwalton:
Thanks for filing @mmocny. This PR helped me realize that the change to support #228 wouldn't end up being much extra code.

I did have some concerns about this implementation, though, so I filed #232 to compare. One issue I encountered with this PR is that some of the tests fail (specifically the ones that test the high-percentile logic) because they create a long first interaction and then assert that that interaction is not INP after 50 additional interactions. But if we don't dedupe cases where the `first-input` entry is also one of the longest interactions, then that logic breaks. In reality I don't think it would be a huge deal, but it was easy enough for me to dedupe those, and then I didn't have to change the test logic.

I also thought it would be easier to do a one-off additional `observe()` call in `onINP()` rather than adding support for additional types in the helper module.

LMKWYT.

mmocny:
Yes I figured those types of tests would break, but decided it wasn't likely a big deal on real sites (i.e. the difference wouldn't come up in p75 field data).

However, if you found a way to dedupe, woot!  I'll take a look at that patch.

PR #232, opened 2022-05-30
Update onINP to also observe first-input entries

### LINKED ISSUE #228: Consider expose FID `duration` as first `INP` score, for cases when its under `durationThreshold`

philipwalton:
Fixes #228 by having the observer created in `onINP()` also observe `first-input` entries (in addition to `event`)—given that they are both instances of `PerformanceEventTiming`.

This PR also includes logic to remove duplicate interactions by checking to see if a `first-input` and `event` entry both have the same `startTime` and `duration` values. This logic is needed until [crbug/1325826](https://crbug.com/1325826) is fixed in Chrome, which adds support for the `interactionId` property on `first-input` entries.

/cc @mmocny

### DIFF
--- a/src/onINP.ts
+++ b/src/onINP.ts
@@ -119,6 +119,26 @@ export const onINP = (onReport: ReportCallback, opts?: ReportOpts) => {
       if (entry.interactionId) {
         processEntry(entry);
       }
+
+      // Entries of type `first-input` don't currently have an `interactionId`,
+      // so to consider them in INP we have to first check that an existing
+      // entry doesn't match the `duration` and `startTime`.
+      // Note that this logic assumes that `event` entries are dispatched
+      // before `first-input` entries. This is true in Chrome but it is not
+      // true in Firefox; however, Firefox doesn't support interactionId, so
+      // it's not an issue at the moment.
+      // TODO(philipwalton): remove once crbug.com/1325826 is fixed.
+      if (entry.entryType === 'first-input') {
+        const noMatchingEntry = !longestInteractionList.some((interaction) => {
+          return interaction.entries.some((prevEntry) => {
+            return entry.duration === prevEntry.duration &&
+                entry.startTime === prevEntry.startTime;
+          });
+        });
+        if (noMatchingEntry) {
+          processEntry(entry);
+        }
+      }
     });
 
     const inp = estimateP98LongestInteraction();
@@ -143,6 +163,10 @@ export const onINP = (onReport: ReportCallback, opts?: ReportOpts) => {
   report = bindReporter(onReport, metric, opts.reportAllChanges);
 
   if (po) {
+    // Also observe entries of type `first-input`. This is useful in cases
+    // where the first interaction is less than the `durationThreshold`.
+    po.observe({type: 'first-input', buffered: true});
+
     onHidden(() => {
       handleEntries(po.takeRecords());
 
--- a/test/views/inp.njk
+++ b/test/views/inp.njk
@@ -89,7 +89,7 @@
       inp.entries = inp.entries.map((e) => ({
         ...e.toJSON(),
         interactionId: e.interactionId,
-        target: e.target.nodeName.toLowerCase(),
+        target: e.target?.nodeName.toLowerCase(),
       }));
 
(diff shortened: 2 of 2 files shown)


src/onINP.ts
```diff
@@ -119,6 +119,26 @@ export const onINP = (onReport: ReportCallback, opts?: ReportOpts) => {
...
         processEntry(entry);
       }
+
+      // Entries of type `first-input` don't currently have an `interactionId`,
+      // so to consider them in INP we have to first check that an existing
+      // entry doesn't match the `duration` and `startTime`.
+      // Note that this logic assumes that `event` entries are dispatched
+      // before `first-input` entries. This is true in Chrome but it is not
+      // true in Firefox; however, Firefox doesn't support interactionId, so
+      // it's not an issue at the moment.
```
mmocny, line 129: I was surprised to learn this, since I think we fire FID first.

However, I guess you mean that the order of Entries in the PO callbacks, when there are multiple observe() calls of different types... This I haven't tested and I'm not sure if order is guaranteed or arbitrary right now.

src/onINP.ts
```diff
@@ -119,6 +119,26 @@ export const onINP = (onReport: ReportCallback, opts?: ReportOpts) => {
...
+      // TODO(philipwalton): remove once crbug.com/1325826 is fixed.
+      if (entry.entryType === 'first-input') {
+        const noMatchingEntry = !longestInteractionList.some((interaction) => {
+          return interaction.entries.some((prevEntry) => {
+            return entry.duration === prevEntry.duration &&
+                entry.startTime === prevEntry.startTime;
+          });
+        });
+        if (noMatchingEntry) {
+          processEntry(entry);
```
mmocny, line 139: What happens if:
- FID is pointerdown, but we do not get a pointerdown event timing entry
- Then we do get an pointerup event timing entry (for the same interaction)

I think it will be treated as 2 different interactions here, also affecting p98.

src/onINP.ts
```diff
@@ -143,6 +163,10 @@ export const onINP = (onReport: ReportCallback, opts?: ReportOpts) => {
   report = bindReporter(onReport, metric, opts.reportAllChanges);
 
   if (po) {
+    // Also observe entries of type `first-input`. This is useful in cases
+    // where the first interaction is less than the `durationThreshold`.
+    po.observe({type: 'first-input', buffered: true});
```
mmocny, line 168: Yeah this seems better :)

test/views/inp.njk
```diff
@@ -89,7 +89,7 @@
       inp.entries = inp.entries.map((e) => ({
         ...e.toJSON(),
         interactionId: e.interactionId,
-        target: e.target.nodeName.toLowerCase(),
+        target: e.target?.nodeName.toLowerCase(),
```
mmocny, line 92: Is this related to FID entry?

mmocny (approved):
I don't think this is perfect either, but I think its a good change and cleaner than the one I suggested.

Cheers!

src/onINP.ts
philipwalton: Yeah, I only tested this in Chrome on Mac (desktop), but no matter what I did, which order I called `observe()` in, or which event I used, I always got the `event` entry processed before the `first-input` entry.

test/views/inp.njk
philipwalton: Yeah, not sure why but I started noticing lots of cases where `target` was undefined on `first-input` entries. I didn't look much further into repo-ing why.

src/onINP.ts
philipwalton: Yeah, I think so. My hope is we can remove this logic entirely once `first-input` entries have an `interactionId` property. My goal here was to minimize these types of occurrences .

Issue #203, opened 2022-02-18
Add diagnostic and contextualizing data for metrics


malchata:
Collecting field data for performance is a tall order to begin with, but analyzing it without contextual data can make interpreting and acting on field data much more difficult. In my opinion, I believe that adding the following diagnostic information for the following metrics could be helpful:

1. TTFB values should be supplemented with [server timing API](https://developer.mozilla.org/docs/Web/HTTP/Headers/Server-Timing) data (where available).
2. CLS values should be supplemented with data from the [layout instability API](https://developer.mozilla.org/docs/Web/API/Layout_Instability_API).
3. LCP values should be supplemented with data from the [largest contentful paint](https://developer.mozilla.org/docs/Web/API/Largest_Contentful_Paint_API) and [element timing](https://developer.mozilla.org/docs/Web/API/Element_timing_API) APIs.
4. FCP values should be supplemented with data from the [paint timing API](https://css-tricks.com/paint-timing-api/).
5. FID values should be supplemented with data from the [user timing](https://developer.mozilla.org/docs/Web/API/User_Timing_API) and [long task](https://developer.mozilla.org/docs/Web/API/Long_Tasks_API) APIs.

By exposing this diagnostic information in this library in a convenient fashion, developers may be more likely to record and act on contextual field data that may help them to better diagnose performance problems in the field.

PR #237, opened 2022-07-10
Add a new attribution build for debugging issues in the field

philipwalton:
Addresses #203. 

This PR adds a new "attribution" build to this library. When used, it augments the `Metric` object with an `attribution` property that contains useful information to debug issues that occur in the field.

This is essentially a codification of the recommendations outlined in the post [Debug Web Vitals in the field](https://web.dev/debug-web-vitals-in-the-field/), and it also includes some additional diagnostic information like outlined in the updated [Optimize LCP](https://web.dev/optimize-lcp/#lcp-breakdown) guide.

The API to use this library doesn't change, to get the attribution data you have to important from `web-vitals/attribution` rather than `web-vitals`:

```diff
- import {onLCP, onFID, onCLS} from 'web-vitals';
+ import {onLCP, onFID, onCLS} from 'web-vitals/attribution';
```

Then, when the metric callback functions are invokes, the `Metric` object passed to them contain a new `attribution` property:

```ts
interface MetricWithAttribution extends Metric {
  attribution: {[key: string]: unknown};
}
```

And as an example of how you might use this in code, where's how you would send data about what element was involved in the largest layout shift contributing to CLS on the page:

```js
import {onCLS} from 'web-vitals/attribution';

onCLS((metric) => {
  navigator.sendBeacon('/analytics', JSON.stringify({
    name: metric.name,
    value: metric.value,
    id: metric.id,

    // This will be a string selector helping identifying the element that shifted.
    // For example: "html>body>main>section>h1.main-heading"
    largestShiftTarget: metric.attribution.largestShiftTarget,
  }));
});
```

The following type definitions gives an overview of what attribution data could be set on each object, depending on which metric is being observed.

**CLS**:

```ts
interface CLSAttribution {
  /**
   * A selector identifying the first element (in document order) that
   * shifted when the single largest layout shift contributing to the page's
   * CLS score occurred.
   */
  largestShiftTarget?: string;
  /**
   * The time when the single largest layout shift contributing to the page's
   * CLS score occurred.
   */
  largestShiftTime?: DOMHighResTimeStamp;
  /**
   * The layout shift score of the single largest layout shift contributing to
   * the page's CLS score.
   */
  largestShiftValue?: number;
  /**
   * The `LayoutShiftEntry` representing the single largest layout shift
   * contributing to the page's CLS score. (Useful when you need more than just
   * `largestShiftTarget`, `largestShiftTime`, and `largestShiftValue`).
   */
  largestShiftEntry?: LayoutShift;
  /**
   * The first element source (in document order) among the `sources` list
   * of the `largestShiftEntry` object. (Also useful when you need more than
   * just `largestShiftTarget`, `largestShiftTime`, and `largestShiftValue`).
   */
  largestShiftSource?: LayoutShiftAttribution;
  /**
   * The loading state of the document at the time when the largest layout
   * shift contribution to the page's CLS score occurred (see `LoadState`
   * for details).
   */
  loadState?: LoadState;
}
```

**FCP**:

```ts
interface FCPAttribution {
  /**
   * The time from when the user initiates loading the page until when the
   * browser receives the first byte of the response (a.k.a. TTFB).
   */
  timeToFirstByte: number;
  /**
   * The time between TTFB and the first contentful paint (FCP).
   */
  renderDelay: number;
  /**
   * The loading state of the document at the time when FCP `occurred (see
   * `LoadState` for details). Ideally, documents can paint before they finish
   * loading (e.g. the `loading` or `domInteractive` phases).
   */
  loadState: LoadState,
  /**
   * The `navigation` entry of the current page, which is useful for diagnosing
   * general page load issues.
   */
  navigationEntry?: PerformanceNavigationTiming | NavigationTimingPolyfillEntry;
}
```

**FID**:

```ts
interface FIDAttribution {
  /**
   * A selector identifying the element that the user interacted with. This
   * element will be the `target` of the `event` dispatched.
   */
  eventTarget: string;
  /**
   * The time when the user interacted. This time will match the `timeStamp`
   * value of the `event` dispatched.
   */
  eventTime: number;
  /**
   * The `type` of the `event` dispatched from the user interaction.
   */
  eventType: string;
  /**
   * The loading state of the document at the time when the first interaction
   * occurred (see `LoadState` for details). If the first interaction occurred
   * while the document was loading and executing script (e.g. usually in the
   * `domInteractive` phase) it can result in long input delays.
   */
  loadState: LoadState;
}
```

**INP**:

```ts
interface INPAttribution {
  /**
   * A selector identifying the element that the user interacted with. This
   * element will be the `target` of the `event` dispatched.
   */
  eventTarget?: string;
  /**
   * The time when the user interacted. This time will match the `timeStamp`
   * value of the `event` dispatched.
   */
  eventTime?: number;
  /**
   * The `type` of the `event` dispatched from the user interaction.
   */
  eventType?: string;
  /**
   * The loading state of the document at the time when the interaction
   * occurred (see `LoadState` for details). If the interaction occurred
   * while the document was loading and executing script (e.g. usually in the
   * `domInteractive` phase) it can result in long delays.
   */
  loadState?: LoadState;
}
```

**LCP**:

```ts
interface LCPAttribution {
  /**
   * The element corresponding to the largest contentful paint for the page.
   */
  element?: string,
  /**
   * The time from when the user initiates loading the page until when the
   * browser receives the first byte of the response (a.k.a. TTFB). See
   * [Optimize LCP](https://web.dev/optimize-lcp/) for details.
   */
  timeToFirstByte: number;
  /**
   * The delta between TTFB and when the browser starts loading the LCP
   * resource (if there is one, otherwise 0). See [Optimize
   * LCP](https://web.dev/optimize-lcp/) for details.
   */
  resourceLoadDelay: number;
  /**
   * The total time it takes to load the LCP resource itself (if there is one,
   * otherwise 0). See [Optimize LCP](https://web.dev/optimize-lcp/) for
   * details.
   */
  resourceLoadTime: number;
  /**
   * The delta between when the LCP resource finishes loading until the LCP
   * element is fully rendered. See [Optimize
   * LCP](https://web.dev/optimize-lcp/) for details.
   */
  elementRenderDelay: number;
  /**
   * The `navigation` entry of the current page, which is useful for diagnosing
   * general page load issues.
   */
  navigationEntry?: PerformanceNavigationTiming | NavigationTimingPolyfillEntry;
  /**
   * The `resource` entry for the LCP resource (if applicable), which is useful
   * for diagnosing resource load issues.
   */
  lcpResourceEntry?: PerformanceResourceTiming;
}
```

One downside of exposing the attribution data via a separate build is it means you either have to get attribution for all metrics or none of them—you can't just get attribution for one, e.g. `onCLS()`, and then use the unattributed versions of all the other metrics (well, you can, but doing so would import both versions and double the file size).

However, given the full attribution code for all metrics only adds 500 bytes (brotli'd) to the download size, I think that's worth the convenience of not changing the API. _Note: in the future, if the [pipeline operator](https://github.com/tc39/proposal-pipeline-operator) comes to JavaScript, it may be worth revisiting this decision._

Another downside is there's no easy way to make the attribution build work with the base+polyfill build without massively increasing complexity (to support ESM, UMD, and IIFE versions of each build, making a base+polyfill+attribution build would double the number of builds).

Based on the analysis done in #238, in this initial PR the attribution build does not make use of the polyfill.

* * *

In addition to the above, this PR also updates this package to use `type: module` as well as package exports, now that they are widely supported. As a result, there are additional changes from `require()` to `import` in this PR that are not related to the above changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@
   - [Send the results to an analytics endpoint](#send-the-results-to-an-analytics-endpoint)
   - [Send the results to Google Analytics](#send-the-results-to-google-analytics)
   - [Send the results to Google Tag Manager](#send-the-results-to-google-tag-manager)
+  - [Send attribution data](#send-attribution-data)
   - [Batch multiple reports together](#batch-multiple-reports-together)
 - [Bundle versions](#bundle-versions)
   - [Which bundle is right for you?](#which-bundle-is-right-for-you)
@@ -18,6 +19,7 @@
 - [API](#api)
   - [Types](#types)
   - [Functions](#functions)
+  - [Attribution](#attribution)
 - [Browser Support](#browser-support)
 - [Limitations](#limitations)
 - [Development](#development)
@@ -38,7 +40,7 @@ The library supports all of the [Core Web Vitals](https://web.dev/vitals/#core-w
 
 ### Other metrics
 
-- [Interaction to next Paint (INP)](https://web.dev/responsiveness/) _(experimental)_
+- [Interaction to next Paint (INP)](https://web.dev/inp/) _(experimental)_
 - [First Contentful Paint (FCP)](https://web.dev/fcp/)
 - [Time to First Byte (TTFB)](https://web.dev/ttfb/)
 
@@ -114,13 +116,34 @@ Note that the code _must_ go in the `<head>` of your pages in order to work. See
 
 _**Tip:** while it's certainly possible to inline the code in `dist/polyfill.js` by copy and pasting it directly into your templates, it's better to automate this process in a build step—otherwise you risk the "base" and the "polyfill" scripts getting out of sync when new versions are released._
 
+<a name="attribution-build"><a>
+
+**3. The "attribution" build**
+
+Measuring the Web Vitals scores for your real users is a great first step toward optimizing the user experience. But if your scores aren't _good_, the next step is to understand why they're not good and work to improve them.
+
+The "attribution" build helps you do that by including additional diagnostic information with each metric to help you identify the root cause of poor performance as well as prioritize the most important things to fix.
+
+The "attribution" build is slightly larger than the "standard" build (by about 500 bytes, brotli'd), so while the code size is still small, it's only recommended if you're actually using these features.
+
+To load the "attribution" build, change any `import` statements that reference `web-vitals` to `web-vitals/attribution`:
+
+```diff
+- import {onLCP, onFID, onCLS} from 'web-vitals';
++ import {onLCP, onFID, onCLS} from 'web-vitals/attribution';
+```
+
+Usage for each of the imported function is identical to the standard build, but when importing from the attribution build, the [`Metric`](#metric) object will contain an additional [`attribution`](#metricwithattribution) property.
+
+See [Send attribution data](#send-attribution-data) for usage examples, and the [`attribution` reference](#attribution) for details on what values are added for each metric.
+
 <a name="load-web-vitals-from-a-cdn"><a>
 
 ### From a CDN
 
 The recommended way to use the `web-vitals` package is to install it from npm and integrate it into your build process. However, if you're not using npm, it's still possible to use `web-vitals` by requesting it from a CDN that serves npm package files.
 
-The following examples show how to load `web-vitals` from [unpkg.com](https://unpkg.com), whether your targeting just Chromium-based browsers (using the "standard" version) or additional browsers (using the "base+polyfill" version):
+The following examples show how to load `web-vitals` from [unpkg.com](https://unpkg.com), whether you're targeting just Chromium-based browsers (using the "standard" version) or additional browsers (using the "base+polyfill" version):
 
 _**Important!** users who want to load version 3 beta from the unpkg CDN should specify a version number or link to the [web-vitals@next](https://unpkg.com/web-vitals@next?module) tag._
 
@@ -214,7 +237,7 @@ Also, in some cases a metric callback may never be called:
 
 In other cases, a metric callback may be called more than once:
 
-- CLS should be reported any time the [page's `visibilityState` changes to hidden](https://developers.google.com/web/updates/2018/07/page-lifecycle-api#advice-hidden).
+- CLS should be reported any time the [page's `visibilityState` changes to hidden](https://developer.chrome.com/blog/page-lifecycle-api/#advice-hidden).
 - All metrics are reported again (with the above exceptions) after a page is restored from the [back/forward cache](https://web.dev/bfcache/).
 
 _**Warning:** do not call any of the Web Vitals functions (e.g. `onCLS()`, `onFID()`, `onLCP()`) more than once per page load. Each of these functions creates a `PerformanceObserver` instance and registers event listeners for the lifetime of the page. While the overhead of calling these functions once is negligible, calling them repeatedly on the same page may eventually result in a memory leak._
@@ -315,7 +338,7 @@ function sendToGoogleAnalytics({name, delta, id}) {
     // Use `sendBeacon()` if the browser supports it.
     transport: 'beacon',
 
-    // OPTIONAL: any additional params or debug info here.
+    // OPTIONAL: any additional attribution params here.
     // See: https://web.dev/debug-web-vitals-in-the-field/
     // dimension1: '...',
     // dimension2: '...',
@@ -350,10 +373,10 @@ function sendToGoogleAnalytics({name, delta, id}) {
     // Use a non-interaction event to avoid affecting bounce rate.
     non_interaction: true,
 
-    // OPTIONAL: any additional params or debug info here.
+    // OPTIONAL: any additional attribution params here.
     // See: https://web.dev/debug-web-vitals-in-the-field/
-    // metric_rating: 'good' | 'ni' | 'poor',
-    // debug_info: '...',
+    // dimension1: '...',
+    // dimension2: '...',
     // ...
   });
 }
@@ -399,6 +422,48 @@ onLCP(sendToGoogleAnalytics);
 The recommended way to measure Web Vitals metrics with Google Tag Manager is using the [Core Web Vitals](https://www.simoahava.com/custom-templates/core-web-vitals/) custom template tag created and maintained by [Simo Ahava](https://www.simoahava.com/).
 
 For full installation and usage instructions, see Simo's post: [Track Core Web Vitals in GA4 with Google Tag Manager](https://www.simoahava.com/analytics/track-core-web-vitals-in-ga4-with-google-tag-manager/).
+
+### Send attribution data
+
+When using the [attribution build](#attribution-build), you can send additional data to help you debug _why_ the metric values are they way they are.
+
+This example sends an additional `debug_target` param to Google Analytics, corresponding to the element most associated with each metric.
+
+```js
+import {onCLS, onFID, onLCP} from 'web-vitals/attribution';
+
+function sendToGoogleAnalytics({name, delta, value, id, attribution}) {
+  const eventParams = {
+    // Built-in params:
+    value: delta, // Use `delta` so the value can be summed.
+    // Custom params:
+    metric_id: id, // Needed to aggregate events.
+    metric_value: value, // Optional.
+    metric_delta: delta, // Optional.
+  }
+
+  switch (name) {
+    case 'CLS':
+      eventParams.debug_target = attribution.largestShiftTarget;
+    case 'FID':
+      eventParams.debug_target = attribution.eventTarget;
+    case 'LCP':
+      eventParams.debug_target = attribution.element;
+  }
+
+  // Assumes the global `gtag()` function exists, see:
+  // https://developers.google.com/analytics/devguides/collection/ga4
+  gtag('event', name, eventParams);
+}
+
+onCLS(sendToGoogleAnalytics);
+onFID(sendToGoogleAnalytics);
+onLCP(sendToGoogleAnalytics);
+```
+
+_**Note:** this example relies on custom [event parameters](https://support.google.com/analytics/answer/11396839) in Google Analytics 4. For Universal Analytics the attribution data should be set using a [custom dimension](https://support.google.com/analytics/answer/2709828) rather than `debug_target` as shown above.
+
+See [Debug Web Vitals in the field](https://web.dev/debug-web-vitals-in-the-field/) for more information and examples.
 
 ### Batch multiple reports together
 
@@ -467,7 +532,7 @@ The following table lists all the bundles distributed with the `web-vitals` pack
     <td><code>web-vitals.js</code></td>
     <td><code>pkg.module</code></td>
     <td>
-      <p>An ES module bundle of all metric functions, without any extra polyfills to expand browser support.</p>
+      <p>An ES module bundle of all metric functions, without any attribution features.</p>
       This is the "standard" version and is the simplest way to consume this library out of the box.
     </td>
   </tr>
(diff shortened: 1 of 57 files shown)


philipwalton:
cc: @anniesullie @mmocny @kyliau @tunetheweb

README.md
```diff
@@ -568,10 +610,43 @@ interface Metric {
...
+- [`CLSMetric`](/src/types/cls.ts#:~:text=interface%20CLSMetric)
+- [`FCPMetric`](/src/types/cls.ts#:~:text=interface%20FCPMetric)
+- [`FIDMetric`](/src/types/cls.ts#:~:text=interface%20FIDMetric)
+- [`INPMetric`](/src/types/cls.ts#:~:text=interface%20INPMetric)
+- [`LCPMetric`](/src/types/cls.ts#:~:text=interface%20LCPMetric)
+- [`TTFBMetric`](/src/types/cls.ts#:~:text=interface%20TTFBMetric)
+
+#### `MetricWithAttribution`
+
+See the [attribution build](#attribution-build) section for details on how to use this feature.
```
tunetheweb, line 628: This section doesn't really show how to use these - it just documents the API.

Could we update the Universal Analytics/GTM section on lines 338 - 467 to include these?
Especially as it currently has the following:

```
    // OPTIONAL: any additional params or debug info here.
    // See: https://web.dev/debug-web-vitals-in-the-field/
    // metric_rating: 'good' | 'ni' | 'poor',
    // debug_info: '...',
    // ...
```

tunetheweb (commented):
I **love** the idea of adding attribution directly to the library!

It's one of the things I always have to look up, and also means that code is likely stale and not updated for any improvements. Including it in the library means users of the library will automatically benefit from those improvements (assuming you keep the library up to date of course).

Some early comments as not gone through all the actual code changes.

anniesullie:
Excited to see this change! I think the list of attributes for each metric is good.

src/attribution/onFCP.ts
```diff
@@ -0,0 +1,60 @@
...
+  if (metric.entries.length) {
+    const navigationEntry = getNavigationEntry();
+
+    if (navigationEntry) {
+      const activationStart = navigationEntry.activationStart || 0;
+      const ttfb = Math.max(0, navigationEntry.responseStart - activationStart);
+
+      metric.attribution = {
+        timeToFirstByte: ttfb,
+        renderDelay: metric.value - ttfb,
```
mmocny, line 34: Is it reasonable to name this renderDelay?

Typically render delay is the time after content is available, but before it is presented.  TTFB is not the time after content is available, so this value is both the sum of some loadTime + renderDelay.

I don't know if it's possible to define loadTime for FCP, so a combined metric may make sense... does is the name renderDelay misleading?

src/attribution/onCLS.ts
```diff
@@ -0,0 +1,77 @@
...
+    const largestEntry = getLargestLayoutShiftEntry(metric.entries);
+    if (largestEntry && largestEntry.sources && largestEntry.sources.length) {
+      const largestSource = getLargestLayoutShiftSource(largestEntry.sources);
+      if (largestSource) {
+        attribution = {
+          largestShiftTarget: getSelector(largestSource.node),
+          largestShiftTime: largestEntry.startTime,
+          largestShiftValue: largestEntry.value,
+          largestShiftSource: largestSource,
+          largestShiftEntry: largestEntry,
```
mmocny, line 43: Just checking: this Entry would already be in the entries array, so this is just a convenience?

I'm just curious if the other attributions should expose "the most important entry" even if its an array size 1.  FID and FCP that seems a bit pointless, but for LCP/INP it seems reasonable?

src/attribution/onINP.ts
```diff
@@ -0,0 +1,74 @@
...
+import {getSelector} from '../lib/getSelector.js';
+import {onINP as unattributedOnINP} from '../onINP.js';
+import {INPMetricWithAttribution, INPReportCallback, INPReportCallbackWithAttribution, ReportOpts} from '../types.js';
+
+
+const attributeINP = (metric: INPMetricWithAttribution): void => {
+  if (metric.entries.length) {
+    const longestEntry = metric.entries.sort((a, b) => {
+      // Sort by: 1) duration (DESC), then 2) processing time (DESC)
+      return b.duration - a.duration || (b.processingEnd - b.processingStart) -
```
mmocny, line 27: Oh I like this.  We haven't used the "longest processing time if durations are equal" approach before.

If I recall, metric.entries will be all the entries with the same interactionId, right?  So if we change Event Timing to expose interactionId on more event types, this will "just work" better.  Cool.

src/attribution/onINP.ts
```diff
@@ -0,0 +1,74 @@
...
+      (a.processingEnd - a.processingStart);
+    })[0];
+
+    metric.attribution = {
```
mmocny, line 31: Should you expose the longest Entry as well?  Similar to CLS...

mmocny (approved):
I love this!  Thanks for the awesome work here.


One overall Nit and a few comment in-line:

- The attribution values use the term `Delay` and `Time` interchangeably.  I think all of these represent durations not timestamps, but I assumed some of the "Time" attribution were timestamps because of the name differences.

Looking at the attributions, I think "Time" means "actively working" and "delay" means "waiting around", but it's a bit of a loose fit for things like TTFB.

Perhaps some more of these "Time" names should be "Delay" by those criteria, and perhaps "Time" should be renamed something like "Duration"?

Very soft opinion on that feedback.

src/attribution/onFCP.ts
philipwalton: I struggled a lot with naming on this one.

I agree that "delay" here doesn't fit with how delay is used in other attribution properties. I was also considering `renderTime`, but `renderTime` is a spec'ed property of the LCP entry, and there it's defined as a timestamp and not a duration, so I thought that might be confusing.

I went with `renderDelay` for consistency with `elementRenderDelay` in LCP attribution, but I agree that's not an entirely apples-to-apples comparison.

Anyway, all that to say, I'm open to renaming this (and others), but I don't currently have a better name in mind, so I'm open to suggestions.

src/attribution/onINP.ts
philipwalton: Yep! I was pretty happy with the simplicity of this approach to be honest :)

src/attribution/onINP.ts
philipwalton: Yeah, I think so.

What would you think about renaming all of these to `longestEventTarget`, `longestEventType`, etc. for consistency with CLS as well?

src/lib/initMetric.ts
```diff
@@ -25,7 +25,7 @@ export const initMetric = (name: Metric['name'], value?: number): Metric => {
   const navEntry = getNavigationEntry();
   let navigationType: Metric['navigationType'];
 
-  if (isBFCacheRestore()) {
+  if (getBFCacheRestoreTime() > 0) {
```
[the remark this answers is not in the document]
philipwalton: Good catch, thanks!

README.md
```diff
@@ -114,6 +115,25 @@ Note that the code _must_ go in the `<head>` of your pages in order to work. See
...
+The "attribution" build helps you do that by including additional diagnostic information with each metric to help you identify the root cause of poor performance as well as prioritize the most important things to fix.
+
+To load the "attribution" build, change any `import` statements that reference `web-vitals` to `web-vitals/attribution`:
+
+```diff
+- import {onLCP, onFID, onCLS} from 'web-vitals';
++ import {onLCP, onFID, onCLS} from 'web-vitals/attribution';
+```
+
+Usage for each of the imported function is identical to the standard build, but when importing from the attribution build, the [`Metric`](#metric) object will contain an additional [`attribution`](#metricwithattribution) property.
```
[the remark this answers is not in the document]
philipwalton: Good suggestion. Will add.

README.md
philipwalton: Yep, will add.

src/attribution/onFCP.ts
mmocny: What I meant was `render` part of the name, not `Delay` part.  I don't mind using `Delay` since I dislike using `Time` for durations.

I think renderDelay here is different than elementRenderDelay, in that FCP really also has a `resourceLoadDelay` and `resourceLoadTime` its just less well defined.  This value if lumping them all together.

philipwalton: In the case of FCP, I'm not sure that we'd want to break the time down further? At least not in a way that's specific to resource load given that pages can reach FCP without loading any resources (and that's historically what we've recommended, e.g. remove render-blocking resources).

src/attribution/onCLS.ts
mmocny: I think it could be convenient.

Also, I can see a world where FID and LCP return more than a single Entry in the `entries` list, and so this would still have value in revealing the "main" one (probably always the first in list):

- For FID, it could be all the events of the interaction.
- For LCP, it could be all previous candidates.

src/attribution/onFCP.ts
mmocny: If there is text content, then the main document is the resource.  If there is no text, then we will not hit FCP until the first image is painted, and so it may not be render-blocking but its still FCP-blocking so to speak.

Either way, whatever the resource, the specific paint that leads to FCP you can always at the very least break down to the time it took to get to the start of the paint operation, and the additional time it took to present the frame from there.  This helps find cases where rAF, style, layout, paint, raster, decode, or draw end up taking a long time.

philipwalton: > You can always at the very least break down to the time it took to get to the start of the paint operation, and the additional time it took to present the frame from there.

But none of that information is exposed via any RUM APIs, correct?

I agree that someone  trying to diagnose FCP issues would look at all of this information in a DevTools trace, but from a RUM analytics perspective, I think the most useful signal is the high-level: 1) is FCP slow due to network or server slowness? Or 2) is FCP slow due to front-end rendering bottlenecks.

I think if/when we have the [render blocking status](https://chromestatus.com/feature/5166965277589504) in Resource Timing, we could add that as an additional signal, but otherwise I'm not sure how else we could break down the `renderDelay` portion generically given existing RUM APIs.

mmocny: To reiterate, I have no concerns with how you have defined this value here for FCP.  I think it's the most reasonable way to attribute today.

I'm really just expressing a soft-concern about using the term "renderDelay" to mean a slightly different thing in this context than it is elsewhere... especially since it's conceivable we'll want to split it out sometimes.

As you say, you can do this split out in the lab-- so if we think that the term "renderDelay" would be a good name for the final subset of the time in lab tooling-- even just for consistency reasons alone, I think the field value should be distinct.

(Also: It is not available to RUM via paint timing yet but I think there is some hope that it could be in the future.  And, even today if you are listening to rAF timings you can sorta compute it yourself).

mmocny: Sorry for the bikeshed.

Of those options, I think I like `responseToRender` best.

Other options:
* `firstByteToFCP` (or replace FCP with paint/render/etc)
* `loadAndRenderDelay`

philipwalton: Let's go with `firstByteToFCP`. It's _very_ clear what timestan it covers, and if in the future we're able to break it down further, we could add additional sub-parts without causing too much confusion or forcing a breaking change.

mmocny:
huzzah!

philipwalton:
@mmocny @tunetheweb I believe I addressed all your feedback. If you have time to take another look, that would be great.

tunetheweb (approved):
LGTM

A couple of suggestions for < GA4 (since many people use this library for web-vitals-report which doesn't support GA4 yet).

PR #243, opened 2022-07-15
Add GitHub Super-Linter and editorconfig

tunetheweb:
I'm a big fan of linting, and GitHub Actions to automate checks. Given, how important and widely used this library is, I think it would be good to add these to ensure cleanliness of code and avoid bugs. Ideally we'd also automated testing the code, but linting is a start.

The GitHub Super-Linter is a GitHub Action which bundles many linters, implemented in many languages, in a Docker image, giving you access to all of them without needing to add any dependencies to the project. It also checks more than just code, but also Markdown, JSON, YAML...etc. For the actual JS/TS it will use the eslint config file already set up in this project.

I've also added an .editorconfig file to enforce certain things (spaces over tabs, no trailing spaces, file format, trailing newline at EOF...etc.) for those that use that (there's a VS Code plugin to automatically pick it up).

The good news is the code it very clean, with only a few small tweaks needed to get it to pass the linters.

With this PR, these standards would be enforced and automatically checked on each commit. It takes a couple of mins to run  and only runs on changed files (running for the whole code base still only takes 3 mins as this repo is small).

Let me know what you think.


philipwalton:
In general I'm very supportive of adding these types of things. And I'd very much like to set up GitHub actions to run the tests, etc. automatically on PR submissions.

I think I'd prefer to wait until after v3 ships though, just to make sure none of these changes slow that down.

tunetheweb:
No bother. Want me to close this for now? Or happy to leave it open to remind us?

BTW a refactor/version upgrade is maybe the best time to have a linter 😉 And by default it won't be merge blocking (though we should add that once we're comfortable with it).

Will also hold off looking at CI tests for now (not sure how feasible this is on a headless server anyway), but definitely would be nice!

tunetheweb:
FYI @philipwalton I merged this with latest main so good to have a review whenever you get a chance.

tunetheweb:
Closing in favour of #288 which allows fixing to be run locally (and automated with Husky).

PR #245, opened 2022-07-21
Correct `switch` statement

tunetheweb:
Well this wasted far too much time of my life wondering why nothing was working! 😁

### DIFF
--- a/README.md
+++ b/README.md
@@ -460,10 +460,13 @@ function sendToGoogleAnalytics({name, delta, value, id, attribution}) {
   switch (name) {
     case 'CLS':
       eventParams.debug_target = attribution.largestShiftTarget;
+      break;
     case 'FID':
       eventParams.debug_target = attribution.eventTarget;
+      break;
     case 'LCP':
       eventParams.debug_target = attribution.element;
+      break;
   }
 
   // Assumes the global `gtag()` function exists, see:


philipwalton:
I'm fine to add this change, but I'm curious as to why it wasn't working? I don't think this code will make any difference given that the `name` property can never be more than one metric at a time.

tunetheweb:
```js
  switch (name) {
    case 'CLS':
      break;
    case 'FID':
      eventParams.debug_target = attribution.eventTarget;
      break;
    case 'LCP':
      eventParams.debug_target = attribution.element;
      break;
  }

  sendToAnalytics(name, eventParams.debug_target)
```

If I have CLS event, without the `breaks` this is equivalent to this:

```js
     eventParams.debug_target = attribution.largestShiftTarget; // Yeah! That's what I want
     eventParams.debug_target = attribution.eventTarget; // Null - oh oh it's been overwritten 😔
     eventParams.debug_target = attribution.element; // Null again - oh oh it's been overwritten again 😔
     sendToAnalytics(name = "CLS", eventParams.debug_target = Null) // That's not what I wanted!!!
```

philipwalton:
Right, right, right...I always forget that subsequent statements don't have to match.

tunetheweb:
Me too - took me far longer than I'd care to admit to figure out why it wasn't working!!

PR #246, opened 2022-07-21
Add a metric rating property

philipwalton:
This PR address #2 by adding a new `rating` property to the `Metric` object.

```ts
interface Metric {
  /**
   * The rating as to whether the metric value is within the "good",
   * "needs improvement", or "poor" thresholds of the metric.
   */
  rating: 'good' | 'ni' | 'poor';
  
  // ...
}
```

Note: I opted to use "ni" rather than the full string "needs improvement" because I felt it was awkward to have such a long value with a space in it. I'm open to feedback or suggestions on that though.

### DIFF
--- a/src/lib/bindReporter.ts
+++ b/src/lib/bindReporter.ts
@@ -17,23 +17,38 @@
 import {Metric, ReportCallback} from '../types.js';
 
 
+const getRating = (value: number, thresholds: number[]) => {
+  if (value > thresholds[1]) {
+    return 'poor';
+  }
+  if (value > thresholds[0]) {
+    return 'needs-improvement';
+  }
+  return 'good';
+};
+
+
 export const bindReporter = (
   callback: ReportCallback,
   metric: Metric,
+  thresholds: number[],
   reportAllChanges?: boolean,
 ) => {
   let prevValue: number;
+  let delta: number;
   return (forceReport?: boolean) => {
     if (metric.value >= 0) {
       if (forceReport || reportAllChanges) {
-        metric.delta = metric.value - (prevValue || 0);
+        delta = metric.value - (prevValue || 0);
 
         // Report the metric if there's a non-zero delta or if no previous
         // value exists (which can happen in the case of the document becoming
         // hidden when the metric value is 0).
         // See: https://github.com/GoogleChrome/web-vitals/issues/14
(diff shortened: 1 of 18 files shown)


src/lib/bindReporter.ts
```diff
@@ -17,23 +17,38 @@
...
 
+const getRating = (value: number, thresholds: number[]) => {
+  if (value > thresholds[1]) {
+    return 'poor';
+  }
+  if (value > thresholds[0]) {
+    return 'ni';
+  }
+  return 'good';
+};
```
tunetheweb, line 28: Should we check its >= 0 before returning `good`?

Maybe return null or such like if < 0

philipwalton: I don't think we have to. This function is only ever called from within a conditional that check if `metric.value >= 0` (see line 40).

tunetheweb: Seems a good safety check to add in case that ever becomes not there case?

philipwalton: I don't want to type that it could be `string | null`, because I think that would unnecessarily concern users of the library (especially TypeScript users).

I think to handle your concern it's better to have a test that checks that case rather than building it into the code (that users have to download even though it's not currently needed). There are lots of tests currently that make sure the metric value is greater than or equal to 0, and those would all start failing if that were ever not the case.

tunetheweb: Ah typescript! Must learn that…

Not a big deal, so fine to resolve this. Plus arguably a negative number IS a “good” score by the metric definitions 😄And the bug if that ever happens is upstream not in this function.

Alternatively could also return “unknown”.

tunetheweb (approved):
LGTM!

Tiggerito:
For my solution, I calculated the rating and returned an object like this:

return {
  rating: 'ni',
  name: 'Needs Improvement'
}

For my GA4 implementation, I sent both values as metric_rating and metric_status. The latter makes it easier to build reports.

Maybe over-engineering for most.

philipwalton:
Update: I changed `'ni'` to `'needs-improvement'` based on the recommendation of colleagues and to be consistent with the [recommendations for enumeration values](https://www.w3.org/TR/design-principles/#naming-consistency) section of the TAG's Design Principles guide.

Issue #185, opened 2021-10-06
v1.1.2 or later can crash apps on Opera Mini extreme data saving mode


jroebu14:
Hi,

We first reported a problem with v1.1.2 here https://github.com/GoogleChrome/web-vitals/issues/159 which we submitted a fix for.

However when we upgraded to v2.0.1, which contains the fix, we discovered the crash still happened on Opera Mini.

To mitigate this we resorted to [wrapping the performance metrics collection functions in a try/catch and logged out the error](https://github.com/bbc/psammead/pull/4573/files). This is the error we see - `Undefined variable: performance`.

It seems the Performance API is not supported at all in Opera Mini extreme data saving mode - https://caniuse.com/?search=performance

We get a moderate amount of traffic on this browser so it would be great if we could help expand support to this browser again. If supporting this browser is too tricky then consumers should be aware that they may have to handle errors coming from this package if they need to serve an audience that use Opera Mini.

Happy to help with this issue.

Thanks!

philipwalton:
Hmmm, I'm happy to add reasonable checks to avoid errors, but I can't actually reproduce this myself.

I've downloaded Opera Mini and turned on extreme data saver mode and I still see the global `performance` object. (Testing on `OPR/60.0.2254.59405`).

Do you know what version of Opera Mini you're seeing the error in?

jroebu14:
Brilliant! Thanks for this!

PR #247, opened 2022-07-23
Update naming of Metric object enums to be consistent

philipwalton:
This PR updates all enum values exposed on the `Metric` object to use naming conventions consistent with those recommended by the TAG's [Design Principles](https://www.w3.org/TR/design-principles/#naming-consistency) guide (i.e. kebab case).

This change updates two of the `navigationType` values, `'back_forward'` and `'back_forward_cache'`, and renames them to `'back-forward'` and `'back-forward-cache'` (changing them from snake case to kebab case). Note that this PR also includes other renames as well, but none of those names were from released versions, so they are not a use-visible change.

Note that technically this is a breaking change since the `navigationType` property was released in [v3.0.0-beta.0](https://github.com/GoogleChrome/web-vitals/blob/v3.0.0-beta.2/CHANGELOG.md#v300-beta0-2022-04-25), but it is only a breaking change for users of the v3 beta versions. It is not a breaking change for users upgrading from version 2 or older.

### DIFF
--- a/README.md
+++ b/README.md
@@ -779,14 +779,14 @@ The `LoadState` type is used in several of the metric [attribution objects](#att
  * State descriptions:
  * - `loading`: the initial document response has not yet been fully downloaded
  *   and parsed. This is equivalent to the corresponding `readyState` value.
- * - `interactive`: the document has been fully loaded and parsed, but
+ * - `dom-interactive`: the document has been fully loaded and parsed, but
  *   scripts may not have yet finished loading and executing.
- * - `content-loaded`: the document is fully loaded and parsed, and all
+ * - `dom-content-loaded`: the document is fully loaded and parsed, and all
  *   scripts (except `async` scripts) have loaded and finished executing.
  * - `complete`: the document and all of its sub-resources have finished
  *   loading. This is equivalent to the corresponding `readyState` value.
  */
-export type LoadState = 'loading' | 'interactive' | 'content-loaded' | 'complete';
+export type LoadState = 'loading' | 'dom-interactive' | 'dom-content-loaded' | 'complete';
 ```
 
 #### `FirstInputPolyfillEntry`
@@ -972,7 +972,7 @@ interface FCPAttribution {
   /**
    * The loading state of the document at the time when FCP `occurred (see
    * `LoadState` for details). Ideally, documents can paint before they finish
-   * loading (e.g. the `loading` or `interactive` phases).
+   * loading (e.g. the `loading` or `dom-interactive` phases).
    */
   loadState: LoadState,
   /**
@@ -1014,7 +1014,7 @@ interface FIDAttribution {
    * The loading state of the document at the time when the first interaction
    * occurred (see `LoadState` for details). If the first interaction occurred
    * while the document was loading and executing script (e.g. usually in the
-   * `interactive` phase) it can result in long input delays.
+   * `dom-interactive` phase) it can result in long input delays.
    */
   loadState: LoadState;
 }
@@ -1047,7 +1047,7 @@ interface INPAttribution {
    * The loading state of the document at the time when the even corresponding
    * to INP occurred (see `LoadState` for details). If the interaction occurred
    * while the document was loading and executing script (e.g. usually in the
-   * `interactive` phase) it can result in long delays.
+   * `dom-interactive` phase) it can result in long delays.
    */
   loadState?: LoadState;
 }
--- a/README.md
+++ b/README.md
@@ -779,14 +779,14 @@ The `LoadState` type is used in several of the metric [attribution objects](#att
  * State descriptions:
  * - `loading`: the initial document response has not yet been fully downloaded
  *   and parsed. This is equivalent to the corresponding `readyState` value.
- * - `domInteractive`: the document has been fully loaded and parsed, but
+ * - `interactive`: the document has been fully loaded and parsed, but
  *   scripts may not have yet finished loading and executing.
- * - `domContentLoaded`: the document is fully loaded and parsed, and all
+ * - `content-loaded`: the document is fully loaded and parsed, and all
  *   scripts (except `async` scripts) have loaded and finished executing.
- * - `loaded`: the document and all of its sub-resources have finished loading.
- *   This is equivalent to a document `readyState` of "complete".
+ * - `complete`: the document and all of its sub-resources have finished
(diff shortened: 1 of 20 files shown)


README.md
```diff
@@ -779,14 +779,14 @@ The `LoadState` type is used in several of the metric [attribution objects](#att
  * State descriptions:
  * - `loading`: the initial document response has not yet been fully downloaded
  *   and parsed. This is equivalent to the corresponding `readyState` value.
- * - `domInteractive`: the document has been fully loaded and parsed, but
+ * - `interactive`: the document has been fully loaded and parsed, but
  *   scripts may not have yet finished loading and executing.
- * - `domContentLoaded`: the document is fully loaded and parsed, and all
+ * - `content-loaded`: the document is fully loaded and parsed, and all
```
tunetheweb, line 784: This one seems an odd change to me. `domContentLoaded` is a well-known [event](https://developer.mozilla.org/en-US/docs/Web/API/Window/DOMContentLoaded_event) and it uses capitalisation we've used before: https://web.dev/critical-rendering-path-analyzing-crp/

src/types/base.ts
```diff
@@ -59,12 +59,13 @@ export interface Metric {
...
-   * by the Navigation Timing API (or `undefined` if the browser doesn't
+   * The type of navigation
+   *
+   * Navigation Timing API (or `undefined` if the browser doesn't
    * support that API). For pages that are restored from the bfcache, this
-   * value will be 'back_forward_cache'.
+   * value will be 'back-forward-cache'.
    */
-  navigationType:  NavigationTimingType | 'back_forward_cache' | 'prerender' | undefined;
+  navigationType:  'navigate' | 'reload' | 'back-forward' | 'back-forward-cache' | 'prerender';
```
tunetheweb, line 68: On line 64 you mention `undefined` but have remove that from here?

README.md
philipwalton: > `domContentLoaded` is a well-known [event](https://developer.mozilla.org/en-US/docs/Web/API/Window/DOMContentLoaded_event)

The event is actually `DOMContentLoaded` :) but either way, note that this API is referring to a state (the timespan between `domContentLoadedEventStart` and `domComplete` in the [Navigation Timing API](https://w3c.github.io/navigation-timing/#sec-PerformanceNavigationTiming)) and not the DOM event itself, so I don't think it's important to be consistent with the event name.

Also the Design Principles doc [specifically mentions cases](https://www.w3.org/TR/design-principles/#naming-consistency:~:text=Note%20that%20in%20particular%2C%20when%20a%20HTML%20attribute%20is%20reflected%20as%20a%20property%2C%20the%20attribute%20and%20property%E2%80%99s%20casings%20won%E2%80%99t%20necessarily%20match.%20For%20example%2C%20the%20HTML%20attribute%20ismap%20on%20img%20elements%20is%20reflected%20as%20the%20isMap%20property%20on%20HTMLImageElement.) where the same term may have different casing based on how it's used, and you can see an example of that with the differences between `DOMContentLoaded` (the event name) and `domContentLoadedEventStart` (the Navigation Timing interface property name).

BTW, I did go back and forth as to whether these names should be `'dom-interactive'` and `'dom-content-loaded'` or `'interactive'` and `'content-loaded'` (and I'm still torn). Using the name `'interactive'` is good because it matches the the [`document.readyState`](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState) value, but on the other hand it's not a 1-to-1 match with the "interactive" readyState value, so maybe that warrants a different name to help differentiate. WDYT?

src/types/base.ts
philipwalton: Correct, I updated the code so the `undefined` would no longer be possible.

Note, I don't think it was ever technically possible before, but according to the type logic is was possible since `getNavigationEntry()` can return `undefined` in cases where the browser doesn't support any performance APIs at all (see #185). However, browsers that do not support any performance APIs should not be able to get this far in the code to ever fire these callbacks (all `onXXX()` functions are gated on _some_ sort of performance API support), which is why I say I don't think it's technically possible, at least not in any  browser I'm aware of.

I made the change because I thought it was much riskier to have the type definition advertise that it is possible to get `undefined` here, and I didn't want developers to be coding defensively for something that I don't think can happen (and now definitely can't because the default is `'navigate'`).

README.md
tunetheweb: Yeah I recognised it wasn’t exactly the event itself (hence why I was ok with different capitalisation more inline with the Navigation timing API.

I still think mentioning DOM in here somewhere (`dom-content-loaded`?) is better as a description of what it is. Just `content-loaded` seems very generic. And maybe the use of hyphens will make it more obvious it’s a state rather than the event or timing?

And I’m similarly torn between matching readystate and that being a good thing, or liable for confusion. I dunno. Naming things is hard…

src/onTTFB.ts
```diff
@@ -82,12 +82,16 @@ export const onTTFB = (onReport: ReportCallback, opts?: ReportOpts) => {
...
 
       report(true);
-    }
-  });
 
-  onBFCacheRestore(() => {
-    metric = initMetric('TTFB', 0);
-    report = bindReporter(onReport, metric, thresholds, opts!.reportAllChanges);
-    report(true);
+      // Only report TTFB after bfcache restores if a `navigation` entry
```
philipwalton, line 86: @tunetheweb FYI: I moved this `onBFCacheRestore()` call inside the check for the navigation entry to ensure that a TTFB metric wasn't reported after a bfcache restore in cases where the browser doesn't support the Navigation Timing API at all.

This still wouldn't affect cases where `navigationType` was undefined because in bfcache restores the type is set to `back-forward-cache`, but I do think it was a bug since browsers like Opera mini in extreme data saver mode would have run into this.

I checked all the other functions to, but they were all fine since they're gated on support for `PerformanceObserver`.

README.md
philipwalton: I'd be fine with this:

```ts
type LoadState = 'loading' | 'dom-interactive' | 'dom-content-loaded' | 'complete';
```

So `loading` and `complete` are the same as their `readyState` versions, and `dom-interactive` and `dom-content-loaded` are the sub-phases of `readyState === 'interactive'`. Let me know what you think.

tunetheweb: Works for me.

philipwalton: Done.

tunetheweb (approved):

PR #248, opened 2022-07-24
Move the Navigation Timing API polyfill to the base+polyfill build

philipwalton:
This PR removes the Navigation Timing polyfill from the "standard" build and adds it to the "base+polyfill" build. This reduces the total size of the "standard" build by about 5% (brotli'd).

When this library was first released, Safari didn't support the Navigation Timing API (Level 2), so this was a lightweight and easy way to get the TTFB metric in all modern browsers. But Safari has now supported this API since version 15, which has been out for almost a year.

I think this library should aim to support the latest stable version(s) of all modern browsers when reasonable, but given that this is a measurement library and not a feature library, I don't think full browser support should be a goal.

Also, this change primarily affects the TTFB metric, which is largely a measure of network/device capabilities rather than browser capabilities, so supporting only the latest versions of Safari seems reasonable for optimization purposes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -643,9 +643,9 @@ For guidance on how to collect and use real-user data to debug performance issue
 
 _**⚠️ Warning ⚠️** the "base+polyfill" build is deprecated. See [#238](https://github.com/GoogleChrome/web-vitals/issues/238) for details._
 
-The `polyfill.js` script adds event listeners (to track FID cross-browser), and it records initial page visibility state as well as the timestamp of the first visibility change to hidden (to improve the accuracy of CLS, FCP, LCP, and FID).
+The `polyfill.js` script adds event listeners (to track FID cross-browser), and it records initial page visibility state as well as the timestamp of the first visibility change to hidden (to improve the accuracy of CLS, FCP, LCP, and FID). It also polyfills the [Navigation Timing API Level 2](https://www.w3.org/TR/navigation-timing-2/) in browsers that only support the original (now deprecated) [Navigation Timing API](https://www.w3.org/TR/navigation-timing/).
 
-In order for it to work properly, the script must be the first script added to the page, and it must run before the browser renders any content to the screen. This is why it needs to be added to the `<head>` of the document.
+In order for the polyfill to work properly, the script must be the first script added to the page, and it must run before the browser renders any content to the screen. This is why it needs to be added to the `<head>` of the document.
 
 The "standard" build of the `web-vitals` library includes some of the same logic found in `polyfill.js`. To avoid duplicating that code when using the "base+polyfill" build, the `web-vitals.base.js` bundle does not include any polyfill logic, instead it coordinates with the code in `polyfill.js`, which is why the two scripts must be used together.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -643,9 +643,9 @@ For guidance on how to collect and use real-user data to debug performance issue
 
 _**⚠️ Warning ⚠️** the "base+polyfill" build is deprecated. See [#238](https://github.com/GoogleChrome/web-vitals/issues/238) for details._
```
tunetheweb, line 644: It seems kinda weird to be adding things to the polyfill, while also deprecating it in the next release.

But guess it's still "supported" for now but deprecated.

README.md
```diff
@@ -643,9 +643,9 @@ For guidance on how to collect and use real-user data to debug performance issue
...
 
-In order for it to work properly, the script must be the first script added to the page, and it must run before the browser renders any content to the screen. This is why it needs to be added to the `<head>` of the document.
+In order for the polyfill to work properly, the script must be the first script added to the page, and it must run before the browser renders any content to the screen. This is why it needs to be added to the `<head>` of the document.
 
 The "standard" build of the `web-vitals` library includes some of the same logic found in `polyfill.js`. To avoid duplicating that code when using the "base+polyfill" build, the `web-vitals.base.js` bundle does not include any polyfill logic, instead it coordinates with the code in `polyfill.js`, which is why the two scripts must be used together.
```
tunetheweb, line 650: Is this still the case with this move?

README.md
```diff
@@ -1145,12 +1147,12 @@ The `web-vitals` code has been tested and will run without error in all major br
 
 Browser support for each function is as follows:
 
-- `onCLS()`: Chromium,
-- `onFCP()`: Chromium, Firefox, Safari
-- `onFID()`: Chromium, Firefox, Safari, Internet Explorer (with the [polyfill](#how-to-use-the-polyfill))
+- `onCLS()`: Chromium
+- `onFCP()`: Chromium, Firefox, Safari 14.1+
```
tunetheweb, line 1151: Separate to this PR, but is it worth noting somewhere in the docs (either here or elsewhere) that FCP on Safari is quite different: https://blog.webpagetest.org/posts/why-first-contentful-paint-doesnt-work-as-a-cross-browser-metric/

Happy to open a separate issue for this.

README.md
philipwalton: I agree it's a bit weird.

In this case I was thinking of it as "deprecating this feature", and an easy way to do that was to move it to the polyfill build. But yeah, from a certain perspective it is making changes to a deprecated feature.

The good thing in this case is this is a polyfill, so it shouldn't cause any breakage for anyone's site. They'll just stop getting TTFB reports for IE and Safari 8-14.

README.md
philipwalton: It's still true for FID. TTFB would work even if the polyfill script weren't added to the `<head>`, but FID wouldn't work at all.

README.md
philipwalton: We should probably discuss whether it's worth pointing this out in this library documentation or in our metric documentation on web.dev. Personally, I think I'm leaning toward the latter.

tunetheweb: Agree it should be in docs, but given this library specifically states it can be used for tracking across these browsers a think it should be linked from here. The complication with that is our docs are per metric so don’t have one place to link. But to be honest FCP is the one that I’m aware of that is the most different.

tunetheweb (approved):

README.md
tunetheweb: Discussed offline. Might be an upcoming post that would be good place to link to. Happy to leave for now.

PR #252, opened 2022-07-30
chore:size limit checks

ittaibaratz:
I am working on another PR for a performance innovation project, and to understand the impact on the bundle size I've added size checks.

I figured it makes a nice contribution to monitor this bundle's size as more functionality is added over time, so here's a tiny PR to add size checks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .DS_Store
 node_modules
+.idea
 
 # Log files
 *.log
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
         "@babel/core": "^7.16.10",
         "@babel/preset-env": "^7.16.11",
         "@rollup/plugin-replace": "^3.0.1",
+        "@size-limit/file": "^6.0.4",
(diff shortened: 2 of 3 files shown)


ittaibaratz:
retest

philipwalton:
Hi @ittaibaratz, I appreciate the thought, and in general I think this package looks useful, but I don't think it's a good fit for this library.

For one thing, this library is quite modular, and each metric is exposed as a separate module, so there isn't really such a thing as a "bundle size", and I wouldn't want to put arbitrary limits on what metrics could be added just to keep the total size of all metric modules below some threshold.

Second, I already obsessively check the size of the library whenever I add new features, and I prefer to measure the size based on brotli compression rather than gzip, and it looks like this package only supports gzip at the moment.

I do think it's useful to print the final, compressed size of the files in `/dist` after a re-build, but I wouldn't want to do anything beyond that at this point.

ittaibaratz:
Thanks @philipwalton , I will close this PR based on your feedback - Really I wanted a way to measure impact of another PR I am doing which I will raise once I figure the CLA stuff with our open source office.

PR #253, opened 2022-08-03
Make web-vitals usable across frames (WIP)

ittaibaratz:
Hi @philipwalton , this PR is still WIP. Our use case might be niche so I want to get your POV before moving further.

**Background:** 

Our main app embeds other apps using iframes. In cases where the iframe use the same origin as the parent, we want to avoid loading our runtime script (Which include web-vitals) twice in order to improve the iframe load performance.

Instead, we want the iframe to reach out to the parent and execute our script from there, which also calls webvitals when we measure & log load times.

**The problem:**

Web vitals in its current form interacts with the objects in the browser's global scope - window, document, etc.
This is a problem for our use case since we want to use the script across same origin iframes.

The changes I am purposing will allow webvitals to run against any frame in a backward compatible manner, i.e.
```
getFID = (onReport: ReportHandler, reportAllChanges?: boolean, win = window)
```

I realize our use case might be niche and we'd better just do a fork, but wanted to check your POV. Also I wonder if there's a different way to meet our needs which does not require this type of rewrite?

Thanks!
Ittai

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -53,7 +53,17 @@
         "node/no-unsupported-features/es-syntax": "off",
         "node/no-missing-require": "off",
         "node/shebang": "off",
-        "no-dupe-class-members": "off"
+        "no-dupe-class-members": "off",
+        "no-restricted-globals": ["ERROR",
+          "clearTimeout",
+          "document",
+          "addEventListener",
+          "removeEventListener",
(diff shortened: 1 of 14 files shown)


tunetheweb:
Closing this as quite out of date now.

PR #254, opened 2022-08-15
chore: remove reportAllChanges from getTTFB

suyalcinkaya:
`reportAllChanges` is not a valid param for `getTTFB`. Hence, this PR removes it from the markdown.


suyalcinkaya:
Closing this since it's not valid anymore.

PR #262, opened 2022-09-24
Fix Types of Attribution

narui1125:
Hello, There may be a problem with types of attributtion.

As an example, a type error occurs in the following code in call `onCLS`.
```ts
import { CLSMetricWithAttribution, MetricWithAttribution, onCLS } from 'web-vitals/attribution';

const sendCLSMetrics = (metric: CLSMetricWithAttribution) => {
    console.log(metric.attribution)
}

onCLS(sendCLSMetrics); // type error

const sendMetrics = (metric: MetricWithAttribution) => {
    console.log(metric.attribution)
}

onCLS(sendMetrics); // type error
```

I think there is a problem with extending callback function. Sorry if there is any problem with my usage.

### DIFF
--- a/src/deprecated.ts
+++ b/src/deprecated.ts
@@ -55,10 +55,3 @@ export {
    */
   onTTFB as getTTFB,
 } from './onTTFB.js';
-
-export {
-  /**
-   * @deprecated Use `ReportCallback` instead.
-   */
-  ReportCallback as ReportHandler,
-} from './types.js';
--- a/src/types/base.ts
+++ b/src/types/base.ts
@@ -14,7 +14,13 @@
  * limitations under the License.
  */
 
+import { CLSAttribution } from './cls.js';
+import { FCPAttribution } from './fcp.js';
+import { FIDAttribution } from './fid.js';
+import { INPAttribution } from './inp.js';
+import { LCPAttribution } from './lcp.js';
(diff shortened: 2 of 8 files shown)


tunetheweb:
As per the README you should be importing it like this:

```js
import {onCLS } from 'web-vitals/attribution';

const sendCLSMetrics = (attribution) => {
    console.log(attribution)
}
``

**Comment by narui1125** [comment]
Thank you for your response.

It seems that the code you provided will work, but there might be an problem when adding the type.(I would like to add type annotations for code completion)

'onCLS' accepts 'CLSReportCallbackWithAttribution', but 'CLSReportCallbackWithAttribution' can also accept not only 'CLSMetricWithAttribution' but also 'CLSMetric' and 'Metric' types,  so the 'attribution' property is not guaranteed to exist on the type.

It seems like the following approach would work, but I think 'CLSReportCallbackWithAttribution' should only accept 'CLSMetricWithAttribution'.

```ts
import { onCLS, CLSReportCallbackWithAttribution } from 'web-vitals/attribution';

const sendCLSMetrics: CLSReportCallbackWithAttribution = (param) => {
  if (!('attribution' in param)) {
    return
  }

  console.log(param.attribution)
}

onCLS(sendCLSMetrics);
```

スクリーンショット 2023-04-09 15 41 52

tunetheweb:
I believe this has been fixed in latest main.

PR #265, opened 2022-10-03
Ensure attribution object is always present in attribution build

### LINKED ISSUE #264: attribution property can be undefined
Using the attribution build, I was seeing `Uncaught TypeError: Cannot read property 'largestShiftTarget' of undefined` errors in our client logs. Unfortunately this means we missed tracking some CLS events until we could deploy a fix with a truthy check for `attribution`.

Our usage is very similar to the [documentation](https://github.com/GoogleChrome/web-vitals#send-attribution-data), which will also run into this error because there is no truthy check for `attribution`.

It seems that either:
a) The TS typings and documentation should be updated to take into account `attribution` possibly being undefined
or
b) The code should be updated so attribution is always at least an empty object

philipwalton:
This PR fixes #264 by ensuring an attribution object is always present on the `Metric` object when using the attribution build. This update matches what is currently documented and expressed by the type definitions.


src/attribution/onCLS.ts
```diff
@@ -42,11 +42,12 @@ const attributeCLS = (metric: CLSMetric): void => {
...
           loadState: getLoadState(largestEntry.startTime),
         };
+        return;
       }
     }
-  } else {
-    (metric as CLSMetricWithAttribution).attribution = {};
   }
+  // Set an empty object if no other attribution has been set.
+  (metric as CLSMetricWithAttribution).attribution = {};
```
tunetheweb, line 50: This works, but took me a while to grok.

It's not so much "if no other attribution has been set" as, "if not already returned previously with an attribution".

Alternatives:
1. Initialise the empty attribution at the top, and override it where necessary (rather than set it at the end... sometimes).
2. Set it with something like this:

```js
(metric as CLSMetricWithAttribution).attribution = (metric as CLSMetricWithAttribution).attribution || {};
```

Think either of those would be clearer? Or think it's clear enough to stick with what you have?

philipwalton: I agree that initializing it at the top is a bit easier to read. However, I didn't really want to do that because I don't like that idea of creating an empty object just to not use it (which will be the case 99% of the time).

This may be a micro-optimization, but given that this is third-party code I thought it would be better to optimize for performance rather than clarify. If this were my own code I'd probably do it the way you suggest.

tunetheweb: Fair point. It actually looks easier in the code - just took me a moment when looking at the git diff to get what you were doing.

tunetheweb (approved):

PR #270, opened 2022-10-18
Clamp LCP and FCP to 0 for prerendered pages

### LINKED ISSUE #268: onLCP does not work on prerendered pages
Steps to reproduce:

- I'm using Chrome 108.0.5355.0 (Official Build) dev (64-bit) ; maybe be sure to be at least at that revision.
- Enable Prerender2 in chrome://flags. Not sure this is necessary anymore but let's just be sure.
- Visit https://equatorial-longing-kick.glitch.me/ .
- Click "This link should be prerendered"
- Notice that the page was prerendered (`document.prerendering` was logged as true), but no LCP is logged.
- Open a new tab, directly to https://equatorial-longing-kick.glitch.me/next.html .
- Notice that this time we get LCP logged.

So it seems like if the page is loaded while prerendered, onLCP is never called. Maybe this is related to the README saying

> CLS, FCP, FID, and LCP are not reported if the page was loaded in the background.

I noticed that there is some special handling for bfcache restore in https://github.com/GoogleChrome/web-vitals/blob/main/src/onLCP.ts#L92-L106 . Maybe there should be something similar for prerendering activation? (Prerendering activation can be detected using the `prerenderingchange` event on `Document`.)

tunetheweb:
Fixes #268 

Not sure how to add a test for this, since the prerendered test (http://localhost:9090/test/lcp?prerender=1) is just a stub and not an actual preprendered page so LCP is not zero.

### DIFF
--- a/src/onFCP.ts
+++ b/src/onFCP.ts
@@ -50,8 +50,9 @@ export const onFCP = (onReport: FCPReportCallback, opts?: ReportOpts) => {
         if (entry.startTime < visibilityWatcher.firstHiddenTime) {
           // The activationStart reference is used because FCP should be
           // relative to page activation rather than navigation start if the
-          // page was prerendered.
-          metric.value = entry.startTime - getActivationStart();
+          // page was prerendered. But in cases where `activationStart` occurs
+          // after the FCP, this time should be clamped at 0.
+          metric.value = Math.max(entry.startTime - getActivationStart(), 0);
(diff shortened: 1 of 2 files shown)


tunetheweb:
Note this will clamp ALL FCP and LCP values, not just prerendered ones. But I'm not sure that's such a bad thing?

PR #278, opened 2022-11-13
Defer execution of observer callbacks

### LINKED ISSUE #277: PerformanceObserver is not disconnected after paint in Safari
The current `onFCP()` has a conditional to check for the existence of the `PerformanceObserver` variable before disconnecting:

https://github.com/GoogleChrome/web-vitals/blob/7f0ed0bfb03c356e348a558a3eda111b498a2a11/src/onFCP.ts#L45-L47

This was added due to an issue where the `po` object might not be defined if the callback is invoked right away; however, this turns out to be [a bug in Safari](https://bugs.webkit.org/show_bug.cgi?id=247863) rather than expected behavior.

A better solution than checking if the observer variable is set would be to always defer invocation of the callback to match the spec. This would ensure the observer can always be disconnected in all browsers after its no longer needed.

philipwalton:
This PR fixes #277 by deferring execution of all Performance Observer callbacks by a microtask, which will allow the current call stack to complete and ensure all module-scoped variables are available within the callback itself. This is needed to address an interop issue with Safari.

This PR also removes some related logic that was only needed to address compat issues in older versions of Safari. Since this logic is not needed in the last two major versions of Safari, it's safe to remove from this library.

### DIFF
--- a/src/lib/observe.ts
+++ b/src/lib/observe.ts
@@ -43,7 +43,12 @@ export const observe = <K extends keyof PerformanceEntryMap>(
   try {
     if (PerformanceObserver.supportedEntryTypes.includes(type)) {
       const po = new PerformanceObserver((list) => {
-        callback(list.getEntries() as PerformanceEntryMap[K]);
+        // Delay by a microtask to workaround a bug in Safari where the
+        // callback is invoked immediately, rather than in a separate task.
+        // See: https://github.com/GoogleChrome/web-vitals/issues/271
+        Promise.resolve().then(() => {
+          callback(list.getEntries() as PerformanceEntryMap[K]);
+        });
       });
       po.observe(Object.assign({
         type,
--- a/src/lib/observe.ts
+++ b/src/lib/observe.ts
@@ -45,7 +45,7 @@ export const observe = <K extends keyof PerformanceEntryMap>(
       const po = new PerformanceObserver((list) => {
         // Delay by a microtask to workaround a bug in Safari where the
         // callback is invoked immediately, rather than in a separate task.
-        // See: https://github.com/GoogleChrome/web-vitals/issues/271
+        // See: https://github.com/GoogleChrome/web-vitals/issues/277
         Promise.resolve().then(() => {
           callback(list.getEntries() as PerformanceEntryMap[K]);
         });
--- a/src/onFCP.ts
+++ b/src/onFCP.ts
@@ -42,9 +42,7 @@ export const onFCP = (onReport: FCPReportCallback, opts?: ReportOpts) => {
   const handleEntries = (entries: FCPMetric['entries']) => {
     (entries as PerformancePaintTiming[]).forEach((entry) => {
       if (entry.name === 'first-contentful-paint') {
-        if (po) {
-          po.disconnect();
-        }
+        po!.disconnect();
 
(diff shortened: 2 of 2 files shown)


src/onFCP.ts
```diff
@@ -42,9 +42,7 @@ export const onFCP = (onReport: FCPReportCallback, opts?: ReportOpts) => {
   const handleEntries = (entries: FCPMetric['entries']) => {
     (entries as PerformancePaintTiming[]).forEach((entry) => {
       if (entry.name === 'first-contentful-paint') {
-        if (po) {
-          po.disconnect();
-        }
+        po!.disconnect();
```
tunetheweb, line 45: You'd think `po` must exist by the time you get to here (otherwise how was it called!), but clearly that wasn't the case previously so what if there are other bugs? Should we just optionally chain here instead just to be sure?

```suggestion
        po?.disconnect();
```

src/lib/observe.ts
```diff
@@ -43,7 +43,12 @@ export const observe = <K extends keyof PerformanceEntryMap>(
   try {
     if (PerformanceObserver.supportedEntryTypes.includes(type)) {
       const po = new PerformanceObserver((list) => {
-        callback(list.getEntries() as PerformanceEntryMap[K]);
+        // Delay by a microtask to workaround a bug in Safari where the
+        // callback is invoked immediately, rather than in a separate task.
+        // See: https://github.com/GoogleChrome/web-vitals/issues/271
```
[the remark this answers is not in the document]
philipwalton: Ahh, good catch, thanks!

src/onFCP.ts
philipwalton: I don't think we need to do that. The fact that `po` wasn't defined here was caught by the tests, so I'm not concerned that there are other bugs lurking that we don't know about.

Note that the reason it was in a conditional before had nothing to do with this bug, it had to do with the fact that, in Safari, we used performance timeline rather than performance observer to get FCP. As soon as I refactored to use performance observer, the tests started failing in Safari, making this bug apparent.

tunetheweb: > Note that the reason it was in a conditional before had nothing to do with this bug…

Yes, that confused me. Gotcha now.

src/onFCP.ts
```diff
@@ -60,23 +58,11 @@ export const onFCP = (onReport: FCPReportCallback, opts?: ReportOpts) => {
     });
   };
 
-  // TODO(philipwalton): remove the use of `fcpEntry` once this bug is fixed.
-  // https://bugs.webkit.org/show_bug.cgi?id=225305
-  // The check for `getEntriesByName` is needed to support Opera:
-  // https://github.com/GoogleChrome/web-vitals/issues/159
-  // The check for `window.performance` is needed to support Opera mini:
-  // https://github.com/GoogleChrome/web-vitals/issues/185
```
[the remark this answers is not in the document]
philipwalton: I'm sure older versions are hanging around, but I have no concerns about that.

Worst case scenario is that some versions of Safari won't report FCP, and I think that's okay. I would much rather keep the bundle size as small as possible for users on current browsers than continue to maintain support for legacy browsers that we don't even know if users are optimizing for.

tunetheweb (approved):

PR #281, opened 2022-11-14
Ignore TTFB for loads where responseStart is zero

### LINKED ISSUE #275: TTFB (under)reporting 0ms on Safari + utm/gclid params
I talked with Barry (tunetheweb) during performance.now conference help desk about the TTFB reporting 0ms under these conditions.

This can be seen in the following attachments as well, showing RUM data (I've got more domains if needed).

mobile P80 data of the last 7 days @ [redacted] (soon to be replatformed, date unknown)
safari-querystrings-ttfb-custom

mobile P80 data of the last 28 days @  [redacted]:
safari-querystrings-ttfb-m2-2

As we started using the attribution build recently, I'm able to say that this happens both with and without the attribution build (both classic script).

Safari at the top, Google Chrome at the bottom. So:
- 0ms at high percentiles for Safari while Chrome is showing totally different data 
- only happens with ad/campaign related query string parameters. 
- both for unique and returning visitors
- Gclid is typically throwing a redirect in the mix, but utm-parameters don't always do this, so I'm clueless here (Rick (rviscomi) asked me to file an issue already, but could ask an Apple-minded colleague to test it as well once in the office).

philipwalton:
This PR fixes #275 by ignoring loads where the `responseStart` property of the navigation entry is 0. The assumption is that `responseStart` should never be zero unless the browser is intentionally concealing navigation timing data.

### DIFF
--- a/src/onTTFB.ts
+++ b/src/onTTFB.ts
@@ -67,21 +67,28 @@ export const onTTFB = (onReport: ReportCallback, opts?: ReportOpts) => {
     const navEntry = getNavigationEntry();
 
     if (navEntry) {
-      // The activationStart reference is used because TTFB should be
-      // relative to page activation rather than navigation start if the
-      // page was prerendered. But in cases where `activationStart` occurs
-      // after the first byte is received, this time should be clamped at 0.
-      metric.value = Math.max(navEntry.responseStart - getActivationStart(), 0);
+      const responseStart = navEntry.responseStart;
 
       // In some cases the value reported is negative or is larger
       // than the current page time. Ignore these cases:
       // https://github.com/GoogleChrome/web-vitals/issues/137
       // https://github.com/GoogleChrome/web-vitals/issues/162
-      if (metric.value < 0 || metric.value > performance.now()) return;
+      if (responseStart < 0 || responseStart > performance.now()) return;
 
-      metric.entries = [navEntry];
+      // If the navigation entry's `responseStart` value is 0, ignore it.
+      // This likely means the request included a cross-origin redirect, and
+      // the browser has removed timing info for privacy/security reasons.
+      // See: https://github.com/GoogleChrome/web-vitals/issues/275
+      if (responseStart > 0) {
+        // The activationStart reference is used because TTFB should be
+        // relative to page activation rather than navigation start if the
+        // page was prerendered. But in cases where `activationStart` occurs
+        // after the first byte is received, this time should be clamped at 0.
+        metric.value = Math.max(responseStart - getActivationStart(), 0);
 
-      report(true);
+        metric.entries = [navEntry];
+        report(true);
+      }
 
       // Only report TTFB after bfcache restores if a `navigation` entry
       // was reported for the initial load.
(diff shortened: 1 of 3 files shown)


src/onTTFB.ts
```diff
@@ -67,21 +67,28 @@ export const onTTFB = (onReport: ReportCallback, opts?: ReportOpts) => {
...
-      // after the first byte is received, this time should be clamped at 0.
-      metric.value = Math.max(navEntry.responseStart - getActivationStart(), 0);
+      const responseStart = navEntry.responseStart;
 
       // In some cases the value reported is negative or is larger
       // than the current page time. Ignore these cases:
       // https://github.com/GoogleChrome/web-vitals/issues/137
       // https://github.com/GoogleChrome/web-vitals/issues/162
-      if (metric.value < 0 || metric.value > performance.now()) return;
+      if (responseStart < 0 || responseStart > performance.now()) return;
```
philipwalton, line 76: Note, I revert this to using `responseStart` instead of `metric.value` because `metric.value` is being clamped at 0 above, which makes this check mostly useless now. Using `responseStart` reverts the logic to what it was originally trying to check.

src/onTTFB.ts
```diff
@@ -67,21 +67,28 @@ export const onTTFB = (onReport: ReportCallback, opts?: ReportOpts) => {
...
 
-      metric.entries = [navEntry];
+      // If the navigation entry's `responseStart` value is 0, ignore it.
+      // This likely means the request included a cross-origin redirect, and
+      // the browser has removed timing info for privacy/security reasons.
+      // See: https://github.com/GoogleChrome/web-vitals/issues/275
+      if (responseStart > 0) {
```
tunetheweb, line 82: Why not change the test above to below to capture this case and avoid the need of this test?:

```js
if (responseStart <= 0 ||  || responseStart > performance.now()) return;
```

That wouldn't set up the onBFCacheRestore, but think that's OK since that is only called if a  `navigation` entry was reported for the initial load.

philipwalton: Yeah, fair point. I did this to ensure the logic ran for bfcache restores, but I suppose it's an open question as to whether or not it should. Also, if it should then it arguably should for the above cases as well, so the early return isn't needed.

WDYT? Should we report zero values for bfcache restores in cases where the original TTFB value was never reported? Or should we just ignore those visits altogether?

I guess one argument for not reporting them is that bugs in Safari might make it look to the site owner that bfcache hit rates are much higher in Safari than in other browsers, when if fact that may not be the case.

tunetheweb: Hmmm… there’s argument for both sides.

I think I’m veering slightly on the don’t report if original nav didn’t have it.

Either way think it’s confusing at present with the comment below saying we don’t report in those cases if we do sometimes.

philipwalton: Ok, I'll update to not report bfcache restores in these cases.

philipwalton:
Ok, updated and rebased to pick up the latest changes on `main`.

tunetheweb (approved):

PR #282, opened 2022-11-15
Defer all observers until after activation

### LINKED ISSUE #271: Observer callbacks should be deferred until after activation
For prerendered pages, in Chrome's current implementation it will always defer running any Performance Observer callbacks until after the page has been activated.

However, I'm not sure if we can guarantee that will always be the case (spec changes, browser bugs, etc.), so I think it makes sense to defer handling any Performance Observer callbacks until after activation.

If callbacks are handled prior to activation then `activationStart` will always be 0, and thus any calculations made using it will be incorrect.

philipwalton:
This PR fixes #271 by adding a new `whenActivated` utility and using it to defer all of the `onXXX()` metric functions. It also tweaks the `getVisibilityWatcher()` logic to better handle prerendering cases (which always happens when the `visibilityState` is "hidden", and are thus tricky to get right).

### DIFF
--- a/src/lib/getVisibilityWatcher.ts
+++ b/src/lib/getVisibilityWatcher.ts
@@ -15,23 +15,50 @@
  */
 
 import {onBFCacheRestore} from './bfcache.js';
-import {onHidden} from './onHidden.js';
+
 
 let firstHiddenTime = -1;
 
 const initHiddenTime = () => {
-  // If the document is hidden and not prerendering, assume it was always
-  // hidden and the page was loaded in the background.
+  // If the document is hidden when this code runs, assume it was always
+  // hidden and the page was loaded in the background, with the one exception
+  // that visibility state is always 'hidden' during prerendering, so we have
+  // to ignore that case until prerendering finishes (see: `prerenderingchange`
+  // event logic below).
   return document.visibilityState === 'hidden' &&
       !document.prerendering ? 0 : Infinity;
 }
 
-const trackChanges = () => {
-  // Update the time if/when the document becomes hidden.
-  onHidden(({timeStamp}) => {
-    firstHiddenTime = timeStamp
-  }, true);
+const onVisibilityUpdate = (event: Event) => {
+  // If the document is 'hidden' and no previous hidden timestamp has been
+  // set, update it based on the current event data.
+  if (document.visibilityState === 'hidden' && firstHiddenTime > -1) {
+    // If the event is a 'visibilitychange' event, it means the page was
+    // visible prior to this change, so the event timestamp is the first
+    // hidden time. However, if the event is a 'prerenderingchange' event and
+    // the document is 'hidden', assume the tab was activated in a background
+    // state and has always been hidden.
+    firstHiddenTime = event.type === 'visibilitychange' ? event.timeStamp : 0;
+
+    // Remove all listeners now that a `firstHiddenTime` value has been set.
+    removeChangeListeners();
+  }
+}
+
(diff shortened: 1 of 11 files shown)


src/lib/getVisibilityWatcher.ts
```diff
@@ -15,24 +15,51 @@
...
+const onVisibilityUpdate = (event: Event) => {
+  // If the document is 'hidden' and no previous hidden timestamp has been
+  // set, update it based on the current event data.
+  if (document.visibilityState === 'hidden' && firstHiddenTime > -1) {
+    // If the event is a 'visibilitychange' event, it means the page was
+    // visible prior to this change, so the event timestamp is the first
+    // hidden time. However, if the event is a 'prerenderingchange' event and
+    // the document is 'hidden', assume the tab was activated in a background
+    // state and has always been hidden.
+    firstHiddenTime = event.type === 'visibilitychange' ? event.timeStamp : 0;
```
tunetheweb, line 41: Took me several attempts to understand this and not sure I fully get it.

I tested and on prerender2 pages the `visibilityState` event is sent first setting it to visible, then the `prerenderingchange` event. I'm not sure that's guaranteed and there was some talk on buganizer 223315889 as to whether that was the correct way, but that's what's happening now.

So this means this code will effectively never run for prerender pages (unless it becomes hidden later, after prerendering) so not sure why you need this comment on "'prerenderingchange' event", nor why you need the 'prerenderingchange' event listener to call this function?

philipwalton: > I tested and on prerender2 pages the `visibilityState` event is sent first setting it to visible, then the `prerenderingchange` event. I'm not sure that's guaranteed...

It's guaranteed by spec currently, so if that were to ever change it would require a spec change. IMO this is definitely the correct order because if the order were revered it would be impossible to determine whether a currently-prerendering page is in a foreground or background tab once the page activates.

> So this means this code will effectively never run for prerender pages (unless it becomes hidden later, after prerendering) so not sure why you need this comment on "'prerenderingchange' event", nor why you need the 'prerenderingchange' event listener to call this function?

Here's the case I'm trying to catch with this logic:

1. Page is prerendered, which means its `visibilityState` is "hidden"
2. Page gets activated in a background tab, and the `prerenderingchange` event fires
3. Since the page is activated in a background tab, no `visibilitychange` event fires and the visibility state does not change (i.e. it was "hidden" before and it remains "hidden" afterward).

In the above case, the only way to determine the "real" visibility state of the page is to wait until prerendering is done and then check the value of `document.visibilityState`.

Does that make sense?

tunetheweb: OK so that's not currently possible with Prerender, which is why I was confused, but agree it's being looked at.

> Here's the case I'm trying to catch with this logic

OK, suggest a change to the comment then,

```suggestion
  if (document.visibilityState === 'hidden' && firstHiddenTime > -1) {
    // If the event is a 'visibilitychange' event, it means the page was
    // visible prior to this change, so the event timestamp is the first
    // hidden time.
    // However, if the event is a 'visibilitychange' event, then it must be a
    // 'prerenderingchange' event and the fact that the document is still
    // 'hidden' from above check, means we can assume the tab was activated
    // in a background state and so has always been hidden.
    firstHiddenTime = event.type === 'visibilitychange' ? event.timeStamp : 0;
```

philipwalton:
@tunetheweb committed your suggestion. LGTY otherwise?

tunetheweb (approved):

PR #283, opened 2022-11-15
Report initial CLS value when reportAllChanges is true

### LINKED ISSUE #279: There is no way to early-report zero-value CLS scores
If the `reportAllChanges` option is set to `true`, every metric exposed by this library will early report _some_ value—except for CLS for pages where there are no layout shifts. The implications of this is it's more likely that a page with no CLS will fail to report any value in cases where unload or background detection is unreliable.

For sites that would like to eagerly report metric values and then update them if/when new values are reported, it should be possible to eagerly report zero-CLS.

Note that it can't be assumed that a page that didn't report any CLS had zero CLS, since pages that don't report FCP also do not report CLS. To account for this, I think it makes sense to report a zero-value CLS as soon as FCP is reached.

philipwalton:
This PR fixes #279 by ensuring that an initial CLS value of 0 is reported as soon as the page reaches FCP. For pages that don't set `reportAllChanges` to true, they shouldn't notice any change; however, for pages that do set `reportAllChanges` to true, they may see additional reports come in.

Note, I do not think this qualifies as a breaking change, because developers should not have been assuming a certain number of reports will be triggered on any particular page. It is possible this breaks some people's test pages, but it should not break anyone's real-world applications.

### DIFF
--- a/src/lib/doubleRAF.ts
+++ b/src/lib/doubleRAF.ts
@@ -1,0 +1,19 @@
+/*
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


tunetheweb (approved):
LGTM!

PR #284, opened 2022-11-15
Support 'restore' as a navigationType

### LINKED ISSUE #280: Consider adding a "restore" navigation type
Currently, for pages that are discarded and then restored, the only way to detect this is via `document.wasDiscarded`. This is easy enough, but it would be better if these were tracked via a distinct `navigationType` value on the [`Metric`](https://github.com/GoogleChrome/web-vitals#metric) object.

In Chrome today, these loads receive a navigation type of `back_forward`, and in Safari they receive a type of `navigate`, both of which seem not quite applicable, so adding a dedicate type to this library seems like it would be useful to (1) help understand how often this happens, and (2) help determine whether these types of loads have any characteristics that are uncorrelated with other types of navigations.

tunetheweb:
Fixes #280

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -58,9 +58,11 @@ interface PerformanceEntryMap {
 
 // Update built-in types to be more accurate.
 declare global {
-  // https://wicg.github.io/nav-speculation/prerendering.html#document-prerendering
   interface Document {
-    prerendering?: boolean
+    // https://wicg.github.io/nav-speculation/prerendering.html#document-prerendering
+    prerendering?: boolean;
+    // https://github.com/WICG/page-lifecycle/blob/main/README.md
+    wasDiscarded?: boolean;
   }
 
   interface Performance {
--- a/src/types.ts
+++ b/src/types.ts
@@ -61,7 +61,7 @@ declare global {
   interface Document {
     // https://wicg.github.io/nav-speculation/prerendering.html#document-prerendering
     prerendering?: boolean;
-    // https://github.com/WICG/page-lifecycle/blob/main/README.md
+    // https://wicg.github.io/page-lifecycle/#sec-api
     wasDiscarded?: boolean;
   }
 
--- a/src/types/base.ts
+++ b/src/types/base.ts
@@ -65,7 +65,7 @@ export interface Metric {
    * support that API). For pages that are restored from the bfcache, this
    * value will be 'back-forward-cache'.
    */
-  navigationType:  'navigate' | 'reload' | 'back-forward' | 'back-forward-cache' | 'prerender';
+  navigationType:  'navigate' | 'reload' | 'back-forward' | 'back-forward-cache' | 'prerender' | 'discarded';
 }
 
 /**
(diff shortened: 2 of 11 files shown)


src/types/base.ts
```diff
@@ -105,3 +105,12 @@ export interface ReportOpts {
  *   loading. This is equivalent to the corresponding `readyState` value.
  */
 export type LoadState = 'loading' | 'dom-interactive' | 'dom-content-loaded' | 'complete';
+
+/**
+ * Extend the document with the new wasDiscarded boolean which is not supported
+ * in typescript yet
+ * https://github.com/WICG/page-lifecycle/blob/main/README.md
+ */ 
+declare global {
```
philipwalton, line 114: This is fine, but can you move it here, similar to `document.prerendering`? https://github.com/GoogleChrome/web-vitals/blob/7f0ed0bfb03c356e348a558a3eda111b498a2a11/src/types.ts#L61-L64

test/e2e/onFCP-test.js
```diff
@@ -235,6 +235,23 @@ describe('onFCP()', async function() {
     assert.strictEqual(fcp2.navigationType, 'back-forward-cache');
   });
 
+  it('reports restore as nav type for wasDiscarded', async function() {
```
philipwalton, line 238: This looks good to me, but it also makes me realize that we don't have any tests for prerendering. We don't necessarily need to do it as part of this PR, but eventually I think it'd be good to update this test (in other test files as well) to just check all the different `navigationTypes` in a single test. WDYT?

test/views/layout.njk
```diff
@@ -100,6 +100,22 @@
         });
       }
 
+      /**
+       * @return {Promise<void>}
+       */
+      self.__stubWasDiscarded = () => {
+        return new Promise((resolve) => {
+          const navEntry = performance.getEntriesByType('navigation')[0];
```
philipwalton, line 108: This line isn't needed.

philipwalton (changes requested):

src/types/base.ts
tunetheweb: That took me ages to figure out! If only I'd looked harder and found this one!

test/e2e/onFCP-test.js
tunetheweb: We do for prerender:

https://github.com/GoogleChrome/web-vitals/blob/43a116b73f6bebcb409b336564eba93a9a4525a2/test/e2e/onFCP-test.js#L55-L76

src/types.ts
```diff
@@ -58,9 +58,11 @@ interface PerformanceEntryMap {
 
 // Update built-in types to be more accurate.
 declare global {
-  // https://wicg.github.io/nav-speculation/prerendering.html#document-prerendering
   interface Document {
-    prerendering?: boolean
+    // https://wicg.github.io/nav-speculation/prerendering.html#document-prerendering
+    prerendering?: boolean;
+    // https://github.com/WICG/page-lifecycle/blob/main/README.md
```
philipwalton, line 64:
```suggestion
    // https://wicg.github.io/page-lifecycle/#sec-api
```

test/e2e/onFCP-test.js
philipwalton: Yeah, though I guess only only some of the tests. I'd originally search `onCLS-test.js` and didn't see it.

I think it's fine to merge this PR as is. We can address this at some point in the future given that we do have some tests for it now.

philipwalton (approved):

test/e2e/onFCP-test.js
tunetheweb: Yeah looks like we're missing for CLS, INP, and FID. Should be easy enough to add so gimme 5 mins, and if trickier than that then will leave until later.

PR #285, opened 2022-11-15
Add missing prerender tests

tunetheweb:
CLS, FID, and INP didn't have prerender tests so this adds them.

### DIFF
--- a/test/e2e/onCLS-test.js
+++ b/test/e2e/onCLS-test.js
@@ -632,6 +632,27 @@ describe('onCLS()', async function() {
     assert.strictEqual(cls.navigationType, 'back-forward-cache');
   });
 
+  it('reports prerender as nav type for prerender', async function() {
+    if (!browserSupportsCLS) this.skip();
+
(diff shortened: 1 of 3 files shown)


test/e2e/onCLS-test.js
```diff
@@ -632,6 +632,27 @@ describe('onCLS()', async function() {
     assert.strictEqual(cls.navigationType, 'back-forward-cache');
   });
 
+  it('reports prerender as nav type for prerender', async function() {
+    if (!browserSupportsCLS) this.skip();
+
+    await browser.url('/test/cls?prerender=1');
+
+    const activationStart = await browser.execute(() => {
```
philipwalton, line 640: This isn't needed if you're just checking the nav type (here or in any of the other tests).

tunetheweb: Ah yeah. Copied from the FCP test but see it is actually using this value, whereas we aren't for CLS, FID, or INP.

It's gone!

philipwalton (approved):

PR #286, opened 2022-11-16
Add ability to run test suite for just one metric

tunetheweb:
Quite often I want to run just one set of tests (e.g. just the INP tests), rather than waiting for full suite to be run.

This PR allows this to be done without having to override the config in wdio.conf.js (and often forgetting to un-override it).

To run just INP tests, start the server and run:

```sh
npm run test:e2e -- --suite INP
```

To run just FID and INP tests, start the server and run:

```sh
npm run test:e2e -- --suite FID --suite INP
```

Note the `--` at the beginning needed to show the is an argument for the inner script, and not npm itself.


philipwalton:
I think this approach is fine, but as far as I can tell this doesn't allow you to specify just running the tests in a certain browser, so it may end up being nicer to just manually parse the command line flags to specify the behavior, rather than using webdriverio's `--suite` option.

For example, I think being able to do something like this would be nice:

```sh
npm test --methods=onINP --browsers=Chrome
```

(Or something like that)

I believe there are lots of npm modules that can help parse command line flag, but I'm not sure what's the latest and greatest there...

PR #288, opened 2022-11-24
Add prettier

tunetheweb:
Makes the following changes:
- Adds Prettier
- Adds some ESLint config to match prettier
- Adds Husky config to run it on commit
- Adds GitHub Action to auto run it in GitHub (in case commit it not made with huskey)
- Adds EditorConfig to do basic clean ups for files not covered by Prettier

Updates Selenium Standalone to allow tests to run on M1 Macs


README.md
```diff
@@ -174,18 +174,18 @@ _**Important!** The [unpkg.com](https://unpkg.com) CDN is shown here for example
...
+    script.src = 'https://unpkg.com/web-vitals@3/dist/web-vitals.iife.js';
+    script.onload = function () {
+      // When loading `web-vitals` using a classic script, all the public
+      // methods can be found on the `webVitals` global namespace.
+      webVitals.onCLS(console.log);
+      webVitals.onFID(console.log);
+      webVitals.onLCP(console.log);
+    };
+    document.head.appendChild(script);
+  })();
```
tunetheweb, line 188:
```suggestion
})();
```

README.md
```diff
@@ -131,7 +131,7 @@ Then, inline the code from `dist/polyfill.js` into the `<head>` of your pages. T
 </html>
 ```
 
-It's important that the code is inlined directly into the HTML. *Do not link to an external script file, as that will negatively affect performance:
+It's important that the code is inlined directly into the HTML. \*Do not link to an external script file, as that will negatively affect performance:
```
philipwalton, line 134:
```suggestion
It's important that the code is inlined directly into the HTML. _Do not link to an external script file, as that will negatively affect performance:_
```

docs/code-of-conduct.md
```diff
@@ -14,17 +14,17 @@ We value diverse opinions, but we value respectful behavior more.
...
-* Not engaging in unwanted physical contact.
+- Not engaging in unwanted physical contact.
 
 Some Google open source projects [may adopt][] an explicit project code of
 conduct, which may have additional detailed expectations for participants. Most
 of those projects will use our [modified Contributor Covenant][].
 
 [may adopt]: https://opensource.google/docs/releasing/preparing/#conduct
-[modified Contributor Covenant]: https://opensource.google/docs/releasing/template/CODE_OF_CONDUCT/
+[modified contributor covenant]: https://opensource.google/docs/releasing/template/CODE_OF_CONDUCT/
```
philipwalton, line 27: Do you know what rule caused the capitalization to be updated here?

These files are part of the standard Google OSS repo example, and I don't think we want to be making changes to them.

package.json
```diff
@@ -96,6 +96,8 @@
     "build:js": "rollup -c",
     "clean": "rm -rf dist tsconfig.tsbuildinfo",
     "dev": "run-p watch test:server",
+    "format": "prettier \"**/*.{mjs,js,json,ts,css,html,md,yml,yaml}\" --write --ignore-path .gitignore",
```
philipwalton, line 99:
```suggestion
    "format": "prettier \"**/*.{cjs,js,json,md,ts}\" --write --ignore-path .gitignore",
```

I'd rather stick to the file types that this repo is actually using rather than a generic list. E.g. this list didn't include `cjs`, which is in the repo, but it included a bunch of file types that are not.

package.json
```diff
@@ -96,6 +96,8 @@
...
+    "format:check": "prettier \"**/*.{mjs,js,json,ts,css,html,md,yml,yaml}\" --check --ignore-path .gitignore",
```
philipwalton, line 100:
```suggestion
    "format:check": "prettier \"**/*.{cjs,js,json,md,ts}\" --check --ignore-path .gitignore",
```

package.json
```diff
@@ -142,6 +144,10 @@
       "pre-commit": "npm run lint"
     }
   },
+  "prettier": {
```
philipwalton, line 147: Can you use these settings?

```json
{
  "arrowParens": "always",
  "bracketSpacing": false,
  "quoteProps": "preserve",
  "singleQuote": true
}
```

src/lib/observe.ts
```diff
@@ -14,17 +14,19 @@
...
-
+import {
+  FirstInputPolyfillEntry,
+  NavigationTimingPolyfillEntry,
+} from '../types.js';
 
 interface PerformanceEntryMap {
-  'event': PerformanceEventTiming[];
-  'paint': PerformancePaintTiming[];
+  event: PerformanceEventTiming[];
```
philipwalton, line 23: These should stay quoted. I also updated the prettier options to preserve object property quote usage.

wdio.conf.cjs
```diff
@@ -66,19 +64,19 @@ module.exports.config = {
   //
   capabilities: [
     {
-      'pageLoadStrategy': 'none',
+      pageLoadStrategy: 'none',
```
philipwalton, line 67: Keep the original quotes around these as well.

philipwalton (changes requested):

philipwalton:
Oh, could you also add `lint-staged`, similar to how it's implemented in the `pending-beacon-polyfill` repo?

See: https://github.com/GoogleChromeLabs/pending-beacon-polyfill/blob/e67b836324136b62a2c22bc0241cae21c7167a24/package.json#L99-L102

And: https://github.com/GoogleChromeLabs/pending-beacon-polyfill/blob/e67b836324136b62a2c22bc0241cae21c7167a24/.husky/pre-commit#L4

This will run prettier and automatically add any changes on commit, which is nice.

docs/code-of-conduct.md
tunetheweb: Hmmm... not sure why that happened and doesn't happen now. Will revert.

package.json
tunetheweb: Do you perhaps want `"quoteProps": "consistent",`? That seems to fix the two issues highlighted below.

tunetheweb:
Ah I guess the benefit of the `lint-staged` command is it only applies to the files being committed. Whereas my way runs it on the whole repo. I think for a small repo like this, there's not a huge difference, so I'd prefer to avoid the code duplication. Also has the added advantage of running all the files (e.g. if upgrading prettier, or config). WDYT?

package.json
philipwalton: No, looking at [an internal discussion](https://groups.google.com/a/google.com/g/javascript-style/c/Ob2Erbr29Uw/m/M2Unems8BgAJ) on prettier preferences, "preserve" is recommended because of how Closure Compiler does property minification. It's not safe to add or remove quote on object properties as that would potentially break the code after its compiled, and since this library is used in google3, I think it's better to not change this.

philipwalton:
Using `lint-staged` has the benefit that you don't have to think about it. If you don't use `lint-staged` and then you have an error, the commit command fails, then you have to run the linter/formatter, and then you have to run commit again. So I see `lint-staged` more as a convenience tool rather than a code quality or enforcement tool (similar to format on save in your editor).

Having the linter/formatter also run as a Github seems nice as a way to ensure all contributions also adhere to the lint/style rules, so I see that as valuable in addition to the convenience of `lint-staged`.

docs/code-of-conduct.md
tunetheweb: So it did change this, and it can't seem to be switched off. However this is just the link reference and [it is case insensitive](https://spec.commonmark.org/0.25/#example-505). So the actual link is rendered correctly with upper case [as shown here](https://github.com/GoogleChrome/web-vitals/blob/4b84d071710797997a14d1698a8c5317b9329912/docs/code-of-conduct.md).

Good to keep this change?

Or, alternatively, moving these to inline links rather than reference links is probably more easily understood by most people used to GitHub markdown syntax, and also passes prettier, but is another change to the file.

Or ignore this file, but that means adding a `.prettierignore` file.

WDYT?

tunetheweb:
OK it looks like `lint-staged` automatically adds the files to git as husky doesn't do that automatically.

I've now changed the pre-commit file to the following:

```
#!/usr/bin/env sh
. "$(dirname -- "$0")/_/husky.sh"

npm run format
npm run lint:fix
git add .
```

Which has the same effect, though perhaps it's easier to add temporary files with this so maybe `lint-staged` is better.

tunetheweb:
OK switched to `lint-staged` as confirmed it doesn't add new files.

Could also change my:
```git add .```
to
```git add -u .```

But figured might as well use `lint-staged` since it seems cleverer than I am.

docs/code-of-conduct.md
philipwalton: > However this is just the link reference and [it is case insensitive](https://spec.commonmark.org/0.25/#example-505). So the actual link is rendered correctly with upper case [as shown here](https://github.com/GoogleChrome/web-vitals/blob/4b84d071710797997a14d1698a8c5317b9329912/docs/code-of-conduct.md).

Ah, I see. I missed that it was just the reference. Yeah, this seems fine.

philipwalton (approved):

PR #290, opened 2022-12-08
fix: safe access navigation entry type

AndersDJohnson:
Closes https://github.com/GoogleChrome/web-vitals/issues/291.

Work in progress...

We could also consider sending `undefined` instead of `'unknown'`.

Not sure how this new value affects the rest of the system, especially the types.


src/lib/initMetric.ts
```diff
@@ -32,9 +32,11 @@ export const initMetric = (name: Metric['name'], value?: number): Metric => {
       navigationType = 'prerender';
     } else if (document.wasDiscarded) {
       navigationType = 'restore';
-    } else {
+    } else if (navEntry.type) {
       navigationType =
           navEntry.type.replace(/_/g, '-') as Metric['navigationType'];
+    } else {
+       navigationType = 'unknown';
```
AndersDJohnson, line 39: TODO: Type here?

AndersDJohnson:
@tunetheweb Done, thanks!

tunetheweb (approved):
LGTM!

cklim4405:
> Closes https://github.com/GoogleChrome/web-vitals/issues/291.
> 
> Work in progress...
> 
> We could also consider sending `undefined` instead of `'unknown'`.
> 
> Not sure how this new value affects the rest of the system, especially the types.
>

PR #293, opened 2022-12-12
Added possibility of multiple independent INP measurements during use…

proofyman:
We want to measure INP in SPA applications, so i made minor changes to add this functionality.

They make possible to send data whenever we want and reset counters in custom builds.


tunetheweb:
I'm planning on having a look at adding SPA support, through the new ["soft navigations" work](https://github.com/WICG/soft-navigations).

I was planning is to allow both the traditional full lifecycle metrics to be measured, as well as the shorter "soft navigation" metrics, to allow the two of them to be measured (and compared). Particularly while CrUX only measures the former.

proofyman:
https://github.com/GoogleChrome/web-vitals/pull/294 - here is correct PR, sorry for double

We've used our own counters(ya.ru), but willing to reuse web-vitals counters code. I shorten description just to remove details. Soft navigation does not solve our problems, because we need to take full control over INP measurement lifecycle.

In this PR I just provide ability to force report(this way we can report INP in specific moments related to our app, they more meaningfull than soft navigations for us) and reset state to new measurements.

So, please accept my changes  =)

PR #294, opened 2022-12-12
Supported multiple INP measurements during single user session in custom builds.

proofyman:
We are using web-vitals counter in production and need fine-grained control over stats for our SPA.

These changes allow us to send INP for every page of out SPA.

### DIFF
--- a/src/onINP.ts
+++ b/src/onINP.ts
@@ -51,7 +51,7 @@ let longestInteractionList: Interaction[] = [];
 
 // A mapping of longest interactions by their interaction ID.
 // This is used for faster lookup.
-const longestInteractionMap: {[interactionId: string]: Interaction} = {};
+let longestInteractionMap: {[interactionId: string]: Interaction} = {};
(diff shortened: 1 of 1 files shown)


proofyman:
https://github.com/GoogleChrome/web-vitals/pull/293#issuecomment-1346843896 - comment about double-PR

proofyman:
@tunetheweb Hi, could you give me any feedback about this PR? Is it possible to merge such changes, or should I clone this repo to implement them.

philipwalton:
Hi @proofyman, thanks for your contribution. We do plan to support SPA navigations in the future, but we want to do that in a standard way using the proposed [SoftNavigation performance entry](https://chromestatus.com/feature/5144837209194496) rather than a custom reset that developers have to manually call.

If your solution is working well for you, feel free to keep using it, just be aware that the INP value you measure will not be comparable to the value you see in the Chrome User Experience Report.

PR #297, opened 2022-12-21
Defer CLS logic until after onFCP callback

### LINKED ISSUE #180: Reporting CLS and LCP in a lab environment using Playwright is flakey / unreliable
Hi, 

we are bumping into an issue where we want to gather some web vitals using this library in a lab environment. We are using Playwright for this and injecting the `web-vitals` code into the browser context. Injecting and instrumenting works fine, but we are getting intermittent issues with some metrics not reporting. 

Specifically:
- **LCP** seems to not be reported unless we force some interaction. A simple click on the body normally does the trick.
- **CLS** seems to come and go. 

Conversely, **FCP** and **TTFB** are pretty reliable.

My question is: is this expected, or are we missing something? Could this be some race condition we can avoid?

Here is an reproducable example. The `web-vitals` UMD module is inlined. The only external dependency is [Playwright](https://playwright.dev/)

```js
const { chromium } = require("playwright")

// UMD module copy-pasted from the NPM version of https://github.com/GoogleChrome/web-vitals
const WEB_VITALS_SCRIPT = `!function(e,t){"object"==typeof exports&&"undefined"!=typeof module?t(exports):"function"==typeof define&&define.amd?define(["exports"],t):t((e="undefined"!=typeof globalThis?globalThis:e||self).webVitals={})}(this,(function(e){"use strict";var t,n,i,a,r=function(e,t){return{name:e,value:void 0===t?-1:t,delta:0,entries:[],id:"v1-".concat(Date.now(),"-").concat(Math.floor(8999999999999*Math.random())+1e12)}},o=function(e,t){try{if(PerformanceObserver.supportedEntryTypes.includes(e)){var n=new PerformanceObserver((function(e){return e.getEntries().map(t)}));return n.observe({type:e,buffered:!0}),n}}catch(e){}},c=function(e,t){var n=function n(i){"pagehide"!==i.type&&"hidden"!==document.visibilityState||(e(i),t&&(removeEventListener("visibilitychange",n,!0),removeEventListener("pagehide",n,!0)))};addEventListener("visibilitychange",n,!0),addEventListener("pagehide",n,!0)},u=function(e){addEventListener("pageshow",(function(t){t.persisted&&e(t)}),!0)},f="function"==typeof WeakSet?new WeakSet:new Set,s=function(e,t,n){var i;return function(){t.value>=0&&(n||f.has(t)||"hidden"===document.visibilityState)&&(t.delta=t.value-(i||0),(t.delta||void 0===i)&&(i=t.value,e(t)))}},d=-1,p=function(){return"hidden"===document.visibilityState?0:1/0},m=function(){c((function(e){var t=e.timeStamp;d=t}),!0)},v=function(){return d<0&&(d=p(),m(),u((function(){setTimeout((function(){d=p(),m()}),0)}))),{get timeStamp(){return d}}},l={passive:!0,capture:!0},h=new Date,y=function(e,a){t||(t=a,n=e,i=new Date,E(removeEventListener),g())},g=function(){if(n>=0&&n<i-h){var e={entryType:"first-input",name:t.type,target:t.target,cancelable:t.cancelable,startTime:t.timeStamp,processingStart:t.timeStamp+n};a.forEach((function(t){t(e)})),a=[]}},S=function(e){if(e.cancelable){var t=(e.timeStamp>1e12?new Date:performance.now())-e.timeStamp;"pointerdown"==e.type?function(e,t){var n=function(){y(e,t),a()},i=function(){a()},a=function(){removeEventListener("pointerup",n,l),removeEventListener("pointercancel",i,l)};addEventListener("pointerup",n,l),addEventListener("pointercancel",i,l)}(t,e):y(t,e)}},E=function(e){["mousedown","keydown","touchstart","pointerdown"].forEach((function(t){return e(t,S,l)}))};e.getCLS=function(e,t){var n,i=r("CLS",0),a=function(e){e.hadRecentInput||(i.value+=e.value,i.entries.push(e),n())},f=o("layout-shift",a);f&&(n=s(e,i,t),c((function(){f.takeRecords().map(a),n()})),u((function(){i=r("CLS",0),n=s(e,i,t)})))},e.getFCP=function(e,t){var n,i=v(),a=r("FCP"),c=o("paint",(function(e){"first-contentful-paint"===e.name&&(c&&c.disconnect(),e.startTime<i.timeStamp&&(a.value=e.startTime,a.entries.push(e),f.add(a),n()))}));c&&(n=s(e,a,t),u((function(i){a=r("FCP"),n=s(e,a,t),requestAnimationFrame((function(){requestAnimationFrame((function(){a.value=performance.now()-i.timeStamp,f.add(a),n()}))}))})))},e.getFID=function(e,i){var d,p=v(),m=r("FID"),l=function(e){e.startTime<p.timeStamp&&(m.value=e.processingStart-e.startTime,m.entries.push(e),f.add(m),d())},h=o("first-input",l);d=s(e,m,i),h&&c((function(){h.takeRecords().map(l),h.disconnect()}),!0),h&&u((function(){var o;m=r("FID"),d=s(e,m,i),a=[],n=-1,t=null,E(addEventListener),o=l,a.push(o),g()}))},e.getLCP=function(e,t){var n,i=v(),a=r("LCP"),d=function(e){var t=e.startTime;t<i.timeStamp&&(a.value=t,a.entries.push(e)),n()},p=o("largest-contentful-paint",d);if(p){n=s(e,a,t);var m=function(){f.has(a)||(p.takeRecords().map(d),p.disconnect(),f.add(a),n())};["keydown","click"].forEach((function(e){addEventListener(e,m,{once:!0,capture:!0})})),c(m,!0),u((function(i){a=r("LCP"),n=s(e,a,t),requestAnimationFrame((function(){requestAnimationFrame((function(){a.value=performance.now()-i.timeStamp,f.add(a),n()}))}))}))}},e.getTTFB=function(e){var t,n=r("TTFB");t=function(){try{var t=performance.getEntriesByType("navigation")[0]||function(){var e=performance.timing,t={entryType:"navigation",startTime:0};for(var n in e)"navigationStart"!==n&&"toJSON"!==n&&(t[n]=Math.max(e[n]-e.navigationStart,0));return t}();n.value=n.delta=t.responseStart,n.entries=[t],e(n)}catch(e){}},"complete"===document.readyState?setTimeout(t,0):addEventListener("pageshow",t)},Object.defineProperty(e,"__esModule",{value:!0})}));`

async function run () {

  const browser = await chromium.launch()
  context = await browser.newContext()

  // START specific instrumentation code

  await context.addInitScript(WEB_VITALS_SCRIPT)

  await context.addInitScript(() => {
    // Wait for the init script to be fully loaded
    // https://github.com/GoogleChrome/web-vitals/issues/38
    window.addEventListener('DOMContentLoaded', () => {
      webVitals.getTTFB(metric => { console.trace(JSON.stringify({ name: metric.name, value: metric.value, blob: metric })) })
      webVitals.getFCP(metric => { console.trace(JSON.stringify({ name: metric.name, value: metric.value, blob: metric })) })
      webVitals.getLCP(metric => { console.trace(JSON.stringify({ name: metric.name, value: metric.value, blob: metric })) })
      webVitals.getCLS(metric => { console.trace(JSON.stringify({ name: metric.name, value: metric.value, blob: metric })) })
    })
  })

  const page = await context.newPage()
  page.on('console', async msg => {
    if (msg.type() === 'trace') {
      console.log(msg.text())
    }
  });

  // STOP specific instrumentation code

  await page.goto("https://checklyhq.com")
  // Doing a 'page.click()' on the body triggers the LCP metric to be reported much more reliably
  // await page.click('body')

  await browser.close()
}

run()
```
The typical output from this code looks like:

```shell
/usr/local/bin/node /Users/timnolet/WebstormProjects/playwright-scratch/web-vitals/web-vitals-inline.js

{"name":"FCP","value":846.6000000238419}
{"name":"TTFB","value":396}

Process finished with exit code 0
```

philipwalton:
This PR fixes #180 by removing some of the logic that would listen and keep track of whether or not FCP had fired, and instead just wraps all of the `onCLS()` logic inside of an `onFCP()` call (and ensuring the logic only runs the first time `onFCP()` is invoked).

The "ensuring the logic only runs the first time `onFCP()` is invoked" bit is accomplished by introducing a new `runOnce()` utility function, and I took this opportunity to refactor some of the other code to make use of that. The end result was a little bit of byte savings, so it was worth it.

Also, #180 caught a bug in the library where, if the library was lazy-loaded, CLS wouldn't properly be reported when `reportAllChanges` was true. I realized there were no tests to ensure things work when the library is lazy-loaded, so I added some, which also required refactoring some of the testing logic.

Lastly, this PR also updates a bit of the formatting style (continuing what was done in #288) since some of that was auto-formatting unrelated files anyway (not sure why, but seems to be fine now).


test/views/layout.njk
```diff
@@ -115,6 +115,47 @@
...
+              }
+            }
+            if (nodes.size === 0) {
+              resolve();
+            }
+          }).observe({type: 'element', buffered: true});
+        }, true);
+      });
+
+            /**
```
tunetheweb, line 152: Nit, formatting:

```suggestion
      /**
```

test/views/layout.njk
```diff
@@ -150,6 +199,30 @@
       if (params.has('wasDiscarded')) {
         self.__stubWasDiscarded();
       }
+
+      // Push to this promise list if you need to wait for some condition in a test.
+      self.__readyPromises = [
+        // new Promise(r => setTimeout(r, 1000))
+      ];
```
tunetheweb, line 206: Should this comment be removed?

```suggestion
      self.__readyPromises = [];
```

tunetheweb (approved):
LGTM but have some comments (mostly for my own curiosity).

src/onCLS.ts
```diff
@@ -49,99 +49,88 @@ export const onCLS = (onReport: CLSReportCallback, opts?: ReportOpts) => {
...
+        onHidden(() => {
+          handleEntries(po.takeRecords() as CLSMetric['entries']);
+          report(true);
+        });
+
+        // Only report after a bfcache restore if the `PerformanceObserver`
+        // successfully registered.
+        onBFCacheRestore(() => {
+          sessionValue = 0;
+          metric = initMetric('CLS', 0);
```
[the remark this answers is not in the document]
philipwalton: I think it could, but I don't think there's any good way to prevent it. We have another similar problem today where someone could load a page and then switch tabs away and then back again before the web-vitals library loads, and we can't detect those cases properly right now in JS (though they are tracked properly in UKM).

The approach I've been taking is to be as accurate as possible for now, without going overboard  with extra complexity, and wait until we eventually have native APIs for all of these edge cases (e.g. [`visibility-state` performance entries](https://chromestatus.com/feature/5683502144028672)).

test/views/layout.njk
philipwalton: 🤦

Kobe:
@philipwalton @tunetheweb does this adjustment have an impact on concrete CLS numbers?

philipwalton:
It does not, other than cases affected by the bug outlined in this comment: https://github.com/GoogleChrome/web-vitals/issues/180#issuecomment-1342366169

PR #305, opened 2023-01-11
Add a note about when to load the library

### LINKED ISSUE #304: Docs should mention "when to load"
To some it might be obvious, but the readme does not mention when to load the library. It mentions install, it mentions how to use (the methods of the lib), integrations in other stuff, but not "when to load".

Only in this [Codelab](https://codelabs.developers.google.com/codelabs/web-vitals-google-analytics-bigquery#1) it is mentioned
> Copy this module script, and paste it **just before the closing `</body>` tag**.

Not everyone reads this Codelab, and I also don't think everyone using the `web-vitals` knows what to do best. 
Questions like:
- "Can/Should I lazy load the lib?" 
- "Do I need to load it as early as possible to not miss any values?" 
- "Do I register the callbacks on DOMContentLoaded/onLoad/using requestIdleCallback?" 

are open questions I can think of in the environment of performance, because there is no guideline.

tunetheweb:
Fixes #304

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,9 @@ The library supports all of the [Core Web Vitals](https://web.dev/vitals/#core-w
 
 <a name="import-web-vitals-from-npm"><a>
 
-The `web-vitals` library uses the `buffered` flag for [PerformanceObserver](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver/observe) meaning it receives all performance entries that happened before the library was loaded. This means you do not need to prioritse the loading of the `web-vitals` library. Given that, it is recommended to defer the library load until after the main content of the page is loaded.
+The `web-vitals` library uses the `buffered` flag for [PerformanceObserver](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver/observe), allowing it to access performance entries that occurred before the library was loaded.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -51,6 +51,8 @@ The library supports all of the [Core Web Vitals](https://web.dev/vitals/#core-w
 
 <a name="import-web-vitals-from-npm"><a>
 
+The `web-vitals` library uses the `buffered` flag for [PerformanceObserver](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver/observe) meaning it receives all performance entries that happened before the library was loaded. This means you do not need to prioritse the loading of the `web-vitals` library. Given that, it is recommended to defer the library load until after the main content of the page is loaded.
```
philipwalton, line 54:
```suggestion
The `web-vitals` library uses the `buffered` flag for [PerformanceObserver](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver/observe), allowing it to access performance entries that occurred before the library was loaded.

This means you do not need to load this library early in order to get accurate performance data. In general, this library should be deferred until after other user-impacting code has loaded.
```

philipwalton (approved):
LGTM (suggested a slight reword)

PR #308, opened 2023-01-19
Add Soft Navigation support

### LINKED ISSUE #119: Web Vital Metrics for Single Page Applications
In React Single page Application(SPA), First Input Delay and Largest Contentful Paint are only measured once on Initial Load. Additionally, Cumulative Layout Shift does not reset to 0 throughout the session. This means CLS could reach to very high value , if user continues to navigate though logical pages like homepage -> search -> Product Listing -> Product Page -> Basket -> Checkout and so on.

Based on current implementation, pageshow event resets CLS to be 0 and it also captures FID and LCP for subsequent page loads.

In React Single Page Application, Could we consider onRouteChanged event for the new **logical page load** and reset CLS to 0. This would also enable us to capture FID and LCP on every logical page of the application.

If Web Vital Metrics would be considered for page ranking, the proposed implementation would give React SPA fair comparison.

### LINKED ISSUE #724: [soft-navs] INP is not reporting when all interactions are too fast (and we don't have first-input fallback)
Event Timing has a durationThreshold and the default value for web-vitals.js is 40ms.

In my testing on desktop, it is not uncommon for sites to have interactions under that threshold (some sites even have interactions come in under 16ms).

Because we "reset" the INP metric with each new soft nav, its also relatively common to have 1 interaction per soft-nav, now.  So basically any page with a fast interaction ends up without an INP value, and only pages with large INP values end up reporting...

---

CLS has a similar problem, but the web-vitals.js library tries to 'flush' a final CLS value of 0 to increase reporting rates (ideally, every page with FCP also gets a CLS, even if its 0).

`onCLS()` will report without any `entries: []` in that case.

So, I think we could do the same for INP.  Whenever visibility changes, or the page unloads, or a new soft-nav is detected, and we are finalizing/flushing the current scores-- we should:

- Always report CLS of 0 (if it hasn't been reported yet)
- Report INP of 0, if `interactionCount` has increased since the start of this navigation (and INP wasn't reported yet)

---

Note: I suspect we may have a similar issue with bfcache restoration and this fix might help there too?

tunetheweb:
**Warning: This branch is likey to stay in draft for some time, and not to be merged into `main` until the Soft Navs experiment completes**

Fixes #119
Fixes #724

Implements experimental support of [Soft Navigations](https://github.com/WICG/soft-navigations) to the library:
- [x] - Updated README
- [x] - TTFB
- [x] - FCP
- [x] - LCP
- [x] - CLS
- [x] - INP
- [x] - Attribution (particularly navigationTiming)
- [x] - Test cases
- [x] - Agree final WDIO setup (needs to wait for a supporting version of Chrome to be available in WDIO)
- [x] - Remove version-specifc info in `package.json` and `package-lock.json`


src/onCLS.ts
```diff
@@ -62,9 +65,52 @@ export const onCLS = (onReport: CLSReportCallback, opts?: ReportOpts) => {
...
+          thresholds,
+          opts!.reportAllChanges
+        );
+        sessionValue = 0;
+        sessionEntries = [];
+      };
+
       // const handleEntries = (entries: Metric['entries']) => {
       const handleEntries = (entries: LayoutShift[]) => {
+        let pageUrl = '';
```
mmocny, line 83: This would compute url every time new shifts are reported (~once per animation frame).

I think you can cache it until a navigation happens.

src/onCLS.ts
```diff
@@ -62,9 +65,52 @@ export const onCLS = (onReport: CLSReportCallback, opts?: ReportOpts) => {
...
         entries.forEach((entry) => {
+          if (
```
mmocny, line 85: Perhaps a cleaner way to do it is to have the helper class that calls handleEntries automatically split entries based on navigationID?

Then, you dont' need to do this once per entry.

Also, that helper that does the split could call something like "finalizeAndReset", so we wouldn't need to do these tests here at all.  The metric specific handlers would just get a URL and maintain a score...

WDYT?

src/onCLS.ts
```diff
@@ -126,6 +173,30 @@ export const onCLS = (onReport: CLSReportCallback, opts?: ReportOpts) => {
...
+                thresholds,
+                opts!.reportAllChanges
+              );
+              currentNav = entry.navigationId;
+            }
+          });
+        };
+
+        if (softNavs(opts)) {
+          observe('soft-navigation', reportSoftNavCLS);
```
mmocny, line 192: Followup from comment above... I think we should just automatically do this from inside the observe() helper, shared by all metrics

mmocny (commented):
Hey, this is excellent work.  Quite a bit to get right, thanks for the effort!

I only review the CLS changes, and I think there's a potentially cleaner way to do it.  I didn't review the specifics of the other metrics since I the change would alter a lot there too, so wanted to discuss that first.

Basically:
- I think the observe() helper, which all metrics use, should be the only place where we look at navigationID and split entries whenever it changes.  See: https://github.com/GoogleChrome/web-vitals/blob/main/src/lib/observe.ts#L52
- Whenever navigationID changes (or, the first time for initial nav), that helper can look up pageURL, etc, that way we have shared cache for all metrics
- Each metric implementation right now already uses helpers like onFCP, onHidden, onBFCacheRestore... in order to "finalize" and/or "reset" its metrics.  I think we could have another helper like onNavigate(), which can be used to flush/reset metrics.

One tricky bit may be that the onNavigate() helper needs to be called before any new Entry that uses that navigationID, and you could have entries with different navigationID in a single PerformanceObserver callbacks, especially when using buffered entries.

That said, I think all this code made be fragile to such cases and needs careful testing.

Keep at it!

src/onCLS.ts
tunetheweb: Given how easy it is to look up the pageUrl, I've decoded to simplify this and go back to reporting only he navId. So this is now gone.

src/onCLS.ts
tunetheweb: While that may look like cleaner code, I wonder about the performance implications of splitting up the array like versus doing this check everytime? In effect you’d need to do that check anyway for each entry AND copy the entries to a new array. Which sounds like a net negative to me.

WDYT?

src/onCLS.ts
tunetheweb: Not all metrics need the soft-navigation observer (FCP and FID don’t for example as they can finalise immediately).

For those that do need this to finalise metrics that may not be finalised before then, there are different steps to make so it’s not generic across all the metrics. We could pass in a callback function to handle that, but that’s exactly what this currently does.

Also currently the observe function is for a single observer with a call back on what to do with that. I’m not sure it’s right to change that to also call another observer for some metrics, with a different callback function.

But maybe I’m just not seeing the way to do this? If you have a look at some of the other metrics and still think there’s an easier/cleaner way to do this, then definitely open to it, if you could give me a little nudge as to how you think it could be structure.

tunetheweb:
That’s for the review @mmocny ! I’ve cleaned it up quite a bit further so it’s a bit smaller now. Have another look if you get a chance and see what you think now.

alekseykulikov:
Hey! I'm super excited about this work. Could you republish the branch?
The current [3.1.1-soft-navs](https://www.npmjs.com/package/web-vitals/v/3.1.1-soft-navs) doesn't report LCP  and instead prints `BARRY`.
Screen Shot 2023-02-08 at 22 30 06

tunetheweb:
Well that's embarrassing! Republished to the `soft-navs` tag (which is an alias to `3.1.1-soft-navs-1` now).

alekseykulikov:
Thanks, that was fast.
Maybe, I'm doing something wrong, but there's still no LCP report, despite actual values are available (Chrome 110):
Screen Shot 2023-02-08 at 23 00 17

tunetheweb:
Can you host the site somewhere so I can have a look?

Here's a demo site where it does work: https://www.tunetheweb.com/experiments/softnavsdemo/

alekseykulikov:
Thanks for the example. 
Found the issue: when `onLCP` is called twice (like in [the example](https://developer.chrome.com/blog/soft-navigations-experiment/)), it doesn't report soft-navs. Other metrics work.

```js
// works (your example)
onLCP(doSoftNavProcessing, { reportSoftNavs: true })

// !! doesn't report soft navs
onLCP(doTraditionalProcessing)
onLCP(doSoftNavProcessing, { reportSoftNavs: true })

// works as expected
onCLS(doTraditionalProcessing)
onCLS(doSoftNavProcessing, { reportSoftNavs: true })
```

tunetheweb:
Oh that's interesting. Will dig into it...

tunetheweb:
That's fixed now if you want to reinstall the `soft-navs` tag (or `3.1.1-soft-navs-2` if you prefer to be explicit).

alekseykulikov:
Thanks, Barry. That was faster than I finished my coffee. 
Another issue:  the `navigationType` for FCP & FID is always `soft-navigation`. So the way to distinguish the hard nav and softs is to use `navigationId !== 1`.
You can replicate the issue in your demo example.

tunetheweb:
Fixed!

Hoping to be able to add automated tests to this branch soon now this has landed in stable and so will be available in our test automation suite. But in the meantime greatly appreciate you finding them!

karreiro:
Hi @tunetheweb,

Thanks a lot for introducing support for Soft Navigations in the library!

While exploring the new features, I noticed something interesting. It seems that not all changes are being reported when `reportAllChanges` is activated, as the `onINP` module is only tracking the longest interactions.

To handle that scenario, I've slightly updated the `onINP` module (https://github.com/GoogleChrome/web-vitals/pull/411) to ensure that all changes are reported when `reportAllChanges` is turned on ([demo](https://private-user-images.githubusercontent.com/1079279/294832168-e72df86a-e373-46b2-930e-aca95576ba9c.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDQ2OTkwOTgsIm5iZiI6MTcwNDY5ODc5OCwicGF0aCI6Ii8xMDc5Mjc5LzI5NDgzMjE2OC1lNzJkZjg2YS1lMzczLTQ2YjItOTMwZS1hY2E5NTU3NmJhOWMuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDEwOCUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDAxMDhUMDcyNjM4WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9Zjg1Y2I5Y2E4OWE0YWUwYTFkODlhNTQzMGIxZjVmNWZmMGQyM2ZmY2FjYTlkZGJmOWQ4ZWE4OTYzN2E0YjZmMCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.gFdLD-nThJP5n6_Y3Bn6H-gKjSqtHLOZPQGmWW7QkMs)). I've directed my PR to your branch, as my changes are somewhat related to soft navs. I'd be glad to hear your thoughts on this.

tunetheweb:
As discussed in #411 that is working as intended. But I raised #413 to make it clearer what exactly that intention is.

tunetheweb:
Updated this branch to v4 code.

wolt-kgarg:
Hi @tunetheweb has there been any updates on the soft nav experiments, the last update was in August 2024 in this [blog](https://developer.chrome.com/docs/web-platform/soft-navigations-experiment). I also couldn't find any way to enable the soft nav in chrome, is this enabled by default? Would appreciate any pointers here. TIA.

joseph-embrace:
Speaking as an SDK author, I see a huge spectrum of "soft navigation" definitions. Some site owners want to track every interaction as a soft nav, and some have a very narrow & specific definition. What is the likelihood that web-vitals would support caller-defined soft navs to accommodate this spectrum?

To take this question even further: do you see the Soft Navigation API as becoming integral to the very definition of the web vitals metrics?

tunetheweb:
That's a very interesting question @joseph-embrace , and previously I've taken a very hard stance on this, as stated previously here: https://github.com/GoogleChrome/web-vitals/issues/448#issuecomment-2026242050

This was primarily due to four reasons:

1. LCP could not be sliced separately, only CLS and INP. And I was not happy to support 2/3 metrics this way and think the expectation (and maybe even assumption!) is that all three will be supported.
2. The requirement to attribute the metrics to an `navigationId` to accurately get the URL to report this under.
3. A desire to avoid confusing support issues when people have different definitions of what metrics the library produces, depending how they use it. While it's true there are some options for this already (e.g. `reportAllChanges`, or `durationThreshold`, or even `generateTarget`), there is still only one definition of what the metric is at the end of page lifetime, even if users choose to beacon early.
4. The aim of the library, which as stated in the README is to:

> The web-vitals library is a tiny (~2K, brotli'd), modular library for measuring all the [Web Vitals](https://web.dev/articles/vitals) metrics on real users, in a way that accurately matches how they're measured by Chrome and reported to other Google tools (e.g. [Chrome User Experience Report](https://developers.google.com/web/tools/chrome-user-experience-report), [Page Speed Insights](https://developers.google.com/speed/pagespeed/insights/), [Search Console's Speed Report](https://webmasters.googleblog.com/2019/11/search-console-speed-report.html)).

However, with the new API shape, and the introduction of `interaction-contentful-paint` over resetting LCP, 1 is not as much of an issue as it was before.

And similarly we depend less on `navigationId` now, so it could be argued that 2 is also less of an issue. Though instead we now use `interacitionId` more instead, which is basically the same, so that's not completely solved.

3 and 4 still stand though. And a similar issue is to measure all experiences equally (and not just to avoid intentionally gaming).

So my hard line stance is not quite as hard as it once was, but I think there are still problems to solve. So never say never, but for now, I'm concentrating on just supporting this with the browser defined soft navs.

> To take this question even further: do you see the Soft Navigation API as becoming integral to the very definition of the web vitals metrics?

That is another interesting question. Especially the cross-browser story since this will be Chrome-only (at least initially, and I suspect for quite some time). We finally get cross-browser LCP and INP support and then we in Chrome go ahead and break it 😭

We've also had a few discussions in the Web Perf Working Group as to whether metrics are standards, or only the underlying APIs. At present the feeling is (and I don't necessarily agree with this btw!) that only APIs are standardised and any algorithms they use to derive metrics is not something that should be standardised.

However, the other argument is that no performance metrics are "right" and everything is just an approximation of what we're trying to do here — which is to measure user experience. So if Chrome/CrUX/PSI/GSC can measure closer to the experience with these APIs, even if it's not cross-browser available, then we should, even if this does cause an interop issue in measuring the metrics. And if it's that much better it should create the demand for more support of these APIs in other browser engines. We know there is a gap in Core Web Vitals for SPA pages and we know we're not measuring it correctly and, now we nearly have the solution for that, we should use it.

Anyway, those are random thoughts from myself (and not necessarily shared by all of the Chrome team btw!). Definitely too much for a PR comment, so if you want to discuss further would you mind opening an issue for this? Or use #448 or #119 for that discussion?

Mimori256:
just want to make sure but will Chrome actually remove the `largestInteractionContentfulPaint` property when the OT ends? The current code keeps it as a non-optional property in the type definition, suggesting it still exists during the OT?

I'm currently using the OT to measure soft-navigation metrics in production, and I'm wondering whether updating to the latest version of web-vitals will be required to continue measuring correctly after the OT ends.

tunetheweb:
Yes it’s already gone from Chrome Canary. At the moment this library supports both so you can upgrade now in preparation and both Chrome 149 and 150+ will be supported.

Ideally we’d remove the legacy name from this library and I have a TODO in the code to remove it. But I’ll probably keep it for the initial v6 launch to ensure a seamless transition for OT participants.

src/attribution/onINP.ts
```diff
@@ -383,6 +383,43 @@ export const onINP = (
...
+        metric.navigationType === 'back-forward-cache')
+    ) {
+      const navStartTime = metric.navigationStartTime;
+      if (navStartTime) {
+        // For simplicity make some assumptions for values we can't get
+        // to avoid undefined/null values which would be unexpected.
+        // - Assume interactionType as pointer as the most common
+        // - Assume interactionTime of nav start time
+        // - Assume nextPaintTime as interactionTime + length
+        const attribution: INPAttribution = {
```
mmocny: Given that this is just attribution data, I would think that skipping attribution entirely is better than faking it.  I don't understand the use cases well enough to have a hard oppinion.

(My understanding is this attribution data is attached after `INPMetric` is already "faked" when `interactionCount` change is observed without entries?  It makes sense to me to fake *that*, for sure.)

src/lib/initMetric.ts
```diff
@@ -23,19 +23,28 @@ import {MetricType} from '../types.js';
 export const initMetric = <MetricName extends MetricType['name']>(
   name: MetricName,
   value: number = -1,
+  interactionId?: number,
+  navigation?: MetricType['navigationType'],
```
mmocny: nit: perhaps call this `navigationType` in the args, and then instead of a local `navigationType`, and then just default it to `"navigate"` if it is nullish?

src/lib/InteractionManager.ts
```diff
@@ -70,12 +71,31 @@ export class InteractionManager {
...
+    // If we have a non-zero interactionCountForNavigation but no
+    // candidateInteractionIndex, then it's below the 16ms limit
+    // so report a dummy 8ms interaction. This is only needed for
+    // soft-navs and bfcache restores as `first-input` handles the
+    // rest.
+    if (
+      interactionCountForNavigation &&
+      candidateInteractionIndex === -1 &&
+      (navigationType === 'soft-navigation' ||
+        navigationType === 'back-forward-cache')
```
mmocny, line 90: QQ: was this default never supported for bf-cache previously?  Nice improvement.

src/lib/observe.ts
```diff
@@ -38,17 +42,63 @@ export const observe = <K extends keyof PerformanceEntryMap>(
   callback: (entries: PerformanceEntryMap[K]) => void,
   opts: PerformanceObserverInit = {},
 ): PerformanceObserver | undefined => {
+  const typesToMonitor: (keyof PerformanceEntryMap)[] = [type];
```
mmocny: I like that these are "observed together" in the same observer because of the guarentees that gives you-- but should the library just use a singleton observer for *all* types? Or is that not feasible given the current api/design?

Not a blocking comment, more of a long term design question.

---

Perhaps more directly related to this patch: I don't know that I love that you call `observe(type)` but implicitly observe a different set of types.

What do you think about having something like `function getEntryTypesToObserveFor(category)` and category maps to a list of entryTypes, which you can observe.

It serves the same purpose, but now the caller of `observe()` knows what it is observing.  The observable effects likely end up the same....

src/lib/observe.ts
```diff
@@ -38,17 +42,63 @@ export const observe = <K extends keyof PerformanceEntryMap>(
...
       const po = new PerformanceObserver((list) => {
         // Delay by a microtask to workaround a bug in Safari where the
         // callback is invoked immediately, rather than in a separate task.
         // See: https://github.com/GoogleChrome/web-vitals/issues/277
         queueMicrotask(() => {
-          callback(list.getEntries() as PerformanceEntryMap[K]);
+          // sort entries to ensure they're in the right order
+          const entries = list.getEntries();
+          // Best to sort by end time (startTime + duration)
+          // See: https://github.com/w3c/performance-timeline/issues/224
```
mmocny, line 57: (for the future)

If/when Event and Layout Shift entries get paint timing, we can use the renderTime (if available) as the end time, rather than start+duration, to work around some of arbitrary nature (i.e. Layout shift start time is not paint aligned, and event timing duration is rounded).

In which case, you would also get a "stable" sort of all the entries that are for the same paint.

src/types/base.ts
```diff
@@ -72,14 +77,34 @@ export interface Metric {
...
    */
   navigationType:
     | 'navigate'
     | 'reload'
     | 'back-forward'
     | 'back-forward-cache'
     | 'prerender'
-    | 'restore';
+    | 'restore'
+    | 'soft-navigation';
```
mmocny, line 84: I know this was discussed/debated at some point, but the soft-nav entry itself exposes a navigationType which comes from the navigation api definitions.

I think this is fine here, but I wonder if long term design of this library could be more flexible even if the web-vitals definition keeps things simple?

Perhaps it could qualify as "attribution" data to expose:
- "soft" vs "hard" (and bfcache/prerender etc), and
- user intent (push, replace, traverse, reload, whatever)

(I might not have those dimensions right.)

src/onFCP.ts
```diff
@@ -88,5 +103,39 @@ export const onFCP = (
...
       });
     }
+
+    if (softNavsEnabled) {
+      // As first-contentful-paint is only reported once, we can handle soft
+      // navigations afterwards on their own for simplicity, as no need to
+      // observe both and sort the entries like for the other metrics
+      const handleSoftNavEntries = (entries: PerformanceSoftNavigation[]) => {
+        entries.forEach((entry) => {
+          handleEntries(po!.takeRecords() as FCPMetric['entries']);
```
mmocny: I think `po` is the original `paint` observer, so every new soft-nav will force re-reading initial FCP records?

Can we not drop this line and expect that observer will call `handleEntries` already by this point?

I *suspect* you will never get any entries back from takeRecords here, and the call to handleEntries will noop, but I'm not sure.

src/onFCP.ts
```diff
@@ -88,5 +103,39 @@ export const onFCP = (
...
+          // Cap FCP at 0 as it's possible the first paint is before the soft
```
mmocny: I dont understand this comment.  I think it might be baggage from back when the startTime was not the initial interaction time?

I think FCP (presentationTime - startTime) will always be > 0, even if paint happens before URL and this is "in the past".

src/onINP.ts
```diff
@@ -88,37 +99,85 @@ export const onINP = (
...
+
+      if (inp && inp._latency !== metric.value) {
+        metric.value = inp._latency;
+        metric.entries = inp.entries;
+        report();
+      }
+    };
+
+    const handleSoftNavEntry = (entry: PerformanceSoftNavigation) => {
+      handleEntries(po?.takeRecords() as INPMetric['entries']);
```
mmocny: I'm trying to understand why the manual call to `handleEntries` is needed here.

I think that `takeRecords()` will return a list of entries that are queued on the observer but haven't been dispatched yet.  This is useful for e.g. unload time recording-- but the `handleSoftNavEntry` is itself called from Perf observer callbacks, so we already have all entries available to us for the current "batch".

I think the result of this call will be roughly to process any entries that might have been added to the observer for the *next* paint(s) that aren't in the batch being reported now.

---

Besides, the call to handleEntries() immediately calls `whenIdleOrHidden` so I don't even think the result of this call does anything other than break up the typical Perf Observe reporting cadence... (and possible break per-paint recording, depending on how entries are created and queued).

src/onINP.ts
```diff
@@ -88,37 +99,85 @@ export const onINP = (
...
       // 123+ that if rolled out fully may make this no longer necessary.
       whenIdleOrHidden(() => {
+        // Only process entries, if at least some of them have interaction ids
+        // (otherwise run into lots of errors later for empty INP entries)
+        if (!entries.some((entry) => entry.interactionId)) return;
         for (const entry of entries) {
-          interactionManager._processEntry(entry);
+          if (entry.entryType === 'soft-navigation') {
+            handleSoftNavEntry(entry as PerformanceSoftNavigation);
+            continue;
```
mmocny, line 164: Nit: given the implicit observer for soft-navs when events are registered-- I would probably just have a list of support entryTypes here

if (soft-nav)
else if (event)
...
else
// unsupported

Rather than using `continue` and then fall back to assuming it is an event timing.

This is a minor style nit.

src/onLCP.ts
```diff
@@ -46,22 +51,140 @@ export const onLCP = (
...
+      entries: (
+        | LargestContentfulPaint
+        | InteractionContentfulPaint
+        | PerformanceSoftNavigation
+      )[],
+    ) => {
       // If reportAllChanges is set then call this function for each entry,
-      // otherwise only consider the last one.
-      if (!opts!.reportAllChanges) {
+      // otherwise only consider the last one, unless soft navs are enabled.
```
mmocny: This "use last one" is per call to handleEntries.

Even for LCP without the `reportAllChanges` flag, this is basically just hyper optimizing for the case where a single PerformanceObserver is called with multiple paints of data and there happened to be multuple candidates.  Right?

Otherwise you would have multiple distinct calls to handleEntries, anyway, with a single new LCP entry each.

---

Net/net, can this optimization just be removed?  I doubt it ever helps.

src/onLCP.ts
```diff
@@ -46,22 +51,140 @@ export const onLCP = (
...
+            ICPEntry.interactionId != metric.interactionId
+          ) {
+            continue;
+          }
+
+          // We're changing the shape to nest LCP entries within
+          // interaction-contentful-paint so handle both for now
+          const renderTime =
+            ICPEntry.largestContentfulPaint?.renderTime ||
+            (entry as InteractionContentfulPaint).renderTime ||
```
mmocny: Nit, why the cast here, should this not just be `ICPEntry.renderTime`?

src/onLCP.ts
```diff
@@ -46,22 +51,140 @@ export const onLCP = (
...
+            0;
+
+          // Paints should never be less than 0 but add cap just in case
+          value = Math.max(renderTime - entry.startTime, 0);
+
+          if (ICPEntry.largestContentfulPaint) {
+            lcpEntryManager._processEntry(ICPEntry.largestContentfulPaint);
+          }
+          if (ICPEntry.largestContentfulPaint) {
```
mmocny: Nit, merge with the previous if()?

src/onLCP.ts
```diff
@@ -88,24 +211,21 @@ export const onLCP = (
...
+        if (event.isTrusted && !isFinalized) {
           // Wrap the listener in an idle callback so it's run in a separate
           // task to reduce potential INP impact.
           // https://github.com/GoogleChrome/web-vitals/issues/383
-          whenIdleOrHidden(stopListening);
-          removeEventListener(event.type, stopListeningWrapper, {
-            capture: true,
+          whenIdleOrHidden(() => {
+            if (!isFinalized) {
+              handleEntries(po!.takeRecords() as LCPMetric['entries']);
```
mmocny: This call to handleEntries-- I think it is actually trying to read any still-unprocessed entries in the queue before flushing.  This one makes sense to me.

However, doesn't handleEntries wrap with `whenIdleOrHidden` and wouldn't that break the ability to process entries sync before finalizing?

mmocny (commented):
Its a large patch!  Started with the comments I noticed.

Didn't review tests or type defs.

src/attribution/onINP.ts
tunetheweb: At the minute the `attribution` property [is not optional](https://github.com/GoogleChrome/web-vitals/blob/main/src/types/inp.ts#L175-L179), though some of the properties within it are.

We could of course change it for this (since this would be a v6 major release so breaking changes are allowed), but I think I'd prefer to keep with the current model. It would be less of a breaking upgrade. Plus I expect these to be the uncommon ones, so worry people will assume there's an `attribution` (based on past experience) and not be ready for when one of these come in.

Alternatively, we could make more properties optional (the `interactionType`, `interactionTime`, and , `nextPaintTime` that I've had to make assumptions about here) and just not set them. Let me mull on that...

src/lib/initMetric.ts
tunetheweb: I dislike reassigning param variables to new values. But I agree this name isn't great so I've changed this to `navigationType` and changed the internal one to `_navigationType`.

src/lib/InteractionManager.ts
tunetheweb: Correct, added as per previous discussion in: https://github.com/GoogleChrome/web-vitals/issues/724#issuecomment-4857344586

src/lib/observe.ts
tunetheweb: For the first I think that's a larger redesign so going to leave that out of scope for here. The benefit of the current design is you can observe the 5 metrics separately. Though I'm not sure how often people do observe only a subset? I'm guessing the 3 Core Web Vitals are probably commonly observed as the main subset. Anyway. Out of scope.

Tried your second point but getting into typescript hell. Will leave this open and come back to it...

src/lib/observe.ts
tunetheweb: Noted. Resolving for now.

src/types/base.ts
tunetheweb: That's quite a breaking change to how we defined `navigationType` now, so I'd rather keep it out of this. It's a pity we both used `navigationType` for this :-(

But I do see the value in exposing your `navigationType` (push, replace, traverse, reload, whatever) somewhere in the API. For hard nav, LCP and FCP have the original navigation object, but we don't make that available for Soft LCP and Soft FCP cause it's a little tricky. Will see if I can add that, which would resolve that.

src/onFCP.ts
tunetheweb: Yeah this is copied from the other metrics, and I suspect it will always noop for FCP, but I also don't think we need it in the others now either (since we now observe then together and so should be in order) so suspect this is just old code.

Going to test just taking this out for all metrics...

src/onFCP.ts
tunetheweb: Ah yes. I might keep the capping anyway as a safety check, but update the comment.

src/onINP.ts
tunetheweb: Yup, as per above comment, I don't think this is needed anymore and is from a previous iteration when we didn't observe them both together. Will test if it can be removed.

src/onINP.ts
tunetheweb: That makes sense here and for CLS. For LCP it makes less sense since there's a lot more code so "early exiting" is more readable. And I think I implemented LCP first and just did similar code here.

Could change INP but think I'll leave as is for consistency.

src/onLCP.ts
tunetheweb: Agreed but see no harm in keeping it since it's existing code. I have update the comment to make this clearer though.

src/onLCP.ts
tunetheweb: Good spot. Think I added `ICPEntry` specifically for this reason and then missed on clean up. Done.

src/onLCP.ts
tunetheweb: Done.

src/onLCP.ts
tunetheweb: No it doesn't. It does for INP, but not for LCP.

For LCP the `whenIdleOrHidden` is only used for for `click` and `keyboard` events that finalise LCP (and this could also be INP hence why we delay until idle). For regular LCP processing we don't wait for idle time.

src/lib/observe.ts
tunetheweb: Cleaned it up in https://github.com/GoogleChrome/web-vitals/pull/308/commits/05d92b1ef0bcdd88231c37c9456dd64999b4de60 so each metric passes the types it wants to observe, rather than `observe` function expanding on the single type they pass.

src/onFCP.ts
tunetheweb: Done in https://github.com/GoogleChrome/web-vitals/pull/308/commits/3af152180e0ea0c6b3c2b3ffd4668d4d1fe01935

src/onINP.ts
tunetheweb: Done in https://github.com/GoogleChrome/web-vitals/pull/308/commits/3af152180e0ea0c6b3c2b3ffd4668d4d1fe01935

src/types/base.ts
tunetheweb: Added `navigationEntry` to the attribution object for soft navs now too, so those types can be looked up.

src/types/base.ts
```diff
@@ -54,6 +54,11 @@ export interface Metric {
    */
   id: string;
 
+  /**
+   * The interactionId that started this interaction for soft navigations
+   */
+  interactionId?: number;
```
philipwalton: It feels strange to me to include this for all metric objects when it's only relevant to metrics after a soft navigation.

It also seems like something that's mostly relevant to attribution use cases (e.g. why would you care about identifying the navigation trigger if you're not doing attribution?).

src/types/base.ts
```diff
@@ -72,14 +77,34 @@ export interface Metric {
...
     | 'prerender'
-    | 'restore';
+    | 'restore'
+    | 'soft-navigation';
+
+  /**
+   * The navigationId the metric happened for. This is particularly relevant for soft navigations where
+   * the metric may be reported for a previous URL.
+   */
+  navigationId: number;
```
philipwalton, line 90: This library treats bfcache restores as a "navigation type", yet (based on looking at the code below) it doesn't give them their own `navigationId` value, nor does it track a `navigationStartTime`.

That doesn't seem ideal to me. I wonder if it would make more sense to generate these values rather reuse the values from the most recent soft nav.

src/attribution/onINP.ts
```diff
@@ -383,6 +383,43 @@ export const onINP = (
...
+    // fall back on so we report dummy values when the interactionCount has
+    // gone up, even if no entry was emitted.
+    // See https://github.com/GoogleChrome/web-vitals/issues/724
+    if (
+      metric.entries.length === 0 &&
+      (metric.navigationType === 'soft-navigation' ||
+        metric.navigationType === 'back-forward-cache')
+    ) {
+      const navStartTime = metric.navigationStartTime;
+      if (navStartTime) {
```
philipwalton: There's a bug here that can throw an error.

This conditional early returns the below attribution data, but it does so ONLY if `navStartTime` is set (and non-zero). However, after a bfcache restore, the `navigationStartTime` value can be zero, which means the code will continue and it will throw an error below when trying to access `metric.entries[0]` below because there won't be any entries (since inside this conditional we know `metric.entries.length === 0`).

I was able to reproduce this error when testing locally.

src/onLCP.ts
```diff
@@ -88,24 +231,21 @@ export const onLCP = (
...
+              if (!softNavsEnabled) {
+                po!.disconnect();
+                removeEventListener(event.type, finalizeLCP);
```
philipwalton: Line 258 below calls `addEventListener()` with `{capture: true}` set, but then this `removeEventListener()` call doesn't include that flag, so it won't be removed.

src/onINP.ts
```diff
@@ -88,37 +100,93 @@ export const onINP = (
...
       // Queue the `handleEntries()` callback in the next idle task.
       // This is needed to increase the chances that all event entries that
       // occurred between the user interaction and the next paint
       // have been dispatched. Note: there is currently an experiment
       // running in Chrome (EventTimingKeypressAndCompositionInteractionId)
       // 123+ that if rolled out fully may make this no longer necessary.
       whenIdleOrHidden(() => {
+        // Only process entries, if at least some of them have interaction ids
+        // (otherwise run into lots of errors later for empty INP entries)
+        if (!entries.some((entry) => entry.interactionId)) return;
```
philipwalton: There was a time when `first-input` entries would report an `interactionId` of `0`. I just want to confirm that that's no longer the case (and hasn't been the case long enough to not worry about it).

src/onLCP.ts
```diff
@@ -46,22 +51,152 @@ export const onLCP = (
...
+    const initNewLCPMetric = (
+      navigation?: Metric['navigationType'],
+      interactionId?: number,
+      navigationId?: number,
+      navigationURL?: string,
+      navigationStartTime?: number,
+    ) => {
+      metric = initMetric(
+        'LCP',
+        0,
```
philipwalton: I'm not sure if this matters in practice, but the default value for `initMetric()` is `-1` in order to avoid reporting metric values if no meaningful value has been set.

That's why CLS initializes with `0` (because no layout shifts does mean zero CLS), but no LCP/ICP entries should not mean a soft LCP of 0, right?

src/onFCP.ts
```diff
@@ -88,5 +109,59 @@ export const onFCP = (
...
+              const firstKey = fcpEntryManager._softNavigationEntryMap
+                .keys()
+                .next().value;
+              if (firstKey !== undefined) {
+                fcpEntryManager._softNavigationEntryMap.delete(firstKey);
+              }
+            }
+          }
+
+          // Cap FCP at 0. It should never be less, but better safe than sorry.
```
philipwalton: Nit: I interpret "cap" to mean "max" or "top", not "min". Maybe "Clamp" instead?

src/onLCP.ts
```diff
@@ -88,24 +231,21 @@ export const onLCP = (
...
-        handleEntries(po!.takeRecords() as LCPMetric['entries']);
-        po!.disconnect();
-        report(true);
-      });
-
-      // Need a separate wrapper to ensure the `runOnce` function above is
-      // common for all three functions
-      const stopListeningWrapper = (event: Event) => {
-        if (event.isTrusted) {
+      const finalizeLCP = (event: Event) => {
```
philipwalton, line 212: AI highlighted a race condition in this function, see #760 for the fix.

src/onLCP.ts
philipwalton: I agree that there's no issue with processing the entries async (at least I don't think there is), but I think there could be an issue with processing the soft nav entry async, right?

If the click the finalizes LCP is a link that triggers a soft nav, then delaying the processing of those entries until idle time will also delay the processing of the soft nav logic itself.

Could that have any unintended consequences?

src/onINP.ts
```diff
@@ -88,37 +100,93 @@ export const onINP = (
...
         for (const entry of entries) {
-          interactionManager._processEntry(entry);
+          if (entry.entryType === 'soft-navigation') {
+            handleSoftNavEntry(entry as PerformanceSoftNavigation);
```
philipwalton, line 163: Similar to my comment for INP, I'm not sure whether there might be some unintended consequences of handling the soft nav entries async in an idle period.

I'm not familiar enough with the API to know whether order or timing matters, but I just wanted to flag this in case it does.

src/attribution/onLCP.ts
```diff
@@ -115,10 +140,11 @@ export const onLCP = (
           // Cap at LCP time (videos continue downloading after LCP for example)
           metric.value,
           Math.max(
-            lcpRequestStart,
+            lcpRequestStart - activationStart,
```
philipwalton: `lcpRequestStart` already subtracts `activationStart` above. Does it need to do it here too?

src/onINP.ts
philipwalton: @tunetheweb that commit updated FCP and LCP but not INP, was the intentional?

Given that `handleEntries()` defers processing of the entries to an idle task, I'm concerned that there's a race condition here. The calls to `updateINPMetric()`, `report()`, and `initNewINPMetric()` all run synchronously in this function before the entries are processed in the idle task.

Couldn't that result in those entries getting added to the newly-created INP metric object?

src/types/base.ts
philipwalton: Just tested, and in Canary when the `back-forward-cache-restoration` entry is emitted, it has its own `navigationId` value, that subsequent ICP entries (in theory) should get.

If this is a direction that Chrome will go in the future, I wonder if it makes sense for web-vitals to do it now?

cc: @mmocny

src/onLCP.ts
```diff
@@ -46,22 +51,152 @@ export const onLCP = (
...
+        onReport,
+        metric,
+        LCPThresholds,
+        opts!.reportAllChanges,
+      );
+      // Reset the finalized flag
+      isFinalized = false;
+      // If it's a soft nav, then need to reset the visibilityWatcher
+      if (navigation === 'soft-navigation') {
+        visibilityWatcher = getVisibilityWatcher(true);
```
philipwalton, line 92: For bfcache restores, the getVisibilityWatcher module itself takes care of resetting. I wonder if it should do the same for soft navs?

Another thing to consider is that the visibility watcher logic makes a lot of global assumptions that apply to all consumers. Do those assumptions hold for a site that is tracking both regular LCP and soft-LCP at the same time?

I think they do, but it may make things less complicated if we had a regular visibility watcher and a soft-nav-aware visibility watcher, WDTY?

Also, what about FCP? I didn't notice a similar visibility watcher reset call in that module?

philipwalton:
Hey @tunetheweb thanks for sticking with this PR all these years!

I tested this out locally on my site and a few test sites and didn't find any issues. That said, I did ask Gemini to do a deep dive and look for potential issues and it did find several that I manually verified and have commented here.

Since Gemini already had the context, I asked it to also implement the fixes, which I've submitted as the following 4 PRs:

- #758 
- #759
- #760
- #761

Like I said, I did manually verify that the first three were actual problems. The 4th one is mostly just stylistic stuff, but all those suggestions seem reasonable to me.

My last question is about the tests. Right now I see the soft nav tests are being skipped if the browser doesn't support soft navs (which no stable browsers currently do). Have you run these yourself against canary or with an experimental flag enabled?

src/onLCP.ts
tunetheweb: I don't think so since the entries are in order. So it will simply process them in the order during idle time.

One thing I did spot is the type casting is incorrect though. This should be:

```suggestion
              handleEntries(po!.takeRecords() as as [InteractionContentfulPaint | LargestContentfulPaint | PerformanceSoftNavigation]);
```

src/types/base.ts
tunetheweb: The reason I do this is that `interactionId` is almost synonymous (or even better in fact!) to `navigationId` since `navigationId` may be "incorrect".

For example, an iCP that happens after the soft nav navigation interaction, but before the URL update and/or paint will have the old `navigationId` in it's performance entry, even though it's the LCP for the new `navigationId`, so users may want to concentrate on the `interactionId` more so than the `navigationId`. Now this library actually correct this for the emitted soft LCP entry, so the metric's `navigationId` will be correct, but the point still stands that `navigationId` is still an important slicing identifier for slicing the performance timeline.

Additionally, this could allow additional slicing for INP and CLS, or even alternative timeline slicing if the library ever supported manual slicing (a frequent ask, but one I'm not keen to support right now).

src/types/base.ts
tunetheweb: I agree this is not ideal. And yes Chrome has [experimented with giving it it's own id](https://github.com/w3c/navigation-timing/issues/184#issuecomment-1343822156), but that's not being actively pursued at the moment IIUC. To be honest that's why I'd rather keep this as is, to avoid muddying the waters with a fake `navigationId` that can't be used for anything (it can't be used to look up the equivalent performance entry to get more info for example).

I do agree resetting the `navigationStartTime` would be good though so will make that change.

src/attribution/onINP.ts
tunetheweb: With above fix to reset `navigationStartTime` the bug can now no longer happen. Another reason to add this!

AFAIK this should only have happened on bfcache restore when there is no interaction, but the interaction count has increased (so basically there was an interaction < 8ms that ideally FID would cover but can't since it was used by the hard nav and only fires once). So defaulting back to bfcache restore time is the best we've got (alternatively we could use current time or average of the two or something else, but I think using bfcache restore time is at least predictable).

src/onLCP.ts
tunetheweb: Fixed.

src/onINP.ts
tunetheweb: I think that's the case. But we still use it in the interaction manager so should be consistent.

However, since we're doing this in the interaction manager, I don't even know why we need this line! So going to remove it and see if I can find the "lots of errors" and the reason why. It was added 4 months ago in [this commit](https://github.com/GoogleChrome/web-vitals/commit/4d744ff4a0584ae15551f36afc78274324a19ffe#diff-4556a1b89955ab6eacaf744b522e797726168ef577c9975d1febe69ec053e2ff) labelled "Refactor to handle buffered values better" but I'm not sure why...

src/onLCP.ts
tunetheweb: The only time we should reinitialise the metric is either on soft nav (which requires a paint) or on bfcache (which fakes with a paint after a doubleRaf()). So in both cases there _should_ always be an LCP. It should also be > 0. So I don't think initing it to 0 really matters but will try -1 just for consistency.

src/onFCP.ts
tunetheweb: Done.

src/onLCP.ts
tunetheweb: Oh based on below FCP comment, I wonder if I did this to ensure negative LCPs were still reported as an LCP of 0 for cases when the first paint was before the soft nav time?

Now we've moved soft nav start time to interaction start time, this is no longer a concern. So we can go back to using -1.

src/attribution/onLCP.ts
tunetheweb: Nope. Not sure why I made this change. Removing.

src/onLCP.ts
tunetheweb: > For bfcache restores, the getVisibilityWatcher module itself takes care of resetting. I wonder if it should do the same for soft navs?

I don't think we should because bfcache restores are event driven, while soft navs are performance timeline driven. For soft navs need to preserve the ordering (hence why we observe all the performance timeline entries together).

Plus a soft nav is not a "visibility change". It could be argued it's kinda like `pagehide`/`unload` but it's not.

tunetheweb: Resolving but re-open if you disagree and wanna discuss further.

src/onINP.ts
tunetheweb: > @tunetheweb that commit updated FCP and LCP but not INP, was the intentional?

Nope. Removed now!

> Given that `handleEntries()` defers processing of the entries to an idle task, I'm concerned that there's a race condition here. The calls to `updateINPMetric()`, `report()`, and `initNewINPMetric()` all run synchronously in this function before the entries are processed in the idle task.
> 
> Couldn't that result in those entries getting added to the newly-created INP metric object?

With the removal of this for `handleSoftNavEntry` I don't think this is a concern anymore. Let me know if you disagree but resolving for now.

src/attribution/onINP.ts
tunetheweb: @philipwalton any thoughts here?

src/types/base.ts
tunetheweb: Resolving for now, but reopen if you wanna discuss this more.

src/types/base.ts
tunetheweb: Resolving for now, but reopen if you wanna discuss more.

src/onINP.ts
tunetheweb: No issue as the events are guaranteed in order and we process them in order. Resolving.

src/onLCP.ts
tunetheweb: I've gone with a simpler fix here: https://github.com/GoogleChrome/web-vitals/pull/308/commits/613f9acca5d4cb18cee087a6529ef9a1bb9ce822

tunetheweb:
> Hey @tunetheweb thanks for sticking with this PR all these years!

Thank you for catching up years of work and giving great feedback!

> I tested this out locally on my site and a few test sites and didn't find any issues. That said, I did ask Gemini to do a deep dive and look for potential issues and it did find several that I manually verified and have commented here.
> 
> Since Gemini already had the context, I asked it to also implement the fixes, which I've submitted as the following 4 PRs:

I accepted two of those, but did alternatives to the other two.

> My last question is about the tests. Right now I see the soft nav tests are being skipped if the browser doesn't support soft navs (which no stable browsers currently do). Have you run these yourself against canary or with an experimental flag enabled?

Yup. You'll notice I've a check box in [the first comment above](https://github.com/GoogleChrome/web-vitals/pull/308#issue-1549938446) to:

> Agree final WDIO setup (needs to wait for a supporting version of Chrome to be available in WDIO)

At the minute I'm waiting for Chrome 150 to land in WDIO (which I honestly thought would be there now!) and then we can test with experimental support (currently commented out). Though the final shape won't be until Chrome 151 so might be best just waiting for that, which will be stable in a couple of weeks.

In the meantime, yes I've been testing locally on Canary (**without** the experimental flag enabled), and also Beta since it hit 151.

src/attribution/onINP.ts
philipwalton: I lean toward making the properties optional. I think'd be fine with this:

```js
const attribution: INPAttribution = {
  // interactionTarget: undefined,
  // interactionType: undefined,
  interactionTime: navStartTime,
  nextPaintTime: navStartTime + metric.value,
  processedEventEntries: [],
  longAnimationFrameEntries: [],
  inputDelay: 0,
  processingDuration: 0,
  presentationDelay: metric.value,
  loadState: getLoadState(navStartTime),
  // longestScript: undefined,
  // totalScriptDuration: undefined,
  // totalStyleAndLayoutDuration: undefined,
  // totalPaintDuration: undefined,
  // totalUnattributedDuration: undefined,
};
```

Arguably we should just omit the properties rather than make them `undefined`, as that would reduce the code size.

I definitely don't think we should set `interactionType` to `'pointer'` if we don't know it was actually a pointer.

src/types/base.ts
philipwalton: What about putting this in the attribution object?

src/types/base.ts
philipwalton: I'm fine with keeping it as-is, to be consistent with the fact that other entries in the performance timeline will get `navigationId` values that Chrome assigns to them.

That said I *really* wish Chrome would change here and update the `navigationId` after a bfcache restore (and emit some entry to expose it).

src/attribution/onINP.ts
tunetheweb: Probably shouldn't include a dummy `interactionTime` and `nextPaintTime` then either? Will make them optional and remove them from here.

src/onLCP.ts
philipwalton: I think the only time it would matter is if  the libraries does other side-effecting things in response to a soft nav entry, e.g. updating some global state object or URL value after a soft nav happens (e.g. similar to what the visibilityState watcher does).

I don't think the library does that currently, but it'll be an important thing to watch out for in the future.

philipwalton:
@tunetheweb I noticed for https://github.com/GoogleChrome/web-vitals/pull/308/changes/63093732d63adbec56639d0114c9385d09f77316 the comments about the made-up values is still there:

```
      // - Assume interactionType as pointer as the most common
      // - Assume interactionTime of nav start time
      // - Assume nextPaintTime as interactionTime + length
```

Also, in the type definitions, I think it would be helpful to add a short note to the bottom of each explaining why it's options. E.g. "This value may not be set in cases where the event duration was less than the threshold, and thus no entry was dispatched."

src/types/base.ts
tunetheweb: That would require quite a bit of refactoring since the attribution object won't have access to this anymore. It basically required adding it to the XXXEntryManager (and not all metric have one of these) to store it to enable the attribution to pick it up later. I did that for FCP and LCP to give them access to the navigation entry in the attribution object, but would now need to do it again for CLS and INP.

Even without that, it also seems a little weird on it's own in there when all the other `navigationXXX` values are in a different place.

I could rename to `navigationInteractionId`. Would that make you more comfortable to have it in the main, non-attribution Metric type?

tunetheweb:
Good spot! Fixed in https://github.com/GoogleChrome/web-vitals/pull/308/commits/3807704b47e5382d42a15fb9a4a9a92fc2362234

src/types/base.ts
tunetheweb: Renamed in https://github.com/GoogleChrome/web-vitals/pull/308/commits/0f35f05e683d0e7d655d105e19408e9e5c4f7437 but kept in main metric object. Hopefully that makes more sense now why it's included rather than seeming somewhat random.

philipwalton: I guess given that `navigationType` is already on the main metric object, maybe it makes sense to keep them all there. Arguably `navigationType` should have been in attribution as well, but I don't feel strongly enough about it to suggest moving it at this point.

src/types/base.ts
```diff
@@ -72,14 +72,39 @@ export interface Metric {
...
+
+  /**
+   * The interactionId that started this interaction for soft navigations
```
philipwalton:
```suggestion
   * For metrics specific to a soft navigation, the interactionId of the interaction that triggered that soft navigation.
```
I think this description could be more clear (feel free to reword).

tunetheweb: Done in https://github.com/GoogleChrome/web-vitals/pull/308/commits/c0715f4b134aab08d65d619fc54aea0f219ba196

Mimori256:
just wondering if there is a possible ordering bug in `onCLS` when `layout-shift` and `soft-navigation` entries are delivered in the same `PerformanceObserver` callback.

`handleSoftNavEntry()` calls `report(true)` and then resets the CLS metric, but pending layout-shift session data is only folded into `metric.value`[after the entry loop in `handleEntries()`

https://github.com/GoogleChrome/web-vitals/blob/606834c5d870229631111e4891b02e5bcd90ab57/src/onCLS.ts#L98-L107

So for a batch like:

```js
[layoutShiftEntry, softNavigationEntry]
```

the layout shift has already been processed into `layoutShiftManager._sessionValue`, but `metric.value` is still stale when `handleSoftNavEntry()` calls `report(true)`. Then `initNewCLSMetric()` resets `_sessionValue`, so that shift can be omitted from the final CLS report for the previous navigation.

INP avoids the same issue by calling `updateINPMetric()` before the forced soft-navigation report but CLS does not have the equivalent e.g. extracting the post-loop fold into an `updateCLSMetric()` helper and calling it before `report(true)` in `handleSoftNavEntry()`.

https://github.com/GoogleChrome/web-vitals/blob/606834c5d870229631111e4891b02e5bcd90ab57/src/onINP.ts#L126-L140

tunetheweb:
Oh you're correct. Fixed with https://github.com/GoogleChrome/web-vitals/pull/308/commits/17f25c75b997d1caa4ff27b34e5ed69b081645c6

PR #323, opened 2023-02-26
Export metric rating thresholds, add explicit `MetricRatingThresholds` type

robatron:
# What?

This PR exports the existing rating thresholds for all metrics, e.g.,

```ts
import {CLSThresholds} from 'web-vitals';

console.log(CLSThresholds); // [ 0.1, 0.25 ]
```

... and defines the existing rating thresholds format with an explicit `MetricRatingThresholds` type: 

```ts
/**
 * The thresholds of metric's "good", "needs improvement", and "poor" ratings.
 *
 * | Metric value    | Rating              |
 * | --------------- | ------------------- |
 * | ≦ [0]           | "good"              |
 * | > [0] and ≦ [1] | "needs improvement" |
 * | > [1]           | "poor"              |
 */
export type MetricRatingThresholds = [number, number];
```

# Why?

Exporting typed metric rating thresholds enables referencing threshold values directly, reducing potential duplication issues and improving long-term codebase maintainability for consumers 💪 

# Details

### Exported metric rating thresholds

The thresholds of each metric's "good", "needs improvement", and "poor" ratings are made available as a top-level exports:

```ts
import {CLSThresholds, FCPThresholds, FIDThresholds, INPThresholds, LCPThresholds, TTFBThresholds} from 'web-vitals';

console.log(CLSThresholds);  // [ 0.1, 0.25 ]
console.log(FCPThresholds);  // [ 1800, 3000 ]
console.log(FIDThresholds);  // [ 100, 300 ]
console.log(INPThresholds);  // [ 200, 500 ]
console.log(LCPThresholds);  // [ 2500, 4000 ]
console.log(TTFBThresholds); // [ 800, 1800 ]
```

These exported metric rating thresholds are simply the existing private thresholds hoisted to module scope and exported alongside their `on*()` functions. E.g.,

```ts
// File: src/onCLS.ts

// Existing threshold hoisted to module scope and exported
export const CLSThresholds: MetricRatingThresholds = [0.1, 0.25];

// Existing `on*()` function
export const onCLS = (onReport: CLSReportCallback, opts?: ReportOpts) => { /* ... */}
```

### New `MetricRatingThresholds` type

This new type simply defines the existing format explicitly (formerly `number[]`):

```ts
export type MetricRatingThresholds = [number, number];
```

This type represents the thresholds of metric's "good", "needs improvement", and "poor" ratings:

| Metric value    | Rating              |
| --------------- | ------------------- |
| ≦ [0]           | "good"              |
| > [0] and ≦ [1] | "needs improvement" |
| > [1]           | "poor"              |

I.e., 

- Metric values up to and including [0] are rated "good"
- Metric values up to and including [1] are rated "needs improvement"
- Metric values above [1] are "poor"


tunetheweb:
Seems a reasonable change, though it does make the built files ever so slightly bigger since we're exporting 6 new values now.

A couple of points:
1. Can we add something to the README so this isn't a hidden function?
2. Can we add some tests to ensure this isn't accidentally reverted or broken in future?

For the second, maybe a tests/unit.js file with the following:

```js
import {CLSThresholds, LCPThresholds} from 'web-vitals';
import assert from 'assert';

assert.strictEqual(CLSThresholds[0], 0.1);
assert.strictEqual(CLSThresholds[1], 0.25);
assert.strictEqual(LCPThresholds[0], 2500);
assert.strictEqual(LCPThresholds[1], 4000);
```

I'm in two minds about putting all 12 test cases in there (two for each metric). On one hand it's a bit repetitive, but on the other it does ensure they are all exported properly so probably for the best despite it being repetitive. Besides I don't expect these thresholds to change often. WDYT?

robatron:
README updated!

I think it's a good idea to verify all thresholds are being exported correctly even if it's a little repetitive, especially since they're not expected to change often. I [added unit tests](https://github.com/GoogleChrome/web-vitals/pull/323/commits/86194e67c7f3745565f6057a488b154387a9c080) for this.

src/onCLS.ts
```diff
@@ -22,7 +22,15 @@ import {doubleRAF} from './lib/doubleRAF.js';
...
-import {CLSMetric, CLSReportCallback, ReportOpts} from './types.js';
+import {
+  CLSMetric,
+  CLSReportCallback,
+  MetricThresholds,
+  ReportOpts,
+} from './types.js';
+
+/** Thresholds for CLS. See https://web.dev/cls/#what-is-a-good-cls-score */
+export const CLSThresholds: MetricThresholds = [0.1, 0.25];
```
brendankenny, line 33: FWIW exporting like this means that users will be able to change the thresholds used, which is likely not a use case we want to get into supporting (maybe it'll be rare if ever, but, https://xkcd.com/1172/).

Could copy or export a function that returns a copy.

It may be worth returning a different format at that point, e.g. an object with named properties could be more self explanatory than `[number, number]`.

package.json
```diff
@@ -107,6 +107,7 @@
     "test": "npm-run-all build -p -r test:*",
```
tunetheweb, line 107: I think this now needs to change to this:


```suggestion
    "test": "npm-run-all build test:unit -p -r test:e2e test:server",
```

tunetheweb: This is because the `-p -r` runs them in parallel (`-p`) but also kills the other when the first finishes (`-r`) - which is basically used to kill the (otherwise never completing) server once e2e tests finish.

Without moving the `test:unit` to it's own section, it is the first to complete and kills the `e2e` and `server`, ending the test far too early.

src/onCLS.ts
robatron: > FWIW exporting like this means that users will be able to change the thresholds used, which is likely not a use case we want to get into supporting (maybe it'll be rare if ever, but, https://xkcd.com/1172/).

Lol, good call. I'll update the exported thresholds to prevent changes.

- [ ] Disallow changes to internal threshold values via exports

> It may be worth making the external version a different format at that point, e.g. an object with named properties could be more self explanatory than [number, number].

I agree it would be better for the format to be more self-explanatory. (Although @tunetheweb [had some concerns](https://github.com/GoogleChrome/web-vitals/pull/323#issuecomment-1446277301) about increased build file sizes.) How about something like this? 

```ts
/**
 * The ranges defining a metric's "good", "needs improvement", and "poor"
 * thresholds:
 *
 * - Metric values ≦ good are "good"
 * - Metric values > good and ≦ needsImprovement are "needs improvement"
 * - Metric values > needsImprovement are "poor".
 */
export type MetricThresholds = {
  good: number;
  needsImprovement: number;
};
```

package.json
robatron: Ah, good to know. [Updated](https://github.com/GoogleChrome/web-vitals/compare/86194e67c7f3745565f6057a488b154387a9c080..b099e8606a611b3d886736351cb60d27b440b2ce).

src/onCLS.ts
tunetheweb: I wonder if we should instead have a general `getMetricThresholds()` function exported that would prevent changing of the values and also allow only a single extra export? For example, `getMetricThresholds('LCP')`, or even `getMetricThresholds('LCP','good')`. That may require moving all the thresholds to that common function and referencing it from each `onXXX` function.

Then again, I'm wondering if this change is now worth the effort given it's not quite a simple as originally thought?

robatron: I think a metric rating threshold getter function makes sense and solves both problems. 

I added a new [`getMetricRatingThresholds()` function](https://github.com/GoogleChrome/web-vitals/pull/323/files#diff-562e9b968ecef139d0bc5134e0b9ba0efad5c52017c1ddbf3419ef6864cf0a4fR43) which takes a metric name and returns a [`MetricRatingThresholds` type](https://github.com/GoogleChrome/web-vitals/pull/323/files#diff-c8915aba955333a21cf5d5713fd07a6afbd7c06aeb0e4f97bfac15618fcabdd4R110) (updated to the above proposed format). Also moved all the [threshold values](https://github.com/GoogleChrome/web-vitals/pull/323/files#diff-562e9b968ecef139d0bc5134e0b9ba0efad5c52017c1ddbf3419ef6864cf0a4fR4) alongside `getMetricRatingThresholds()` in [a new module](https://github.com/GoogleChrome/web-vitals/pull/323/files#diff-562e9b968ecef139d0bc5134e0b9ba0efad5c52017c1ddbf3419ef6864cf0a4f). 

WDYT?

.nvmrc
```diff
@@ -0,0 +1 @@
+lts/*
```
robatron, line 1: The [built-in node.js test runner](https://nodejs.org/docs/latest-v18.x/api/test.html) this PR depends on is available starting in the latest LTS version, [v18](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V18.md#test-runner-module-experimental). This file allows `nvm` to [automatically activate](https://github.com/nvm-sh/nvm#nvmrc) this version for local development.

tunetheweb (approved):
This LGTM.

@robatron could you update the initial comment of this PR since it's changed a good bit since you opened this PR. Just in case anyone comes across it later and is confused.

@brendankenny could you give this another review?

robatron:
> could you update the initial comment of this PR since it's changed a good bit since you opened this PR

Done! Also squashed the commits (→ https://github.com/GoogleChrome/web-vitals/pull/323/commits/3909d101711f83bb89f4eac8d67e1b1e40677f70) and should be ready to merge anytime 👍

.nvmrc
```diff
@@ -0,0 +1 @@
+lts/Hydrogen
```
robatron, line 1: The [built-in node.js test runner](https://nodejs.org/docs/latest-v18.x/api/test.html) this PR depends on is available starting in the latest LTS version, [v18](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V18.md#test-runner-module-experimental). This file allows `nvm` to [automatically activate](https://github.com/nvm-sh/nvm#nvmrc) this version for local development.

tunetheweb:
> > could you update the initial comment of this PR since it's changed a good bit since you opened this PR
> Done!

Ta!

> Also squashed the commits (→ [3909d10](https://github.com/GoogleChrome/web-vitals/commit/3909d101711f83bb89f4eac8d67e1b1e40677f70))

FYI, for next time, we Squash and Merge, so no need to do this (and actually would prefer not to force push so easier to track changes in the PR!).

> and should be ready to merge anytime 👍

Cool, would like @brendankenny 's review before merging.

robatron:
Hey @brendankenny , any other changes you'd like me to make before @tunetheweb merges?

philipwalton:
@robatron sorry to jump in here at the last minute, but could you explain to me a bit more about the use case for wanting to know the metric thresholds outside of the report callback function?

I'm concerned that this is adding functionality that 99.9% of users won't need and probably shouldn't use, so I just wanted to make sure I better understand the use case.

Another implementation option, given that this likely isn't an API that most users will need, would be to create a new top-level module export that doesn't contribute to the main bundle, e.g.:

```js
import {onLCP} from 'web-vitals';
import {LCPThreshold} from 'web-vitals/thresholds';
```

Potentially we could even have a generic `web-vitals/meta` export that could house information like this that prevents duplication but doesn't contribute to the main bundle size.

WDYT?

philipwalton:
Ok, I thought about this a bit more, and I think having a seperate export (e.g. `web-vitals/thresholds`) might be overkill, and it would also add complexity to the build, so keeping the thresholds in the main bundle seems fine, but I do think we want to make sure they're as small as possible.

Also, I'm not sure the format `{good: X, needsImprovement: Y}` is any more clear that `[X, Y]`, given that both of these only contain two of the three ratings and neither of them clarify whether the "good" threshold is inclusive or exclusive of X. I think `{good: X, poor: Y}` is slightly better but still has all of the problems I just mentioned.

Given this, and the fact that `[X, Y]` will product the smallest bundle size and least amount of code changes, I think I'm leaning toward that. What do others think?

And if we go this route, I think it makes the most sense for the metric modules to export the thresholds, so that folks only using one or two metrics don't pay the cost of all of the thresholds.

So I'm thinking something like:

```js
// File: ./onCLS.js

// The thresholds are a top-level module export for each metric.
export const LCPThresholds = [2500, 4000];

// The metric function exported as it is today.
export const onLCP = () => { ... };
```

Also, I understand @brendankenny's concern about people modifying the thresholds, but honestly I'm not too worried about that. If we do want to prevent that from happening, we could always export `Object.freeze([2500, 4000])`, which would compress better than a function that returns an object. But having it as module export already prevents reassignment, and an array is exactly as mutable as the `{good: X, needsImprovement: Y}` object in the current changes, so my preference would be to just export the array as is.

tunetheweb:
I do think being able to export the thresholds is a good thing. And agree the Array without labels is probably sufficient.

> So I'm thinking something like:

But that would be 6 exports (with quite long names), whereas a function to get the exports (based in a `metric` parameter) is only one export. So surely - good bit smaller?

>  so that folks only using one or two metrics don't pay the cost of all of the thresholds.

That cost seems pretty small to me isn’t it?

philipwalton:
> But that would be 6 exports (with quite long names), whereas a function to get the exports (based in a `metric` parameter) is only one export. So surely - good bit smaller?

All modern bundlers will remove those exports when building application code, and assign them to a local variable (which could be minified, so in that sense they're basically free).

I just tried a build that exported all of the symbols I outlined above and it increased the brotli'd size of the final bundle by 38 bytes, so that would be the maximum size increase if pulling the build from a CDN, but if bundling it locally then it shouldn't increase it at all (I'm pretty sure).

tunetheweb:
Compression is amazing! Funny, I was just making that point (to compare compressed size of a bundle, rather than uncompressed size) today and then forgot my own advice!

OK then so back to original implementation basically @robatron . Apologies for all the iterations here and ending up back to the beginning basically!

robatron:
Lol. Ok, back to [the original implementation](https://github.com/GoogleChrome/web-vitals/pull/323/commits/e935df08ba38dd0ffaad67c517b641fa2e258d77) then! (Another reason I shouldn't have [pre-squashed my commits](https://github.com/GoogleChrome/web-vitals/pull/323#issuecomment-1450722745) 😆)

I reverted the relevant code and updated the PR title and description accordingly. 

WDYT?

brendankenny:
> Also, I understand @brendankenny's concern about people modifying the thresholds, but honestly I'm not too worried about that. If we do want to prevent that from happening, we could always export `Object.freeze([2500, 4000])`, which would compress better than a function that returns an object. But having it as module export already prevents reassignment, and an array is exactly as mutable as the `{good: X, needsImprovement: Y}` object in the current changes, so my preference would be to just export the array as is.

Hi @philipwalton :)

FWIW I was thinking more like
```js
const thresholds = [0.1, 0.25];

export const onCLS = () => {
  // use `thresholds` as before.
}

export const CLSThresholds = [...thresholds];
```
at which point
```
export const CLSThresholds = {good: thresholds[0], poor: thresholds[1]};
```
doesn't matter much, size-wise. This would prevent you having to consult the docs on which number is which, at least, though I do take the point that we don't really have names we use for the thresholds, and those names conflate thresholds with buckets.

robatron:
> FWIW I was thinking more like

Yeah, I [had implemented something similar](https://github.com/GoogleChrome/web-vitals/pull/323/commits/bd6842285a412cd3266a730a6e658f32bbda5cfb#diff-562e9b968ecef139d0bc5134e0b9ba0efad5c52017c1ddbf3419ef6864cf0a4fR54) in the [previous implementation](https://github.com/GoogleChrome/web-vitals/pull/323/commits/bd6842285a412cd3266a730a6e658f32bbda5cfb) to prevent modification:

```ts
export const getMetricRatingThresholds = (metricName: Metric['name']): MetricRatingThresholds | null => {
  try {
    // Return a copy to prevent changes
    const {good, needsImprovement} = metricRatingThresholds[metricName];
    return {good, needsImprovement};
  } catch (e) { /* ... */ }
};
```

README.md
```diff
@@ -1176,6 +1197,20 @@ interface TTFBAttribution {
 }
 ```
 
+### Rating Thresholds:
```
philipwalton, line 1200: Can you move this section above the "Attribution:" section and below the "Functions:" section? Since this is part of the standard build, I think it makes sense to put it before the attribution stuff and keep that section last.

philipwalton (approved):
LGTM. Thanks for hanging in there with us on this!

README.md
robatron: [Done](https://github.com/GoogleChrome/web-vitals/pull/323/commits/daf6bed4793a46692816165353caab2debc6af2b)! Also [added a note](https://github.com/GoogleChrome/web-vitals/pull/323/commits/daf6bed4793a46692816165353caab2debc6af2b#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R984) encouraging the use of the `rating` supplied by the report callback functions:

> _**Note:** It's typically not necessary (or recommended) to manually calculate metric value ratings using these thresholds. Use the [`Metric['rating']`](#metric) supplied by the [`ReportCallback`](#reportcallback) functions instead._

Wdyt?

philipwalton: Sure, I think that's helpful. Thanks!

tunetheweb:
OK let's merge this. Will do a release then.

PR #348, opened 2023-05-26
Fix attribution types

### LINKED ISSUE #331: attribution typescript error
Using the example from the README.md

```
import {onCLS, onFID, onLCP} from 'web-vitals/attribution';

function sendToGoogleAnalytics({name, delta, value, id, attribution}) {
  const eventParams = {
    // Built-in params:
    value: delta, // Use `delta` so the value can be summed.
    // Custom params:
    metric_id: id, // Needed to aggregate events.
    metric_value: value, // Optional.
    metric_delta: delta, // Optional.
  };

  switch (name) {
    case 'CLS':
      break;
    case 'FID':
      eventParams.debug_target = attribution.eventTarget;
      break;
    case 'LCP':
      eventParams.debug_target = attribution.element;
      break;
  }

  // Assumes the global `gtag()` function exists, see:
  // https://developers.google.com/analytics/devguides/collection/ga4
  gtag('event', name, eventParams);
}

onCLS(sendToGoogleAnalytics);
onFID(sendToGoogleAnalytics);
onLCP(sendToGoogleAnalytics);
```

Generates a typescript error:

```
Argument of type '({ name, delta, value, id, attribution }: { name: any; delta: any; value: any; id: any; attribution: any; }) => void' is not assignable to parameter of type 'CLSReportCallbackWithAttribution'.
  Types of parameters '__0' and 'metric' are incompatible.
    Property 'attribution' is missing in type 'CLSMetric' but required in type '{ name: any; delta: any; value: any; id: any; attribution: any; }'.
```

It seems the function still thinks it's a CLSMetric and not a CLSMetricWithAttribution

tunetheweb:
Fixes #331

### DIFF
--- a/src/types/base.ts
+++ b/src/types/base.ts
@@ -116,6 +116,10 @@ export interface ReportCallback {
   (metric: Metric): void;
 }
 
+export interface ReportCallbackWithAttribution {
+  (metric: MetricWithAttribution): void;
+}
+
 export interface ReportOpts {
   reportAllChanges?: boolean;
   durationThreshold?: number;
--- a/src/types/cls.ts
+++ b/src/types/cls.ts
@@ -14,7 +14,12 @@
  * limitations under the License.
  */
 
-import {LoadState, Metric, ReportCallback} from './base.js';
+import {
+  LoadState,
+  Metric,
+  ReportCallback,
+  ReportCallbackWithAttribution,
+} from './base.js';
 
 /**
  * A CLS-specific version of the Metric object.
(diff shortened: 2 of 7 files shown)


src/attribution/onCLS.ts
```diff
@@ -81,7 +81,7 @@ export const onCLS = (
   opts?: ReportOpts
 ) => {
   unattributedOnCLS(
-    ((metric: CLSMetric) => {
+    ((metric: CLSMetricWithAttribution) => {
```
tunetheweb, line 84: This looks like it was missed previously. All the other metrics apart from CLS have this set correctly.

cklim4405: > This looks like it was missed previously. All the other metrics apart from CLS have this set correctly.

PR #353, opened 2023-06-12
Remove mention of polyfill from docs

mxdvl:
## What does this change?

Remove the mention of the polyfill from the _browser support_ section of the README.

## Why is this change needed

This has been removed in the latest version: #242

It’s confusing to see it mentioned in the browser support section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -530,7 +530,7 @@ addEventListener('visibilitychange', () => {
   }
 });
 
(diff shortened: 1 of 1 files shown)


tunetheweb:
The polyfills still exist for now. The #242 just added deprecation notices before we remove them completely (likely in v4)

Issue #350, opened 2023-06-02
v3.3.2 throw error in IE9.


laoboxie:
web-vitals@3.3.2 throw error in IE9.
The error is caused by performance.now()
lib: https://unpkg.com/web-vitals@3.3.2/dist/web-vitals.iife.js

images:
![image](https://github.com/GoogleChrome/web-vitals/assets/23701019/c4d97026-8ad2-4ee3-9fde-2cdfe9601214)
![image](https://github.com/GoogleChrome/web-vitals/assets/23701019/39c73bad-fb04-4f90-a6a9-919eeb02c1c0)

philipwalton:
I can't reproduce this error when testing on IE9 via browserstack.com. @laoboxie can you clarify exactly what code you're running that's causing this error?

I'm trying the following, which doesn't error for me in IE9.

```html
<script>
(function () {
  var script = document.createElement('script');
  script.src = 'https://unpkg.com/web-vitals@3.3.2/dist/web-vitals.iife.js';
  script.onload = function () {
    webVitals.getTTFB(console.log);
  };
  document.head.appendChild(script);
})();
</script>
```

Also, when I look at the code, I do not believe it's possible for the part of the code that calls `performance.now()` to be reached in IE9 because that line is within a conditional that only runs if the browser finds a `navigation` performance entry (which would not be true in IE9). See: https://github.com/GoogleChrome/web-vitals/blob/1d2c191e6763ea3ebeef05e9141847dbce554228/src/onTTFB.ts#L71-L82

So unless I'm missing something then I think the documentation is correct in that the `onTTFB()` function doesn't produce a result in IE9 but it also doesn't error:

> The web-vitals code has been tested and will run without error in all major browsers as well as Internet Explorer back to version 9.

And I think it's fine to continue to make this claim because I do think we want people to be able to run this code anywhere without worry that it will error and potentially break other critical functionality on their site.

tunetheweb:
Closing this as no further updates heard back. Feel free to reopen if necessary.

PR #355, opened 2023-06-12
Remove Internet Explorer from list of supported browsers

tunetheweb:
As noted in #350 web-vitals no longer works in IE9 as of version 2.1.2. It probably still works in IE10 and IE11 but is no longer tested in either and so is at risk of breaking in either soon.

With all versions of IE not officially unsupported I think it's best to remove any language of it's support from the README. Perhaps this should be a breaking change and saved for v4, but as noted v3 (and also the latest version of v3) already are not reflective of the actual situation so I'm tempted not to leave this for v4 and just merge this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1218,16 +1218,16 @@ interface TTFBAttribution {
 
 ## Browser Support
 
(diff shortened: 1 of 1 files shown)


tunetheweb:
Closing for now as not needed.

PR #356, opened 2023-06-13
Metric type fixes

### LINKED ISSUE #351: 3.3.2: type errors using attribution callbacks
Following the fixes for https://github.com/GoogleChrome/web-vitals/issues/331 in https://github.com/GoogleChrome/web-vitals/pull/348, I upgraded to 3.3.2 to see if it fixed the issues I was facing, however I am still having problems. For example:

```ts
import { onCLS } from "web-vitals/attribution";
onCLS((metric) => {
  // Expected type: `string | undefined`
  // Actual type: `unknown`
  metric.attribution.largestShiftTarget;
  //                 ^?
});

```

Suggested fix:

```diff
-export interface CLSReportCallbackWithAttribution extends ReportCallbackWithAttribution {
+export interface CLSReportCallbackWithAttribution {
     (metric: CLSMetricWithAttribution): void;
 }
```

We would need to make a similar change for the other metrics as well.

### LINKED ISSUE #358: why does onCLS return non CLS metrics?
According to the typings `onCLS` can return CLS Metrics or generic Metrics:
webVitalMetricCollector_ts_—_isomorph

Is that a typing bug?

tunetheweb:
Fixes #351 
Fixes #358

It appears [you can't just change a type of an interface in Typescript by repeating it](https://stackoverflow.com/questions/41285211/overriding-interface-property-type-defined-in-typescript-d-ts-file).

This PR uses `Omit` to completely replace the `metric` value, but then has to do an assertion for `ReportCall` of the more specific metric `XXXReportCall` type to the generic `ReportCall` type.

### DIFF
--- a/src/onTTFB.ts
+++ b/src/onTTFB.ts
@@ -18,7 +18,12 @@ import {bindReporter} from './lib/bindReporter.js';
 import {initMetric} from './lib/initMetric.js';
 import {onBFCacheRestore} from './lib/bfcache.js';
 import {getNavigationEntry} from './lib/getNavigationEntry.js';
-import {MetricRatingThresholds, ReportCallback, ReportOpts} from './types.js';
+import {
+  MetricRatingThresholds,
+  ReportCallback,
+  ReportOpts,
+  TTFBReportCallback,
+} from './types.js';
 import {getActivationStart} from './lib/getActivationStart.js';
 import {whenActivated} from './lib/whenActivated.js';
 
@@ -55,13 +60,13 @@ const whenReady = (callback: () => void) => {
  * includes time spent on DNS lookup, connection negotiation, network latency,
  * and server processing time.
  */
-export const onTTFB = (onReport: ReportCallback, opts?: ReportOpts) => {
+export const onTTFB = (onReport: TTFBReportCallback, opts?: ReportOpts) => {
   // Set defaults
   opts = opts || {};
 
   let metric = initMetric('TTFB');
   let report = bindReporter(
-    onReport,
+    onReport as ReportCallback,
     metric,
     TTFBThresholds,
     opts.reportAllChanges
@@ -95,7 +100,7 @@ export const onTTFB = (onReport: ReportCallback, opts?: ReportOpts) => {
       onBFCacheRestore(() => {
         metric = initMetric('TTFB', 0);
         report = bindReporter(
-          onReport,
+          onReport as ReportCallback,
           metric,
           TTFBThresholds,
           opts!.reportAllChanges
--- a/src/types/cls.ts
+++ b/src/types/cls.ts
@@ -81,7 +81,7 @@ export interface CLSMetricWithAttribution extends CLSMetric {
 /**
  * A CLS-specific version of the ReportCallback function.
  */
-export interface CLSReportCallback extends ReportCallback {
+export interface CLSReportCallback extends Omit<ReportCallback, 'metric'> {
   (metric: CLSMetric): void;
 }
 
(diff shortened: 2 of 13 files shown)


jantimon:
fixes #358

src/types/cls.ts
```diff
@@ -81,14 +81,14 @@ export interface CLSMetricWithAttribution extends CLSMetric {
 /**
  * A CLS-specific version of the ReportCallback function.
  */
-export interface CLSReportCallback extends ReportCallback {
+export interface CLSReportCallback extends Omit<ReportCallback, 'metric'> {
```
philipwalton, line 84: Is anything gained by doing it this way rather than just doing this (without the `extends`)?

```suggestion
export interface CLSReportCallback {
```

src/types/cls.ts
```diff
@@ -81,14 +81,14 @@ export interface CLSMetricWithAttribution extends CLSMetric {
...
   (metric: CLSMetric): void;
 }
 
 /**
  * A CLS-specific version of the ReportCallback function with attribution.
  */
 export interface CLSReportCallbackWithAttribution
-  extends ReportCallbackWithAttribution {
+  extends Omit<ReportCallbackWithAttribution, 'metric'> {
```
philipwalton, line 92: Same question here, as well as in all the other report callback variants...

Given that this is just a function and not a type with a number of different properties on it, it's not clear to me that the `Omit<>` is providing any value (e.g. making the code more DRY or anything like that).

src/onTTFB.ts
```diff
@@ -55,13 +60,13 @@ const whenReady = (callback: () => void) => {
...
  */
-export const onTTFB = (onReport: ReportCallback, opts?: ReportOpts) => {
+export const onTTFB = (onReport: TTFBReportCallback, opts?: ReportOpts) => {
   // Set defaults
   opts = opts || {};
 
   let metric = initMetric('TTFB');
   let report = bindReporter(
-    onReport,
+    onReport as ReportCallback,
```
philipwalton, line 69: I think this is probably fine for now since this is internal code...

...but if we're going to be exporting metric-specific versions of all of these types, it seems like ideally we would be able to make the `bindReporter()` function understand those types as well. E.g. something similar to what we do here: https://github.com/GoogleChrome/web-vitals/blob/1d2c191e6763ea3ebeef05e9141847dbce554228/src/lib/observe.ts#L40-L44

src/types/cls.ts
tunetheweb: I swear I tried that before and [couldn't even get it to build](https://github.com/GoogleChrome/web-vitals/issues/351#issuecomment-1588266261). But it works now, so must have been something else wrong! Will remove this as suggested.

src/types/cls.ts
tunetheweb: Ditto

src/onTTFB.ts
tunetheweb: I couldn't figure out a way to do this as the metrics are not the same (particularly attribution has an extra `attribution` attribute). But open to pointers here.

Also, unlike in the `observe` example (where the specific type had to be passed on to the observer), we're treating all the metrics the same in `bindreporter`, so seemed easier to just assert this as a ReportCallback type and not treat them as the 12 individual types (6 metrics, and 6 metrics with attribution).

philipwalton (approved):

src/onTTFB.ts
brendankenny: Yeah, it's the mismatch between the specific callbacks for the `on*` methods (e.g. `TTFBReportCallback` for `onTTFB()`), and the fact that `bindReporter()` currently requires _its_ `onReport` to be able to handle _any_ of the metrics (which e.g. `TTFBReportCallback` purposefully doesn't) that you get the type error and the need for the type assertion.

I'm happy to take a stab at parameterizing `bindReporter` after this PR lands. Because `bindReporter` truly doesn't care about the handler internally, it should be relatively disruption-free type rearranging (and I believe should streamline some metric type stuff at the same time).

brendankenny: PR in #359

PR #359, opened 2023-06-16
make bindReporter generic over metric type

brendankenny:
Can land after #356

Parameterizes the `bindReporter` type on the `metric` passed in so that the `callback` can now be typed specifically for that metric (e.g. a result for `CLSMetric` must be using a `CLSReportCallback`).

This required `initMetric` to be parameterized similarly, so the returned metric type is now based on the metric `name` passed in.

```ts
let metric = initMetric('FCP');
//  ^ FCPMetric
```

---

The main change required was using a union of the metric types wherever the base `Metric` interface is currently used. This leaves most uses of it unchanged, but allows narrowing to which specific metric is being used based on the `name` property.

I renamed the existing `Metric` to be `MetricBase` and named the union `Metric` to keep the uses of it unchanged, but if someone has better naming suggestions, changing is no problem :)

### DIFF
--- a/src/lib/initMetric.ts
+++ b/src/lib/initMetric.ts
@@ -20,7 +20,10 @@ import {getActivationStart} from './getActivationStart.js';
 import {getNavigationEntry} from './getNavigationEntry.js';
 import {Metric} from '../types.js';
 
-export const initMetric = (name: Metric['name'], value?: number): Metric => {
+export const initMetric = <MetricName extends Metric['name']>(
+  name: MetricName,
+  value?: number
+) => {
   const navEntry = getNavigationEntry();
   let navigationType: Metric['navigationType'] = 'navigate';
 
@@ -39,12 +42,15 @@ export const initMetric = (name: Metric['name'], value?: number): Metric => {
     }
   }
 
+  // Use `entries` type specific for the metric.
+  const entries: Extract<Metric, {name: MetricName}>['entries'] = [];
+
   return {
     name,
     value: typeof value === 'undefined' ? -1 : value,
-    rating: 'good', // Will be updated if the value changes.
+    rating: 'good' as const, // If needed, will be updated when reported. `const` to keep the type from widening to `string`.
     delta: 0,
-    entries: [],
+    entries,
     id: generateUniqueID(),
     navigationType,
   };
--- a/src/lib/initMetric.ts
+++ b/src/lib/initMetric.ts
@@ -18,14 +18,14 @@ import {getBFCacheRestoreTime} from './bfcache.js';
 import {generateUniqueID} from './generateUniqueID.js';
 import {getActivationStart} from './getActivationStart.js';
 import {getNavigationEntry} from './getNavigationEntry.js';
-import {Metric} from '../types.js';
+import {MetricType} from '../types.js';
 
-export const initMetric = <MetricName extends Metric['name']>(
+export const initMetric = <MetricName extends MetricType['name']>(
   name: MetricName,
   value?: number
 ) => {
   const navEntry = getNavigationEntry();
-  let navigationType: Metric['navigationType'] = 'navigate';
+  let navigationType: MetricType['navigationType'] = 'navigate';
 
   if (getBFCacheRestoreTime() >= 0) {
     navigationType = 'back-forward-cache';
@@ -38,12 +38,12 @@ export const initMetric = <MetricName extends Metric['name']>(
       navigationType = navEntry.type.replace(
         /_/g,
         '-'
-      ) as Metric['navigationType'];
+      ) as MetricType['navigationType'];
     }
   }
 
   // Use `entries` type specific for the metric.
-  const entries: Extract<Metric, {name: MetricName}>['entries'] = [];
+  const entries: Extract<MetricType, {name: MetricName}>['entries'] = [];
 
   return {
     name,
--- a/src/lib/polyfills/interactionCountPolyfill.ts
+++ b/src/lib/polyfills/interactionCountPolyfill.ts
@@ -15,7 +15,6 @@
  */
 
 import {observe} from '../observe.js';
-import {Metric} from '../../types.js';
 
 declare global {
   interface Performance {
@@ -27,8 +26,8 @@ let interactionCountEstimate = 0;
 let minKnownInteractionId = Infinity;
 let maxKnownInteractionId = 0;
 
-const updateEstimate = (entries: Metric['entries']) => {
-  (entries as PerformanceEventTiming[]).forEach((e) => {
+const updateEstimate = (entries: PerformanceEventTiming[]) => {
+  entries.forEach((e) => {
     if (e.interactionId) {
       minKnownInteractionId = Math.min(minKnownInteractionId, e.interactionId);
       maxKnownInteractionId = Math.max(maxKnownInteractionId, e.interactionId);
--- a/src/onCLS.ts
+++ b/src/onCLS.ts
@@ -26,7 +26,6 @@ import {
   CLSMetric,
   CLSReportCallback,
   MetricRatingThresholds,
-  ReportCallback,
   ReportOpts,
 } from './types.js';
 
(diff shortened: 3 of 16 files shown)


src/lib/polyfills/interactionCountPolyfill.ts
```diff
@@ -27,8 +26,8 @@ let interactionCountEstimate = 0;
 let minKnownInteractionId = Infinity;
 let maxKnownInteractionId = 0;
 
-const updateEstimate = (entries: Metric['entries']) => {
-  (entries as PerformanceEventTiming[]).forEach((e) => {
+const updateEstimate = (entries: PerformanceEventTiming[]) => {
```
brendankenny, line 29: `observe()` checks the callback type, might as well trust it :)

src/onCLS.ts
```diff
@@ -66,9 +65,8 @@ export const onCLS = (onReport: CLSReportCallback, opts?: ReportOpts) => {
       let report: ReturnType<typeof bindReporter>;
 
       let sessionValue = 0;
-      let sessionEntries: PerformanceEntry[] = [];
+      let sessionEntries: LayoutShift[] = [];
```
brendankenny, line 68: minor type bug that's an explicit error now with these changes (on `sessionEntries.push(entry)` below), but no functional change

src/types/base.ts
```diff
@@ -18,8 +18,14 @@ import {
   FirstInputPolyfillEntry,
   NavigationTimingPolyfillEntry,
 } from './polyfills.js';
+import type {CLSMetric} from './cls.js';
+import type {FCPMetric} from './fcp.js';
+import type {FIDMetric} from './fid.js';
+import type {INPMetric} from './inp.js';
+import type {LCPMetric} from './lcp.js';
+import type {TTFBMetric} from './ttfb.js';
```
brendankenny, line 26: `import type` vs `import` here doesn't change anything in the output since these files (currently) have only types in them, but generally good to be explicit that the circular dependency is purely in type-space and so won't ever be an issue

src/types/base.ts
```diff
@@ -85,17 +91,14 @@ export interface Metric {
     | 'restore';
 }
 
-/**
- * A version of the `Metric` that is used with the attribution build.
- */
-export interface MetricWithAttribution extends Metric {
```
brendankenny, line 91: this isn't used or referred to from anywhere—metrics with attribution extend their individual metrics directly (e.g. `LCPMetricWithAttribution extends LCPMetric`)—so I removed, but I can restore if it's still useful for something

src/lib/initMetric.ts
```diff
@@ -39,12 +42,15 @@ export const initMetric = (name: Metric['name'], value?: number): Metric => {
...
 
+  // Use `entries` type specific for the metric.
+  const entries: Extract<Metric, {name: MetricName}>['entries'] = [];
+
   return {
     name,
     value: typeof value === 'undefined' ? -1 : value,
-    rating: 'good', // Will be updated if the value changes.
+    rating: 'good' as const, // Will be updated if the value changes.
     delta: 0,
```
tunetheweb, line 52: Why do we need this `as const`? And how can it be a `const` if the value changes?

tunetheweb (commented):
LGTM, but with one question.

src/lib/initMetric.ts
brendankenny: > Why do we need this `as const`? And how can it be a `const` if the value changes?

It's kind of a misnomer in this case. Practically speaking, it's telling the compiler not to widen the type of `rating` to `string` and keep it as the literal `'good'` (since `rating` is typed to `'good' | 'needs-improvement' | 'poor'`, the type has to be one (or more) of those three, and not the more general `string`).

One way of thinking about it is that the _value_ `'good'` is being treated as `const`, but assigned to the property `rating`, which is not `const` and can be changed freely after this.

The other common approach besides `as const` would be `rating: 'good' as 'good'`, but
- that's slightly less idiomatic these days (`as const` was added to remove awkward self assertions, though usually uglier ones than this one)
- it's currently disallowed by the `prefer-as-const` lint rule.

src/types/base.ts
philipwalton: I'm not exactly sure if I did this on purpose and in this case, but in many cases I created a type that I could just copy and paste to the README for the docs. E.g. https://github.com/GoogleChrome/web-vitals/#metricwithattribution

Not the most sophisticated doc generation :) but the point was that the docs would be easy to create from the types files.

In this case if it's truly not being used then I guess there's no additional "checking" that's happening by keeping it here, so it's probably fine to remove. But then I guess the question is what should we put in the docs?

src/types/base.ts
```diff
@@ -85,17 +91,14 @@ export interface Metric {
...
-  /**
-   * An object containing potentially-helpful debugging information that
-   * can be sent along with the metric value for the current page visit in
-   * order to help identify issues happening to real-users in the field.
-   */
-  attribution: {[key: string]: unknown};
-}
+/** The union of supported metric types. */
+export type Metric =
```
philipwalton, line 95: I don't think we can do this in v3 because I think it would technically be breaking, but given that we're now defining types of each metric, I wonder if we even need the `Metric` type at all? WDYT?

philipwalton (approved):

src/types/base.ts
brendankenny: The union is needed in enough places (`bindReporter`, `initMetric`, and `ReportCallback` in this file) that it's convenient to have a name for it, otherwise each of those files would have to define their own and would all have to be updated if/when the next metric is added.

I didn't think of it as a breaking change because
- the only places the old `Metric` was used were internal (again `bindReporter`/`initMetric` and the generic `ReportCallback` type) and all the user-exposed functions return specific metrics (which are unchanged)
- The (new) `Metric` union is assignable to `MetricBase` (the old `Metric`), so even if users have fished out the old `Metric` type for use, in many cases it should be usable unchanged.

But there's no reason we have to repurpose the name (like `MetricWithAttribution`, `Metric` is also documented in the readme). I could switch `MetricBase` back to `Metric` and `Metric` could be...`MetricType`? Blanking on a good name, but open to suggestions :)

src/types/base.ts
brendankenny: probably easier to revert and come at it if/when the documentation changes :)

src/types/base.ts
philipwalton: > I didn't think of it as a breaking change because

Yeah, I don't think what you've done is a breaking change. But I do think it would be considered a breaking chance if we renamed `Metric` to `MetricType` (for example) because presumably some people importing this package are also importing the `Metric` type. So that's why I was saying maybe we could make that change at the next major version release (along with any other type changes we want to make).

src/types/base.ts
tunetheweb: I added this in #356 so definitely good to remove this so you're cleaning up #356 in this. And, as it's a PR into that branch, we should merge this and then merge #356

brendankenny: #348 added `ReportCallbackWithAttribution`, `MetricWithAttribution` was added in #237

> And, as it's a PR into that branch, we should merge this and then merge #356

Oh, I was only doing that so the diff made sense. Sorry to hold you up. You can merge #356 any time and I'll just change the PR to compare against `main`.

src/types/base.ts
brendankenny: OK, just to be clear because reusing the `Metric` name has only cause me to be confused in this conversation :) I'm suggesting changing this PR so:
- `Metric` stay `Metric` (not `MetricBase`)
- The new type that's a union of the individual metrics be given some other name (e.g. `MetricType`)

Existing imports of `Metric` would be unchanged and the second one is brand new, so wouldn't affect anything, and since it's all internal, there's no reason for anyone to reach in to import it.

So is there still a problem after making that change and what are you suggesting instead?

philipwalton: Sure, that SGTM.

src/types/base.ts
tunetheweb: Whichever's easier for you!

PR #368, opened 2023-07-12
Prevent FID entries being emitted as INP for non-supporting browsers

### LINKED ISSUE #367: Remove INP collection on Firefox
web-vitals.js is currently reporting an INP value on Firefox that does not match the INP definition. I propose to remove Firefox support.

tunetheweb:
Fixes #367

### DIFF
--- a/src/onINP.ts
+++ b/src/onINP.ts
@@ -218,9 +218,12 @@ export const onINP = (onReport: INPReportCallback, opts?: ReportOpts) => {
     );
 
     if (po) {
-      // Also observe entries of type `first-input`. This is useful in cases
+      // If browser supports interactionId (and so supports INP), also
+      // observe entries of type `first-input`. This is useful in cases
       // where the first interaction is less than the `durationThreshold`.
-      po.observe({type: 'first-input', buffered: true});
+      if ('interactionId' in PerformanceEventTiming?.prototype) {
+        po.observe({type: 'first-input', buffered: true});
+      }
 
       onHidden(() => {
         handleEntries(po.takeRecords() as INPMetric['entries']);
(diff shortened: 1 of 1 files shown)


src/onINP.ts
```diff
@@ -218,9 +217,12 @@ export const onINP = (onReport: INPReportCallback, opts?: ReportOpts) => {
     );
 
     if (po) {
-      // Also observe entries of type `first-input`. This is useful in cases
+      // If browser supports interactionId (and so supports INP), also
+      // observe entries of type `first-input`. This is useful in cases
       // where the first interaction is less than the `durationThreshold`.
-      po.observe({type: 'first-input', buffered: true});
+      if ('interactionId' in PerformanceEventTiming?.prototype) {
```
philipwalton, line 223:
```suggestion
      if ('interactionId' in PerformanceEventTiming.prototype) {
```

You don't need the `?` here because `PerformanceEventTiming` is guaranteed to exist if the `PerformanceObserver` object exists (which there is a conditional for).

Also, FYI, I've been avoiding using most ES2015+ features (like optional chaining) in this library because they add transpilation bloat, which increases the size of our bundles in /dist.

philipwalton (changes requested):

src/onINP.ts
tunetheweb: Ah was curious why they weren't use, when quicker than a extra check. I presumed transpilation would effectively just add that extra check for me. I should probably review the ones I started adding recently.

tunetheweb: Actually turns out I've only been using them in my soft navs branch. Will update there.

philipwalton: Yeah, I would have noticed if you were using them in the `main` branch :)

> I presumed transpilation would effectively just add that extra check for me.

We have transpilation steps for the `iife` and `umd` bundles, and we don't have one for the module bundle. The idea is that, if a person is only running this code on modern browsers (via `<script type=module>`), then their users shouldn't have to pay the transpilation cost. BUT, you want to be pretty sure that the feature is not going to break in browsers that support modules but not the specific feature being used.

Though it looks like this feature has been stable in modern browsers for at least the past 2 years, so perhaps it's time to reconsider (in a separate PR though): https://caniuse.com/mdn-javascript_operators_optional_chaining

philipwalton (approved):

PR #371, opened 2023-07-25
Remove reference to Universal Analytics from README

tunetheweb:
With UA now having been retired in favor of GA4 we should remove these references from the README.

Additionally remove references to [Web Vitals Report](https://github.com/GoogleChromeLabs/web-vitals-report) since it does not support GA4.

### DIFF
--- a/README.md
+++ b/README.md
@@ -326,7 +326,7 @@ onLCP(sendToAnalytics);
 
 ### Send the results to Google Analytics
 
-For details of have to query this data in [BigQuery](https://cloud.google.com/bigquery), or visualise it in [Looker Studio](https://lookerstudio.google.com/), see the [Measure and debug performance with Google Analytics 4 and BigQuery](https://web.dev/vitals-ga4/) post.
+Google Analytics does not support reporting metric distributions in any of its built-in reports; however, if you set a unique event parameter value (in this case, the metric_id, as shown in the example below) on every metric instance that you send to Google Analytics, you can create a report yourself by first getting the data via the [Google Analytics Data API](https://developers.google.com/analytics/devguides/reporting/data/v1) or via [BigQuery export](https://support.google.com/analytics/answer/9358801) and then visualizing it any charting library you choose.
 
 [Google Analytics 4](https://support.google.com/analytics/answer/10089681) introduces a new Event model allowing custom parameters instead of a fixed category, action, and label. It also supports non-integer values, making it easier to measure Web Vitals metrics compared to previous versions.
 
@@ -356,6 +356,8 @@ onCLS(sendToGoogleAnalytics);
 onFID(sendToGoogleAnalytics);
 onLCP(sendToGoogleAnalytics);
 ```
+
+For details on how to query this data in [BigQuery](https://cloud.google.com/bigquery), or visualise it in [Looker Studio](https://lookerstudio.google.com/), see [Measure and debug performance with Google Analytics 4 and BigQuery](https://web.dev/vitals-ga4/).
 
 ### Send the results to Google Tag Manager
 
--- a/README.md
+++ b/README.md
@@ -326,86 +326,9 @@ onLCP(sendToAnalytics);
 
 ### Send the results to Google Analytics
 
-Google Analytics does not support reporting metric distributions in any of its built-in reports; however, if you set a unique dimension value (in this case, the metric `id`, as shown in the examples below) on every metric instance that you send to Google Analytics, you can create a report yourself using the [Google Analytics Reporting API](https://developers.google.com/analytics/devguides/reporting) and any data visualization library you choose.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -326,86 +326,9 @@ onLCP(sendToAnalytics);
 
 ### Send the results to Google Analytics
 
-Google Analytics does not support reporting metric distributions in any of its built-in reports; however, if you set a unique dimension value (in this case, the metric `id`, as shown in the examples below) on every metric instance that you send to Google Analytics, you can create a report yourself using the [Google Analytics Reporting API](https://developers.google.com/analytics/devguides/reporting) and any data visualization library you choose.
+Google Analytics does not support reporting metric distributions in any of its built-in reports; however, if you set a unique dimension value (in this case, the `metric_id`, as shown in the examples below) on every metric instance that you send to Google Analytics, you can create a report yourself using the [Google Analytics Reporting API](https://developers.google.com/analytics/devguides/reporting) and any data visualization library you choose.
```
philipwalton, line 329: The Reporting API is for UA only. The Data API is the API for GA4, which is still in alpha/beta: https://developers.google.com/analytics/devguides/reporting/data/v1

Also it's  not clear to me whether "unique dimension value" is the right terminology given that this is going to be GA4 only; it might be "unique event parameter. However, I'm not 100% sure how you get those values via the API. It might still require creating a custom dimension in the UI, which is then defined as a custom event parameter value (again, not 100% sure).

Alternatively, we could just not recommend the API at all and stick to the recommendation of using BigQuery...

README.md
```diff
@@ -326,86 +326,9 @@ onLCP(sendToAnalytics);
...
-  });
-}
-
-onCLS(sendToGoogleAnalytics);
-onFID(sendToGoogleAnalytics);
-onLCP(sendToGoogleAnalytics);
-```
+For details of have to query this data in [BigQuery](https://cloud.google.com/bigquery), or visualise it in [Looker Studio](https://lookerstudio.google.com/), see the [Measure and debug performance with Google Analytics 4 and BigQuery](https://web.dev/vitals-ga4/) post.
 
 #### Using `gtag.js` (Google Analytics 4)
```
philipwalton, line 333: This heading can probably be removed as well, since there's now only going to be one option.

README.md
tunetheweb: OK let's just remove this section.

README.md

README.md
```diff
@@ -326,88 +326,7 @@ onLCP(sendToAnalytics);
 
 ### Send the results to Google Analytics
 
-Google Analytics does not support reporting metric distributions in any of its built-in reports; however, if you set a unique dimension value (in this case, the metric `id`, as shown in the examples below) on every metric instance that you send to Google Analytics, you can create a report yourself using the [Google Analytics Reporting API](https://developers.google.com/analytics/devguides/reporting) and any data visualization library you choose.
```
philipwalton, line 329: Overall I think that intro content is still important to keep, especially the bit about GA not supporting distributions (and thus no p75). I just want to make sure we're using the correct terminology. I'm 99% sure the Data API can do this, what I'm not clear on is whether or not the API can query the event param directly or if you need to register a custom dimension in the UI first.

Note: I'm not suggesting we figure this all out in this PR, but I also want to make sure we're not saying anything incorrect. What about something like:

> Google Analytics does not support reporting metric distributions in any of its built-in reports; however, if you set a unique event parameter value (in this case, the `metric_id`, as shown in the example below) on every metric instance that you send to Google Analytics, you can create a report yourself by first getting the data via the [Google Analytics Data API](https://developers.google.com/analytics/devguides/reporting/data/v1) or via [BigQuery export](https://support.google.com/analytics/answer/9358801) and then visualizing it any charting library you choose.

README.md
```diff
@@ -326,88 +326,7 @@ onLCP(sendToAnalytics);
...
-  });
-}
-
-onCLS(sendToGoogleAnalytics);
-onFID(sendToGoogleAnalytics);
-onLCP(sendToGoogleAnalytics);
-```
-
-#### Using `gtag.js` (Google Analytics 4)
+For details of have to query this data in [BigQuery](https://cloud.google.com/bigquery), or visualise it in [Looker Studio](https://lookerstudio.google.com/), see the [Measure and debug performance with Google Analytics 4 and BigQuery](https://web.dev/vitals-ga4/) post.
```
philipwalton, line 329:
```suggestion
For details on how to query this data in [BigQuery](https://cloud.google.com/bigquery), or visualize it in [Looker Studio](https://lookerstudio.google.com/), see [Measure and debug performance with Google Analytics 4 and BigQuery](https://web.dev/vitals-ga4/).
```

Also, I'd put this content after the code example.

philipwalton (changes requested):

README.md
tunetheweb: LGTM!

README.md

philipwalton (approved):

PR #372, opened 2023-07-25
Fix INP `durationThreshold` bug when set to 0

nicholasray:
When INP's `durationThreshold` was set to a value of `0`, it would actually ignore all values under 40 ms because the logical OR (||) operator was used and `0` is falsey.

Use nullish coalescing operator instead to prevent this from happening.


tunetheweb (approved):

tunetheweb:
Thanks! And apologies for the delay in getting round to this one. Have another fix to add and will then look to push out a release.

PR #414, opened 2024-01-11
Better attribution type

### LINKED ISSUE #408: Type 'XXXMetricWithAttribution' is not assignable to type 'MetricWithAttribution'
When trying to be more generic and use `MetricWithAttribution` with `FIDMetricWithAttribution` (or other `XXXMetricWithAttribution`) I get this Typescript error:

```ts
Type 'FIDMetricWithAttribution' is not assignable to type 'MetricWithAttribution'.
  Types of property 'attribution' are incompatible.
    Type 'FIDAttribution' is not assignable to type '{ [key: string]: unknown; }'.
      Index signature for type 'string' is missing in type 'FIDAttribution'.
```

How to reproduce:
```ts
import { FIDMetricWithAttribution, MetricWithAttribution } from 'web-vitals/attribution';

const fidMetric: FIDMetricWithAttribution = {...};
const metricWithAttribution: MetricWithAttribution = fidMetric;
```

I fixed this by creating a "new" `MetricWithAttribution` with the correct types:
```ts
import {
  CLSAttribution,
  FCPAttribution,
  FIDAttribution,
  FIDMetricWithAttribution,
  INPAttribution,
  LCPAttribution,
  Metric,
  TTFBAttribution,
} from 'web-vitals/attribution';

interface MetricWithAttribution extends Metric {
  attribution:
    | FIDAttribution
    | CLSAttribution
    | LCPAttribution
    | TTFBAttribution
    | INPAttribution
    | FCPAttribution;
}

const fidMetric: FIDMetricWithAttribution = {};
const metricWithAttribution: MetricWithAttribution = fidMetric;
```

The existing [MetricWithAttribution](https://github.com/GoogleChrome/web-vitals/blob/main/src/types/base.ts#L106) definition should be updated or fixed to accommodate these use cases.

tunetheweb:
Fixes #408

### DIFF
--- a/README.md
+++ b/README.md
@@ -672,7 +672,13 @@ interface MetricWithAttribution extends Metric {
    * can be sent along with the metric value for the current page visit in
    * order to help identify issues happening to real-users in the field.
    */
-  attribution: {[key: string]: unknown};
+  attribution:
+    | CLSAttribution
+    | FCPAttribution
+    | FIDAttribution
+    | INPAttribution
+    | LCPAttribution
+    | TTFBAttribution;
 }
 ```
 
--- a/src/types/base.ts
+++ b/src/types/base.ts
@@ -18,12 +18,12 @@ import {
   FirstInputPolyfillEntry,
   NavigationTimingPolyfillEntry,
 } from './polyfills.js';
-import type {CLSMetric} from './cls.js';
-import type {FCPMetric} from './fcp.js';
-import type {FIDMetric} from './fid.js';
-import type {INPMetric} from './inp.js';
-import type {LCPMetric} from './lcp.js';
-import type {TTFBMetric} from './ttfb.js';
+import type {CLSMetric, CLSAttribution} from './cls.js';
+import type {FCPMetric, FCPAttribution} from './fcp.js';
+import type {FIDMetric, FIDAttribution} from './fid.js';
+import type {INPMetric, INPAttribution} from './inp.js';
+import type {LCPMetric, LCPAttribution} from './lcp.js';
+import type {TTFBMetric, TTFBAttribution} from './ttfb.js';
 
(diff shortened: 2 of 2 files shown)


brendankenny (approved):
This works for the issue in #408, but the better solution would be some variation on

```
export type MetricWithAttribution =
  | CLSMetricWithAttribution
  | FCPMetricWithAttribution
  | FIDMetricWithAttribution
  | INPMetricWithAttribution
  | LCPMetricWithAttribution
  | TTFBMetricWithAttribution;
```

I think we were considering that with the `Metric` interface as well, but settled in doing that in a separate `MetricType`. I don't totally remember why, though. Documentation? Or to avoid a breaking change?

If it's a top-level union, then it both fixes the type widening in #408, but also lets you narrow without an assertion, e.g. if you have a generic `metricWithAttribution` you can check the `name` so the compiler knows which metric it's dealing with

```
if (metricWithAttribution.name === 'LCP') {
  const lcpEntry = metricWithAttribution.attribution.lcpEntry; // Property known to be present.
  // ...
}
```

nbd if this isn't a path we want to go down today, but both `Metric` and `MetricWithAttribution` feel like they're leaving typescript functionality behind and making them slightly harder to use than they need to be.

tunetheweb:
Anyone any more thoughts on this?

@brendankenny @mcanu @philipwalton

philipwalton:
@brendankenny can you take a look at this and make a recommendation? I think any change here might be a breaking change, so it'd be good to get this in the v4 release if possible.

philipwalton:
Also, once we have a decision we can update this PR's base branch to the [v4 branch](https://github.com/GoogleChrome/web-vitals/tree/v4).

tunetheweb:
CLosing in favour of #471

PR #421, opened 2024-01-23
Pick the first non-null target for INP attribution

rviscomi:
Some pointer entries have a known [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1367329) causing them to have null event targets. In cases where there are mouse or click entries in the same interaction that do have non-null event targets, always prefer those.

This PR works around the bug by changing the `eventTarget` in INP attribution to use the first entry in the list having a non-null target, rather than always using the longest entry in the list, which may sometimes be a pointer entry.

### DIFF
--- a/src/attribution/onINP.ts
+++ b/src/attribution/onINP.ts
@@ -37,8 +37,10 @@ const attributeINP = (metric: INPMetric): void => {
       );
     })[0];
 
+    const targetEntry = metric.entries.find(entry => entry.target);
+
     (metric as INPMetricWithAttribution).attribution = {
-      eventTarget: getSelector(longestEntry.target),
+      eventTarget: getSelector(targetEntry?.target),
       eventType: longestEntry.name,
       eventTime: longestEntry.startTime,
       eventEntry: longestEntry,


src/attribution/onINP.ts
```diff
@@ -37,8 +37,10 @@ const attributeINP = (metric: INPMetric): void => {
       );
     })[0];
 
+    const targetEntry = metric.entries.find(entry => entry.target);
```
tunetheweb, line 40:
```suggestion
    // TODO: remove when 1367329 is resolved
    // https://bugs.chromium.org/p/chromium/issues/detail?id=1367329
    const targetEntry = metric.entries.find(entry => entry.target);
```

src/attribution/onINP.ts
```diff
@@ -37,8 +37,12 @@ const attributeINP = (metric: INPMetric): void => {
       );
     })[0];
 
+    // TODO: remove when 1367329 is resolved
+    // https://bugs.chromium.org/p/chromium/issues/detail?id=1367329
+    const targetEntry = metric.entries.find(entry => entry.target);
```
tunetheweb, line 42: Actually let's be a bit more explicit as to why we're doing this, as this is a reference library.

```suggestion
    // Currently Chrome can return a null target for certain event types
    // (especially pointer events). As the event target should be the same
    // for all events in the same interaction, we pick the first non-null one.
    // TODO: remove when 1367329 is resolved
    // https://bugs.chromium.org/p/chromium/issues/detail?id=1367329
    const firstTargetEntry = metric.entries.find(entry => entry.target);
```

src/attribution/onINP.ts
```diff
@@ -37,8 +37,12 @@ const attributeINP = (metric: INPMetric): void => {
...
+
     (metric as INPMetricWithAttribution).attribution = {
-      eventTarget: getSelector(longestEntry.target),
+      eventTarget: getSelector(targetEntry?.target),
```
tunetheweb, line 45:
```suggestion
      eventTarget: getSelector(firstTargetEntry?.target),
```

tunetheweb (changes requested):

src/attribution/onINP.ts
```diff
@@ -37,8 +37,12 @@ const attributeINP = (metric: INPMetric): void => {
       );
     })[0];
 
+    // TODO: remove when 1367329 is resolved
+    // https://bugs.chromium.org/p/chromium/issues/detail?id=1367329
```
brendankenny, line 41: FWIW this should be easy to test based on the repro in the crbug (`eventEntry.target` is null but `eventTarget` is non-null), which would then start failing when it was fixed in Chrome and this workaround would be ready to be removed.

tunetheweb: Are you offering to add a test case to this PR? Am I reading this right? 😁

However, even if fixed, we'd likely want to keep this around for a bit until that version of Chrome becomes widely used so not sure the test passing would be a good sign to remove this workaround.

tunetheweb:
@mmocny can you review?

mmocny (approved):
Seems like a strict improvement, thanks Rick.

mmocny:
This PR is similar to the [strategy I've been advertising](https://gist.github.com/mmocny/4e2c9d1905059e23cda79cf97fec6ad6#file-loginteractions-js-L26) for measuring interaction targets.

--

There is also a [different strategy](https://gist.github.com/mmocny/5914f883142429ab7df9c0197fe131ed) that works around the other problem with event targets... that they can be removed and GC-ed by the time Event Timing is reported.

That gist is very rough, but gives you the idea: manage a map of event type + timestamp to whatever metadata you want, saved during event dispatch, then look it up from Event Timing.

src/attribution/onINP.ts
```diff
@@ -37,8 +37,15 @@ const attributeINP = (metric: INPMetric): void => {
       );
     })[0];
 
+    // Currently Chrome can return a null target for certain event types
+    // (especially pointer events). As the event target should be the same
+    // for all events in the same interaction, we pick the first non-null one.
+    // TODO: remove when 1367329 is resolved
+    // https://bugs.chromium.org/p/chromium/issues/detail?id=1367329
+    const firstTargetEntry = metric.entries.find((entry) => entry.target);
```
philipwalton, line 45:
```suggestion
    const firstEntryWithTarget = metric.entries.find((entry) => entry.target);
```

src/attribution/onINP.ts
```diff
@@ -37,8 +37,15 @@ const attributeINP = (metric: INPMetric): void => {
...
+
     (metric as INPMetricWithAttribution).attribution = {
-      eventTarget: getSelector(longestEntry.target),
+      eventTarget: getSelector(firstTargetEntry?.target),
```
philipwalton, line 48:
```suggestion
      eventTarget: getSelector(firstEntryWithTarget && firstEntryWithTarget.target),
```

src/attribution/onINP.ts
philipwalton: I think this phrasing is slightly more clear.

src/attribution/onINP.ts
philipwalton: Note that we don't currently use optional chaining anywhere in this repo, and I'm not sure we should introduce it outside of a major version bump (since it could potentially introduce errors for some users where they weren't seeing errors before).

philipwalton (changes requested):

tunetheweb (approved):

src/attribution/onINP.ts
philipwalton: @rviscomi were you able to reliably reproduce a null target in your testing? If so maybe we can just add a test case that would fail with the current logic and then pass with the new logic (and also start failing again if we remove the fix too early).

rviscomi: I think @tunetheweb has a repro here: https://www.tunetheweb.com/experiments/button/

tunetheweb: That repo doesn't necessarily always report the event with the missing target as the INP event. But it does show all the other events in the interaction. So good to repo the issue of missing targers, but not necessarily the issue or missing targets in INP events.

brendankenny: I can take a stab. Seems straightforward

rviscomi: Thanks @brendankenny!

@philipwalton @tunetheweb can we merge this PR as-is or do you want to couple it with the test case?

philipwalton:
I added a test to cover the logic Rick added. It fails when run against `main` but passed when run on this branch.

philipwalton (approved):

PR #422, opened 2024-01-23
add test for null event target workaround

brendankenny:
I was able to reproduce the issue in https://crbug.com/1367329 by having _no_ pointerdown or pointerup listeners. The click event would then have a `target`, but pointerdown/pointerup would not.

Unfortunately I got stuck (see TODO) on making it so that the pointerdown or pointerup events were longer than the click event so that they'd get sorted first and define the attribution information. I'm curious about any ideas others have to arrange this.

Also happy to take opinions on making the e2e test page simpler instead of continuing to grow complications :)

### DIFF
--- a/test/e2e/onINP-test.js
+++ b/test/e2e/onINP-test.js
@@ -525,6 +525,33 @@ describe('onINP()', async function () {
       const [inp2] = await getBeacons();
       assert.equal(inp2.attribution.loadState, 'loading');
     });
+
+    it.only('works around null attribution.eventTarget bug', async function () {
+      if (!browserSupportsINP) this.skip();
+
+      await browser.url('/test/inp?noPointerEvents&click=100&attribution=1');
+
+      const h1 = await $('h1');
+      await h1.click();
+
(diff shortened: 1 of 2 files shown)


mmocny:
Did you try using mousedown or touchstart etc events?  Those dispatch together with pointerdown and would make the duration of pointer events appear to be long.

Or is that type of handler enough to trigger the effect?

brendankenny:
> Did you try using mousedown or touchstart etc events? Those dispatch together with pointerdown and would make the duration of pointer events appear to be long.

Nope! Right now it's literally just a wait loop in the `click` handler. That sounds like golden info.

@philipwalton is going to take a look when he gets the chance to weigh in on if these extra contortions are worth the test coverage.

philipwalton:
> Did you try using mousedown or touchstart etc events? Those dispatch together with pointerdown and would make the duration of pointer events appear to be long.

Thanks for the suggestion, @mmocny! Using adding a blocking `mousedown` listener worked to repro the bug and create a test for this.

PR #438, opened 2024-02-29
Fix test flakiness in Firefox

### LINKED ISSUE #436: Firefox LCP tests failing
First time it's been run since they added support so need to investigate.

philipwalton:
Fixes #436.

Geckodriver has a bug where, if the webdriver `PageLoadStrategy` option is set to "none" then calls to `browser.url()` can resolve before the browser has actually navigated to the page.

This library needs to set the `PageLoadStrategy` to "none" for many tests, so the only good way to fix this is to wrap the `browser.url()` function (the `navigateTo()` function in thie PR) and manually verify that the correct URL is loaded before resolving. As a convenience this wrapped function also exposes an option to conditionally wait for a given document`readyState` value.

In addition, since Firefox added support for LCP but does NOT support Element Timing, this PR adds a workaround for LCP tests to wait for `load` + `rAF` as a proxy for ensuring all images are loaded and painted.


tunetheweb (approved):
Nice investigation and solutions!

PR #441, opened 2024-03-10
Remove missed polyfill code

philipwalton:
This PR is a continuation of https://github.com/GoogleChrome/web-vitals/pull/435 and removes code missed in that PR.


tunetheweb (approved):
Thanks. I thought some of this was still needed as the polyfill was still used for FCP for bfcache restore. But guess I missed some removal (maybe types.ts) so needed the rest for it to build. Anyway confirmed it builds and passes tests fine without this code so good to merge.

PR #442, opened 2024-03-11
Add INP breakdown timings and LoAF attribution

philipwalton:
This PR updates the INP attribution logic from an "interaction-based" model to a "frame-based" model. This better aligns this library's attribution data with what's exposed in the Interactions track in DevTools, as well as with the Long Animation Frame API.

The following are the most important changes that PR introduces:

**In `/src/onINP.ts`:**

- Before, the `metric.entries` array would only include entries with the same `interactionId` value. Now, `metric.entries` will include *all* `event` entries whose processing time occurs between the user interaction and the next frame painted. This allows developers to more accurately attribute the full "processing time" of the interaction.

**In `/src/attribution/onINP.ts`:**

- Support for the Long Animation Frame API has been added. The new `longAnimationFrameEntry` property will look for a corresponding `long-animation-frame` entry, and report it if found (i.e. if the browser supports the API and the frame was long enough for the entry to be dispatched).
- Breakdown timings for `inputDelay`, `processingTime`, and `presentationDelay` have been added (see the README for detailed description of each).
- All of the `event*` names in the `INPAttribution` interface have been renamed to `interaction*` since now there is no single `event` entry that gets reported (so the term "interaction" now is much more suitable).

To see these changes visualized, below are two screenshots of traces from the INP Demo page showing the new INP attribution data in action. Notice how the "Interaction" and "Event" measures in the Timings track line up perfectly with what DevTools displays in the Interactions track (including the breakdown timing whiskers), as well as with the events in the main thread flame chart below:

Screenshot 2024-03-10 at 9 29 36 PM

Screenshot 2024-03-10 at 9 24 53 PM


README.md
```diff
@@ -877,31 +877,86 @@ interface FIDAttribution {
...
    */
-  eventTime?: number;
+  interactionTime: number;
   /**
-   * The `type` of the `event` dispatched corresponding to INP.
+   * The type of interaction. This will be either 'pointer' or 'keyboard'
+   * since those are the only types of interactions considered for INP.
    */
-  eventType?: string;
+  interactionType: 'pointer' | 'keyboard';
```
philipwalton, line 896: This is to better align with what's shown in DevTools. Also, since now there's no single `eventEntry` that is reported, it doesn't make sense to expose a single event type.

src/attribution/onINP.ts
```diff
@@ -25,38 +26,102 @@ import {
   ReportOpts,
 } from '../types.js';
 
+// The maximum number of LoAF entries with interactions to keep in memory.
+// 10 is chosen here because it corresponds to the maximum number of
+// long interactions needed to compute INP, so no matter which one of those
+// interactions ends up being the INP candidate, it should correspond to one
+// of the 10 longest LoAF entries containing an interaction.
+const MAX_LOAFS_TO_CONSIDER = 10;
```
philipwalton, line 34: FYI: rather than try to match each LoAF up with with pending INP candidate, I thought it would be simpler to just always store the longest 10 LoAFs with an interaction, as that should by a superset of the frames represented by the 10 INP candidate frames (since sometimes more than one interaction can be in the same frame);.

src/attribution/onINP.ts
```diff
@@ -25,38 +26,102 @@ import {
...
+// longest one is first. The list is at most MAX_LOAFS_TO_CONSIDER long.
+const longestLoAFsList: PerformanceLongAnimationFrameTiming[] = [];
+
+// A PerformanceObserver, observing new `long-animation-frame` entries.
+// If this variable is defined it means the browser supports LoAF.
+let loafObserver: PerformanceObserver | undefined;
+
+const handleEntries = (entries: PerformanceLongAnimationFrameTiming[]) => {
+  entries.forEach((entry) => {
+    if (entry.firstUIEventTimestamp > 0) {
```
philipwalton, line 46: I want to confirm that this is indeed the right way to limit LoAF collection to just those that contain a potential INP candidate.

src/attribution/onINP.ts
```diff
@@ -25,38 +26,102 @@ import {
...
-          (a.processingEnd - a.processingStart)
-      );
-    })[0];
-
-    // Currently Chrome can return a null target for certain event types
-    // (especially pointer events). As the event target should be the same
-    // for all events in the same interaction, we pick the first non-null one.
-    // TODO: remove when 1367329 is resolved
-    // https://bugs.chromium.org/p/chromium/issues/detail?id=1367329
-    const firstEntryWithTarget = metric.entries.find((entry) => entry.target);
```
philipwalton, line 45: FYI, now that we store all entries in the same frame (rather than just those with the same `interactionId`, this logic is required to get the proper target (i.e. we need to keep logic this even after this bug is fixed).

src/attribution/onINP.ts
```diff
@@ -25,38 +26,109 @@ import {
...
-  (metric as INPMetricWithAttribution).attribution = {};
+
+  // If the browser supports the Long Animation Frame API and a
+  // `long-animation-frame` entry was found matching this interaction,
+  // use that entry's `renderTime` since it could be more accurate (and
+  // it accounts for non-event listener work such as timers or other
+  // pending tasks that could get run before rendering starts). If not,
+  // use the `processingEnd` value of the last entry in the list, which
+  // should match the `renderTime` value in most cases.
+  const renderStart = longAnimationFrameEntry
```
philipwalton, line 103: @mmocny I want to confirm that you're OK with using either `processingEnd` or LoAF's `renderStart` here to mark the end of `processingTime` / beginning of `presentationDelay`.

src/onINP.ts
```diff
@@ -57,20 +63,88 @@ const getInteractionCountForNavigation = () => {
...
+
+// A mapping between a particular frame's render time and all of the
+// event timing entries that occurred within that frame. Since event timing
+// entries round their duration to the nearest 8sm, this ends up being
+// a best-effort guess, but but debugging and attribution practices it's
+// useful to see all events the occurred within the same frame as a particular
+// interaction, so it's worth doing the best-effort guess.
+// const pendingEntriesGroupMap: Map<number, PendingEntriesGroup> = new Map();
+const pendingEntriesGroupMap: {[renderTime: string]: PendingEntriesGroup} =
+  Object.create(null);
```
philipwalton, line 82: FYI: I'm using `Object.create(null)` here (rather than `{}`) because I need to use a `for...in` loop later and I don't want to have to add an `Object.hasOwnProperty()` check. And the reason I need a `for...in` loop is I want to be able to break out early. And the reason I don't want to use a `for...of` loop is because when that gets transpiled to ES5 it adds a bunch of Babel helpers, so this is the easiest solution...

src/onINP.ts
```diff
@@ -57,20 +63,88 @@ const getInteractionCountForNavigation = () => {
...
+    }
+
+    // If a previous render time is more than 2 seconds before the current
+    // render time, assume all entries within that frame have been fully
+    // processed and remove it from the pending frames map (to free up space).
+    // NOTE: this is not a perfect heuristic, but if it's wrong, the worst
+    // thing that will happen is some entries may not get reported along
+    // with INP (but that is extremely unlikely). It will not impact
+    // the accuracy of the INP value that is reported.
+    if (prevRenderTime + 2000 < renderTime) {
```
philipwalton, line 124: @mmocny this is the "windowing" solution I ended up going with, e.g. keep all entries in memory and drop them once they're more than 2 seconds old.

Let me know if you think 2 seconds is too long (or too short), or if you have a better suggestion to handle out-of-order dispatch.

src/onINP.ts
```diff
@@ -162,31 +245,21 @@ export const onINP = (onReport: INPReportCallback, opts?: ReportOpts) => {
...
         }
 
-        // Entries of type `first-input` don't currently have an `interactionId`,
-        // so to consider them in INP we have to first check that an existing
-        // entry doesn't match the `duration` and `startTime`.
-        // Note that this logic assumes that `event` entries are dispatched
-        // before `first-input` entries. This is true in Chrome (the only browser
-        // that currently supports INP).
-        // TODO(philipwalton): remove once crbug.com/1325826 is fixed.
-        if (entry.entryType === 'first-input') {
```
philipwalton, line 176: FYI: now that some `first-input` entries have an `interactionId` property, this logic no longer works (oops!).

The problem is since line 165 above processes all entries if they have an `interactionId`, this logic here will no longer prevent duplicate `event` and `first-input` entries from getting included in the `metric.entries` array.

Anyway, I figured out that a better solution is to just ignore `first-input` entries if *any* event timing entries have already been processed, since that means the first input was long enough to get picked up by the Event Timing API.

test/server.js
```diff
@@ -26,6 +26,7 @@ nunjucks.configure('./test/views/', {noCache: true});
 // Turn off all caching for tests.
 app.use((req, res, next) => {
   res.set('Cache-Control', 'no-cache');
+  res.set('Access-Control-Allow-Origin', '*');
```
philipwalton, line 29: Unrelated to this PR, but since I was testing my local changes with the INP Demo page, it was useful to be able to load code from the `/dist` folder over localhost (which is generally useful), so I decided to keep this in.

src/attribution/onINP.ts
tunetheweb: We can also have LoAFs without any INP-worthy interaction. Even with filtering on `firstUIEventTimestamp` (see below).

README.md
```diff
@@ -877,31 +877,86 @@ interface FIDAttribution {
 ```ts
 interface INPAttribution {
   /**
-   * A selector identifying the element that the user interacted with for
-   * the event corresponding to INP. This element will be the `target` of the
-   * `event` dispatched.
+   * A selector identifying the element that the user first interacted with
+   * as part of the frame where the INP candidate interaction occurred.
+   * If `interactionTarget` is an empty string, that generally means the
+   * element was removed from the DOM as part of the interaction.
```
tunetheweb, line 883: I'm not sure this is 100% accurate (though maybe "generally" gives us enough leeway).

We've seen instances where the empty string is returned. Plus we're planning on recording this even if it is removed from the DOM (what's the latest on that @mmocny ?)

src/attribution/onINP.ts
```diff
@@ -25,38 +26,109 @@ import {
...
+
+  // The first entry may not have a target defined, so use the first
+  // one found in the entry list.
+  const firstEntryWithTarget = metric.entries.find((entry) => entry.target);
+
+  // Determine the type of interaction based on the first entry with
+  // a matching keydown/keyup or pointerdown/pointerup entry name.
+  const firstEntryWithType = metric.entries.find((entry) =>
+    entry.name.match(/^(key|pointer)(down|up)$/),
+  );
```
tunetheweb, line 115: This won't include `click` events. Can we guarantee that the equivalent `pointerup`/`pointerdown` will always be reported for those? I mean they seem to be but is this guaranteed?

I also see that you had to change your tests to use `simulateUserLikeClick` because of this. I wonder how many other tests we'll break because of this. Although I get it's a major release, I wonder if we should just include `click` events too?

src/types.ts
```diff
@@ -88,4 +88,10 @@ declare global {
     url: string;
     element?: Element;
   }
+
+  // https://w3c.github.io/longtasks/#performancelonganimationframetiming
```
tunetheweb, line 92: This has just moved.

```suggestion
  // https://w3c.github.io/long-animation-frames/#performancelonganimationframetiming
```

src/types/inp.ts
```diff
@@ -31,31 +31,86 @@ export interface INPMetric extends Metric {
  */
 export interface INPAttribution {
   /**
-   * A selector identifying the element that the user interacted with for
-   * the event corresponding to INP. This element will be the `target` of the
-   * `event` dispatched.
+   * A selector identifying the element that the user first interacted with
+   * as part of the frame where the INP candidate interaction occurred.
+   * If `interactionTarget` is an empty string, that generally means the
+   * element was removed from the DOM as part of the interaction.
```
tunetheweb, line 37: Same comment as above. If we change that, we should change this too.

tunetheweb (commented):
In general LGTM. But do have some comments/questions

src/attribution/onINP.ts
noamr: I'd say perhaps the presentation delay should start at `longAnimationFrameEntry.startTime + longAnimationFrameEntry.duration`. That would be the time the main thread offloads to painting/compositing.

README.md
```diff
@@ -877,31 +877,86 @@ interface FIDAttribution {
...
+  interactionType: 'pointer' | 'keyboard';
+
   /**
-   * The `PerformanceEventTiming` entry corresponding to INP.
+   * If the browser supports the Long Animation Frame API and a
+   * `long-animation-frame` entry is detected that corresponds to the INP
+   * frame, it will be reported here.
    */
-  eventEntry?: PerformanceEventTiming;
+  longAnimationFrameEntry?: PerformanceLongAnimationFrameTiming;
```
vanderhoop, line 903: Multiple LoAFs can overlap with the INP window, so does limiting this to just one make sense?

src/onINP.ts
tunetheweb: Should we add some comment to the code to explain this?

And do we need line 80 which is commented out?

src/onINP.ts
```diff
@@ -57,20 +63,88 @@ const getInteractionCountForNavigation = () => {
...
 // A list of longest interactions on the page (by latency) sorted so the
-// longest one is first. The list is as most MAX_INTERACTIONS_TO_CONSIDER long.
+// longest one is first. The list is at most MAX_INTERACTIONS_TO_CONSIDER long.
 let longestInteractionList: Interaction[] = [];
 
 // A mapping of longest interactions by their interaction ID.
 // This is used for faster lookup.
-const longestInteractionMap: {[interactionId: string]: Interaction} = {};
+const longestInteractionMap: {[interactionId: string]: Interaction} =
+  Object.create(null);
```
vanderhoop, line 72: Genuine curiosity: is there a benefit to this change from `{}` (which has less characters)?

src/onINP.ts
vanderhoop: Ahh, I just got tripped up by the `Object.create(null)` further up, as I'd never seen it before, and it caught my attention, so a comment might make sense.

vanderhoop: > FYI: I'm using Object.create(null) here (rather than {}) because I need to use a for...in loop later and I don't want to have to add an Object.hasOwnProperty() check. And the reason I need a for...in loop is I want to be able to break out early. And the reason I don't want to use a for...of loop is because when that gets transpiled to ES5 it adds a bunch of Babel helpers, so this is the easiest solution...

Does a `Map` scratch all the relevant itches?

You'd need an `Array.from` call to use it with `for in` loop, but here's what it could look like:

```es6
const keys = Array.from(map.keys());

for (key in keys) {
  console.log(key)
  break;
}
```

src/attribution/onINP.ts
philipwalton: Yeah, click is a bit of an exception because a click can happen as part of either a `pointer` interaction or a `keyboard` interaction.

I liked the simplicity of just having two options, and I think in 99.99% of all cases there will be a pointer or keyboard event too, but I seem to remember some discussion of cases  (perhaps involving screen readers?) where the will just be a click event. @mmocny do you know?

src/attribution/onINP.ts
philipwalton: Right, the main question is could there ever be cases where one of the 10 INP candidate interactions *doesn't* correspond to one of the 10 longest LoAFs with interactions?

src/onINP.ts
brendankenny: @vanderhoop I think the biggest issue with any alternate approach is IE9 support without pulling in a bunch of Babel code.

@philipwalton `Object.keys()` and then a classic `for (let i = 0; i < keys.length; i++) { const key = keys[i]; ...}` loop could also work, and would avoid surprising readers of the code and the need for a comment. It's also somewhat faster than `for...in` (in microbenchmarks), though either way seems fine.

src/attribution/onINP.ts
philipwalton: > I also see that you had to change your tests to use `simulateUserLikeClick` because of this.

To clarify, the change to the tests actually had nothing to do with this. I made that change because just using the `click()` method in webdriver.io would mean clicks happen instantly (unlike how users actually click), and the `pointerdown`, `mousedown`, `pointerup`, `mouseup`, and `click` events would all happen within the same interaction. Also there were some weird webdriver bugs where other events like `pointermove` and `mouseover` would get dispatched when using the `click()` method. Switching to using the actions API was both more user-like and also avoided the bug, so that's why I did it.

src/attribution/onINP.ts
noamr: This might also discard input delays. Like if the UI event is handled after the LoAF

src/attribution/onINP.ts
noamr: OK, so that would be `renderStart`

tunetheweb: Yup, which is what is being used here (not obvious as the comment was started on the line above):

https://github.com/GoogleChrome/web-vitals/blob/be0b1d38e5689902430523d841604a7d00ee969c/src/attribution/onINP.ts#L103-L105

src/attribution/onINP.ts
tunetheweb: Wouldn't the UI event result in extending the current frame normally in most cases as it's high priority? Or are you talking about if UI event happens during layout/paint on previous frame so has to wait until next frame? In that case the `firstUIEventTimestamp` would be in the previous LoAF so it wouldn't be counted?

I think there's a question of how much we're willing to accept these sorts of outliers versus holding a lot more potential LoAFs in memory? But yeah, these are why I didn't wanna manage a buffer of entries myself and had hoped to use the browser's buffer (before I realised it wasn't a rolling buffer and only first 200 entries).

src/attribution/onINP.ts
mmocny: I left a review comment above this way above in the code comment section.

Re-stating: I think the current proposal is more confusing, not less.  Some non-processing work is labelled processing (idle time waiting for rendering), some processing work is not (rAF), and some rendering work is labelled as both (forced style and layout).

I think the desire the use exactly 3 breakdowns makes it tough, but I if you want to do this and don't want the middle section to be exactly just event-timings, then I would split on:
- Input delay
- "all main thread work" (processing)
- Presentation delay

That has a nice symmetry to it.

That middle section will often just be entirely == all of LoAF, which also makes it easier for sub-attribution of those parts.

Two major disadvantages, I think:
- Relies on LoAF.  The fallback to event-timing processing end is a significant difference.  The "5 breakdowns" approach that treated "processing" as only event processing doesn't have that problem.
- rAF / Style / Layout, if these are lumped together in processing, means we'll have proportionately many more cases where "processing is the problem" to the point where it might just be either input delay (sometimes) or processing (nearly always) and presentation delay (almost never)
  - You can alleviate that by going to more than 3 breakdowns, e.g. by summing up the total forces style and layout, or summing the total event processing vs other processing/idle time.

src/attribution/onINP.ts
```diff
@@ -25,38 +26,109 @@ import {
...
+  // pending tasks that could get run before rendering starts). If not,
+  // use the `processingEnd` value of the last entry in the list, which
+  // should match the `renderTime` value in most cases.
+  const renderStart = longAnimationFrameEntry
+    ? longAnimationFrameEntry.renderStart
+    : lastEntry.processingEnd;
+
+  // The first entry may not have a target defined, so use the first
+  // one found in the entry list.
+  const firstEntryWithTarget = metric.entries.find((entry) => entry.target);
```
mmocny, line 109: Repeating comment: I would sort/split metric.entries on those which have interactionID first, so as to ignore early events such as pointerover.

src/attribution/onINP.ts
mmocny: I don't recall specific steps to dispatch a "trusted" click without pointer/keyboard, since most IME will still send key events as well-- but I do think it could happen.

However, click events definitely can dispatch in unique animation frames after pointer/key events.

Here's a repro: add a touchstart event listener with lots of processing time, enough that you can lift finger and send touchend/pointerup before touchstart is done.

In that case, the browser process CAN send pointerup but CANNOT send click event yet, because touchstart MIGHT call preventDefault.

src/attribution/onINP.ts
```diff
@@ -25,38 +26,109 @@ import {
...
-      loadState: getLoadState(longestEntry.startTime),
-    };
-    return;
+  const sortedEntries = metric.entries.sort((a, b) => {
+    return a.processingStart - b.processingStart;
+  });
+
+  const firstEntry = sortedEntries[0];
+  const lastEntry = sortedEntries[sortedEntries.length - 1];
+  const renderTime = firstEntry.startTime + firstEntry.duration;
```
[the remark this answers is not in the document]
mmocny: ...I also like to take Math.min(renderTime, lastEntry.processingEnd), just in case.

Specifically, for cases where an interaction does NOT have a paint but still had a long task-- then you have a 50/50 chance that duration was rounded down and now  renderTime appears less than processingTime.

src/onINP.ts
```diff
@@ -57,20 +63,88 @@ const getInteractionCountForNavigation = () => {
...
+
+// A mapping between a particular frame's render time and all of the
+// event timing entries that occurred within that frame. Since event timing
+// entries round their duration to the nearest 8sm, this ends up being
```
mmocny, line 76: Nit: `sm` -> `ms`

src/onINP.ts
mmocny: There is no single duration value that would guarantee anything, because some buffering can happen e.g. for as long as the user is holding down a finger.

Another way we could end up with >2s is just because the Performance timeline fires Observers with less than NORMAL priority tasks, and so may take a long time between observation calls-- even if the timeline data is much less than 2s apart.

---

However, because you are only doing this cleanup when new event timings arrive, I think this defends against a lot of the reasons this wouldn't work.

I think this [feature request](issues.chromium.org/issues/328902994) might also help resolve some of this.

---

I'll need more time to think through better alternatives, but this seems reasonable enough for most cases to me.  Any improvements in the future I don't think would be breaking API.

src/onINP.ts
```diff
@@ -162,31 +245,21 @@ export const onINP = (onReport: INPReportCallback, opts?: ReportOpts) => {
 
     const handleEntries = (entries: INPMetric['entries']) => {
       entries.forEach((entry) => {
-        if (entry.interactionId) {
-          processEntry(entry);
+        // Skip `first-input` entries if any `event` entries have already
```
mmocny, line 248: I haven't audited that carefully-- but we could have a case where some event timings (like hover/pointerover) may already be reported, but then we get our first discrete input.  It might still be <40ms and not be reported as and `event` but be reported as `first-input`.  So `first-input` is still valuable sometimes as a specicial event timing...

src/onINP.ts
mmocny: See comment above, but:

> that means the first input was long enough to get picked up by the Event Timing API.

I think this is the part that misled you.  Not all event timings qualify towards FID.

(Also, technically FID criteria is not 100% INP criteria.  It's actually diverging from INP as we add features.  I'd like to re-define FID as First-Interaction, but only after FID deprecation)

src/onINP.ts
```diff
@@ -57,20 +63,88 @@ const getInteractionCountForNavigation = () => {
...
+
+/**
+ * Gets the "frame-normalized" render time for the given entry.
+ * In other words, all entries passed to this function that were processed
+ * within the same frame should return the same render time. This allows you
+ * to group by this render time value.
+ * This function works by referencing `pendingEntriesGroupMap` and using
+ * an existing render time if one is found (otherwise creating a new one).
+ */
+const groupEntryByRenderTime = (entry: PerformanceEventTiming) => {
```
mmocny, line 92: Nit: you might want to make a distinction between event entry renderTime, and animation frame estimated renderTime (or eventGroup estimated renderTime).

One entry at a time it's not so bad, but when you overwrite the current entry renderTime with the frame estimated renderTime, you start to get values that might be surprisingly off by a bit.

Also, estimated frame render time could be slightly better than just the "first entry renderTime"

src/onINP.ts
```diff
@@ -84,28 +158,37 @@ const processEntry = (entry: PerformanceEventTiming) => {
...
+
     // If the interaction already exists, update it. Otherwise create one.
     if (existingInteraction) {
-      existingInteraction.entries.push(entry);
-      existingInteraction.latency = Math.max(
-        existingInteraction.latency,
-        entry.duration,
-      );
+      if (entry.duration > existingInteraction.latency) {
+        existingInteraction.latency = entry.duration;
```
mmocny, line 166: I know this isn't new code, but, is the term `latency` exposed in any way?  Feels inconsistent.

src/onINP.ts
```diff
@@ -84,28 +158,37 @@ const processEntry = (entry: PerformanceEventTiming) => {
...
       longestInteractionMap[interaction.id] = interaction;
       longestInteractionList.push(interaction);
     }
 
     // Sort the entries by latency (descending) and keep only the top ten.
     longestInteractionList.sort((a, b) => b.latency - a.latency);
-    longestInteractionList.splice(MAX_INTERACTIONS_TO_CONSIDER).forEach((i) => {
-      delete longestInteractionMap[i.id];
-    });
+    if (longestInteractionList.length > MAX_INTERACTIONS_TO_CONSIDER) {
```
mmocny, line 185: Is this guard actually needed or just making it more documented?

src/types/inp.ts
```diff
@@ -31,31 +31,86 @@ export interface INPMetric extends Metric {
  */
 export interface INPAttribution {
```
mmocny, line 32: (I have run out of time and skipping the comments in this file)

README.md
philipwalton: @mmocny I know you said you wouldn't expect an arbitrary length list of LoAFs, but what would you think about an array of LoAFs that intersect with [startTime, processingEnd]?

That's a bit more awkward than a single LoAF, but it has the benefit that it gives developers all the information about what scripts are blocking INP (for both input delay and processing time).

If we only want to expose a single LoAF then I think the reasonable options are to pick either:
1. The longest LoAF (@tunetheweb's original proposal), or
2. The LoAF that intersects with [processingStart, processingEnd]

WDYT?

src/onINP.ts
philipwalton: Ahh, yes, good call. I can go back to checking startTime/duration, but I'll still move this check to the top of this function so it doesn't get called twice in cases where the `first-input` entry *does* have an interaction ID.

src/onINP.ts
philipwalton: See me other comment as well... admittedly my comment here was confusing, but right now the logic will only skip `first-input` entries if an INP candidate has already been found, which means an `event` entry with an `interactionId` value has been dispatched.

README.md
```diff
@@ -877,31 +877,86 @@ interface FIDAttribution {
 ```ts
 interface INPAttribution {
   /**
-   * A selector identifying the element that the user interacted with for
-   * the event corresponding to INP. This element will be the `target` of the
-   * `event` dispatched.
+   * A selector identifying the element that the user first interacted with
```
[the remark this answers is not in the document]
philipwalton: Good catch. I'll update this to be the the first entry in processing order with both a `target` and an `interactionId` property.

README.md
philipwalton: > what about cases where the second interaction in the frame was the one that contributed the most processing time, and so was really "the longest"? Could that occasionally be confusing?

I do think it could be confusing, but I think that's a general problem/challenge with the INP definition and not something specific to this library.

That said, with the changes in this PR it should be possible for a developer to determine what other events were processed in the same animation frame, so hopefully that addresses the root of your concern.

src/attribution/onINP.ts
philipwalton: Thanks all for the info. Based on this, I've updated the implementation to actually look for intersections and not just look at matching `firstUIEventTimestamp`.

src/attribution/onINP.ts
philipwalton: Ok, based on these comments (and other discussions with folks) I've updated the implementation so that `processingTime` is *just* the time reported by the Event Timing API for events that were dispatched within the same animation frame as the INP candidate interaction.

src/onINP.ts
philipwalton: > However, because you are only doing this cleanup when new event timings arrive, I think this defends against a lot of the reasons this wouldn't work.

Right, that was my hope. BTW, I've updated the implementation a bit to instead just keep track of the 10 most recently reported animation frame times and assume that any new event entry that is dispatched is not going to be more than 10 frames old.

If (for some reason) that turns out not to be true in some case, what would happen is some entries might not get properly attributed to their INP candidate, and I think that's an acceptable trade off (e.g. nothing will break, it's just won't be perfect attribution).

src/attribution/onINP.ts
philipwalton: I've updated the implementation to check for LoAF intersection with Event Timing entries, so I'm going to resolve this.

src/onINP.ts
philipwalton: I updated the implementation and `Object.create(null)` is no longer used, so I'm going to resolve this.

src/onINP.ts
philipwalton: Resolving as I came up with an even more betterer way to do this...

README.md
philipwalton: I think we should stick to my current wording. I don't think we need to document all scenarios that could lead to an empty string. The main thing I want to make sure is clear is that the string CAN be empty (e.g. this value won't always have a selector).

src/attribution/onINP.ts
philipwalton: I've reverted the implementation back to just reporting the `entry.name` value from the first `event` entry corresponding to the INP candidate interaction.

README.md
philipwalton: We discussed this a bit internally and agreed that it makes the most sense to now expose an array of LoAF entries, and with my most recent changes that array will include any LoAF entry that intersects with [startTime, processingEnd] for the INP candidate interaction. @mmocny I'll tag you into that part of the code so you can take a lok.

README.md
```diff
@@ -877,31 +877,86 @@ interface FIDAttribution {
...
+   * as part of the frame where the INP candidate interaction occurred.
```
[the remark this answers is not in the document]
philipwalton: Good call. I've tried to use that everywhere now (at least in any comments that are part of the JSDocs).

src/onINP.ts
philipwalton: This usage of the term "latency" matches the definitions originally proposed here: https://web.dev/blog/responsiveness#measure_interaction_latency

Also see the definitions from this Google-internal [doc](https://docs.google.com/document/d/1mjbw-5fDNOBaDOjyz6kq54ayj6QHzNbGXtID4UpgVWw/edit?resourcekey=0-G2oQ0rOGyK9OJASVxz49TQ&tab=t.0#bookmark=id.lbfs6zy3rnb6).

src/onINP.ts
philipwalton: The `renderTime` values here are just used for grouping, they aren't exposed anywhere in the `attribution` object.

src/onINP.ts
philipwalton: I guess not technically, but it seemed more efficient to have an if statement than unconditionally calling `splice()` on an array that may not need it.

src/onINP.ts
philipwalton: Resolving this as I've updated the logic to handle `first-input` entries.

src/onINP.ts
philipwalton: Fixed.

README.md
```diff
@@ -877,31 +877,86 @@ interface FIDAttribution {
...
+   * and the browser is able to start rendering the next frame. If the user
+   * interacts again before processing has finished (which is common if event
+   * processing takes a long time), then any event processing time from
+   * subsequent interactions could get included in this processing time value
+   * since that delays the next paint for this interaction. If multiple
+   * interactions occur within the same frame, only the first interaction is
+   * considered for INP since it will be the longest.
+   *
+   * Note: if a `long-animation-frame` entry was detected for this frame
+   * its `renderStart` property will be used to mark the `processingTime`
```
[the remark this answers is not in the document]
philipwalton: As mentioned in other comments, I reverted this logic to just include that actual event processing time in the `processingTime` phase.

src/attribution/onINP.ts
```diff
@@ -25,38 +26,109 @@ import {
...
+
+  let longAnimationFrameEntry;
+  for (const loaf of longestLoAFsList) {
+    const loafEnd = loaf.startTime + loaf.duration;
+    if (
+      firstEntry.startTime === loaf.firstUIEventTimestamp ||
```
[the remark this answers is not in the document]
philipwalton: Updated to match LoAFs based on event timing intersection (as mentioned in another comment).

src/attribution/onINP.ts
```diff
@@ -25,38 +26,109 @@ import {
...
+      (loafEnd <= renderTime && loafEnd >= firstEntry.processingStart)
```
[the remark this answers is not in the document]
philipwalton: Resolving as this code has been updated.

src/attribution/onINP.ts
philipwalton: Good call. Done.

src/attribution/onINP.ts
```diff
@@ -25,38 +28,224 @@ import {
...
+  const firstEntry = metric.entries[0];
+  const renderTime = entryToRenderTimeMap.get(firstEntry)!;
+  const group = pendingEntriesGroupMap.get(renderTime)!;
+
+  const processedEventEntries = group.entries;
+  const processingStart = firstEntry.processingStart;
+  const processingEnd = group.processingEnd;
+
+  const longAnimationFrameEntries: PerformanceLongAnimationFrameTiming[] =
+    getIntersectingLoAFs(firstEntry.startTime, processingEnd);
```
philipwalton, line 208: @mmocny FYI the logic to select which LoAF entries get attributed to INP is here.

README.md
```diff
@@ -877,31 +877,86 @@ interface FIDAttribution {
...
+   * browser was first able to start processing event listeners for that
+   * interaction. This time captures the delay in processing an interaction
+   * due to the main thread being busy with other work.
    */
-  loadState?: LoadState;
+  inputDelay: number;
+
+  /**
+   * The time from when the first event listener started running in response to
+   * a user interaction until when all processing code has finished running,
```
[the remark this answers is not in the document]
philipwalton: The most recent changes only includes event processing time.

src/onINP.ts
philipwalton: > is the term `latency` exposed in any way? Feels inconsistent.

To directly answer you question though, no this is not exposed in any way. It's an implementation detail.

README.md
```diff
@@ -877,31 +877,86 @@ interface FIDAttribution {
...
+   * The time from when the user interacted with the page until when the
+   * browser was first able to start processing event listeners for that
+   * interaction. This time captures the delay before event processing can
+   * begin due to the main thread being busy with other work.
+   */
+  inputDelay: number;
+
+  /**
+   * The time from when the first event listener started running in response to
+   * the user interaction until when all event processing has finished.
```
tunetheweb, line 939:
```suggestion
   * the user interaction until when all event listener processing has finished.
```

README.md
```diff
@@ -877,31 +877,86 @@ interface FIDAttribution {
...
+   */
+  processingTime: number;
+
+  /**
+   * The time from when the browser finished processing all events for the
```
tunetheweb, line 944:
```suggestion
   * The time from when the browser finished processing all event listeners for the
```

Arguable the subsequent processing of the rendering is part of "processing the event", but not the "event listener [JavaScript code]". Trying to think of a nice way of putting it without adding the full "event listener JavaScript code" but think just saying the "event listener" captures that in a way that's reasonably clear, but not too verbose. WDYT?

src/attribution/onINP.ts
```diff
@@ -25,38 +28,224 @@ import {
...
+
+    // Store the grouped render time for this entry for reference later.
+    if (entry.interactionId || entry.entryType === 'first-input') {
+      entryToRenderTimeMap.set(entry, renderTime);
+    }
+  });
+
+  // Queue cleanup of entries that are not part of any INP candidates.
+  if (idleCallback < 0) {
+    idleCallback = requestIdleCallback(cleanupEntries);
```
tunetheweb, line 139: Do we need a fallback if `requestIdleCallback` is not supported?

I mean that shouldn't happen as all browsers that support INP support that and Safari has added it, but still feels like a guard, and fallback to immediate call, could help here.

src/attribution/onINP.ts
```diff
@@ -25,38 +28,224 @@ import {
...
+  // that occurred within 2 seconds of the most recent render time.
+  const renderTimesToKeep = new Set(
+    (previousRenderTimes as (number | undefined)[]).concat(
+      longestInteractionList.map((i) => entryToRenderTimeMap.get(i.entries[0])),
+    ),
+  );
+
+  pendingEntriesGroupMap.forEach((_, key) => {
+    renderTimesToKeep.has(key) || pendingEntriesGroupMap.delete(key);
+  });
```
tunetheweb, line 160: Is the 2 second check still here or has that been removed now?

src/attribution/onINP.ts
```diff
@@ -25,38 +28,224 @@ import {
...
+  const nextPaintTimeCandidates = [
+    AveragePresentationTime,
+    processingEnd,
+  ].concat(
+    longAnimationFrameEntries.map((loaf) => loaf.startTime + loaf.duration),
+  );
+
+  // To get the most accurate next paint time, use signals from the
+  // Event Timing API and LoAF to make the best possible guess.
+  const nextPaintTime = Math.max.apply(Math, nextPaintTimeCandidates);
```
tunetheweb, line 233: Out of curiosity why do you apply `Math` as the firts argument instead of `null`?

src/lib/whenIdle.ts
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright 2022 Google LLC
```
tunetheweb, line 2: Nit

```suggestion
 * Copyright 2024 Google LLC
```

src/lib/onHidden.ts
```diff
@@ -14,18 +14,10 @@
...
+export const onHidden = (cb: () => void) => {
+  document.addEventListener('visibilitychange', () => {
+    if (document.visibilityState === 'hidden') {
+      cb();
     }
-  };
-  addEventListener('visibilitychange', onHiddenOrPageHide, true);
-  // Some browsers have buggy implementations of visibilitychange,
-  // so we use pagehide in addition, just to be safe.
-  addEventListener('pagehide', onHiddenOrPageHide, true);
```
tunetheweb, line 30: Do we feel this is no longer the case? Is that why you removed this?

src/lib/whenIdle.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Runs the passed callback during the next idle period, or immediately
+ * if the browser's visibility state is (or becomes) hidden.
+ */
+export const whenIdle = (cb: () => void) => {
+  if (document.visibilityState === 'hidden') {
+    cb();
+  } else {
+    const handle = requestIdleCallback(cb);
```
tunetheweb, line 27: Again do we need a guard (and an immediate execute?) if this is not supported?

src/attribution/onINP.ts
```diff
@@ -90,10 +279,26 @@ export const onINP = (
   onReport: INPReportCallbackWithAttribution,
   opts?: ReportOpts,
 ) => {
+  if (!loafObserver) {
+    loafObserver = observe('long-animation-frame', handleLoAFEntries);
+  }
+  if (!eventObserver) {
+    eventObserver = observe('event', groupEntriesByRenderTime, {
+      durationThreshold: opts!.durationThreshold ?? 40,
```
tunetheweb, line 287: We've hardcoded that `40` in two places now. Could we import it from the regular `onINP` rather than duplicating it here?

src/types.ts
```diff
@@ -88,4 +88,11 @@ declare global {
     url: string;
     element?: Element;
   }
+
+  // https://w3c.github.io/long-animation-frame/#sec-PerformanceLongAnimationFrameTiming
+  interface PerformanceLongAnimationFrameTiming extends PerformanceEntry {
+    renderStart: DOMHighResTimeStamp;
+    firstUIEventTimestamp: DOMHighResTimeStamp;
```
tunetheweb, line 95: Can we remove this since we're not using it it now?

```suggestion
```

src/types/inp.ts
```diff
@@ -31,31 +31,86 @@ export interface INPMetric extends Metric {
...
+  inputDelay: number;
+
+  /**
+   * The time from when the first event listener started running in response to
+   * the user interaction until when all event processing has finished.
+   */
+  processingTime: number;
+
+  /**
+   * The time from when the browser finished processing all events for the
```
tunetheweb, line 98: Duplicate suggestion to above so you don't change one but not the other:

```suggestion
   * The time from when the browser finished processing all event listeners for the
```

src/types/inp.ts
```diff
@@ -31,31 +31,86 @@ export interface INPMetric extends Metric {
...
+   * The time from when the user interacted with the page until when the
+   * browser was first able to start processing event listeners for that
+   * interaction. This time captures the delay before event processing can
+   * begin due to the main thread being busy with other work.
+   */
+  inputDelay: number;
+
+  /**
+   * The time from when the first event listener started running in response to
+   * the user interaction until when all event processing has finished.
```
tunetheweb, line 93: Duplicate suggestion to above so you don't change one but not the other:

```suggestion
   * the user interaction until when all event listener processing has finished.
```

test/utils/nextFrame.js
```diff
@@ -15,8 +15,8 @@
  */
 
 /**
- * Returns a promise that resolves once the browser window has loaded and
- * all load callbacks have finished executing.
+ * Returns a promise that resolves once the browser has run the next
+ * animation frame.
  * @return {Promise<void>}
  */
 export function nextFrame(state) {
```
tunetheweb, line 22: Not changed in this PR but do we need `state` here? Doesn't ever seem to be called with an argument.

```suggestion
export function nextFrame() {
```

Also function name could maybe be clearer (`awaitNextFrame`) but again not changed here so fine to leave as is.

README.md
```diff
@@ -877,31 +877,86 @@ interface FIDAttribution {
...
-   * A selector identifying the element that the user interacted with for
-   * the event corresponding to INP. This element will be the `target` of the
-   * `event` dispatched.
+   * A selector identifying the element that the user first interacted with
+   * as part of the frame where the INP candidate interaction occurred.
+   * If `interactionTarget` is an empty string, that generally means the
+   * element was removed from the DOM as part of the interaction.
    */
-  eventTarget?: string;
+  interactionTarget: string;
```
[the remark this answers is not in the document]
philipwalton: I think they were originally optional because INP could be zero if interactionCount was greater than 0 but no event entries were dispatched. However, once we started also observing `first-input` entries, I think they should always be set.

README.md
philipwalton: Yeah, I think so. Good call.

src/attribution/onINP.ts
philipwalton: > Do we need a fallback if `requestIdleCallback` is not supported?

I don't think so? Both Firefox and Chrome support this API, so the only risk would be if Safari suddenly adds support for Event Timing but doesn't enable rIC for some reason. And at that point we could just update the library to "enable Safari support". WDYT?

src/attribution/onINP.ts
philipwalton: It's been removed in favor of what's explain in the comments above ([line 144-148](https://github.com/GoogleChrome/web-vitals/blob/a41d463664913ab734f0b6685bf4b57777df2f45/src/attribution/onINP.ts#L144-L148)).

tunetheweb: So can we remove it from the comment?

src/attribution/onINP.ts
philipwalton: You could do either, but since the `this` context in a normal call to `Math.max()` would be `Math`, it seemed to make sense to use that.

philipwalton: Interestingly it looks like the [examples on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply) use `null` (though they don't explain why). TBH I don't think it matters at all and presumably they both minify/compress to the same code size.

tunetheweb: Yeah that's why I asked after checking there! Both seem to work but was curious.

src/attribution/onINP.ts
philipwalton: Fair point. I don't want to extend the public API of `onINP.js` but I'll find a shared place to import it from.

src/lib/onHidden.ts
philipwalton: Yeah, TBH I've wanted to remove this for a long time.

Last I checked (when working on the PendingBeacon polyfill 1.5 years ago) all bugs I was aware of have been fixed. And for old browsers that still have these bugs, that primarily affects Safari (which doesn't support the metrics that this is needed for) and old Chrome (which also doesn't support these metrics), so I think it's 100% safe to remove this.

It's also potentially racy in the sense that `document.visibilityState` will show as "visible" in the pagehide event, so the logic added to this PR that checks visibility state first and then calls `onHidden()` could break in a browser that doesn't reliably support `visibilitychange` anyway.

There's also the issue that if your app's code is doing something on `visibilitychange` and then you're making some call to the web-vitals library, it would be too late (e.g. web-vitals code already ran on `pagehide`).

src/attribution/onINP.ts
philipwalton: Oops! (Good catch.)

src/attribution/onINP.ts
philipwalton: Let's keep it as `Math` (I do think that's "more correct"). FWIW I'm 99% sure I remember a time when calling `console.log.apply(null, ['Hello World!])` would error because it didn't have the correct `this` context, so you had to do `console.log.apply(console, ['Hello World!])`.

src/attribution/onINP.ts
philipwalton: Ok, I added (because the fallback is trivial), but I think we should revisit this if/when Safari ships support.

As a meta point, I don't want this library (or any public code our team creates) to promote the sense that modern web APIs are not 100% reliable and that you should be weary of using them without fallbacks, transpiling, bundling polyfills, etc.

README.md
mmocny: The loaf that intersects with processingStart / processingEnd I would consider "the" LoAF for this Interaction, and it feels uniquely important vs just a list of all of them.

Maybe if we returned the list in reverse chronological order (which matches order of importance) that would be fine.

---

The one single preceding LoAF might well be useful for debugging input delay.

However, an long string of preceding LoAFs that overlap startTime, I think are more likely to be red-herrings.

Input Delay is only affected by Tasks/LoAFs when the event is already queued on the main thread (and so blocking processingStart).  But startTime is based on hardware timeStamp, and the event can be stuck in hardware, or platform, or browser process... in this case LoAFs aren't actually related.

Once the event is on the main thread, it is scheduled with the HIGHEST priority and there cannot be a case where we somehow do multiple animation frames before deciding to schedule the event processing.

Unless I'm overlooking something ;)

Of course, that is based on today's specifics implementation details of Chromium.  It's already different on Firefox, though they don't report LoAF, anyway.

---

I would personally prefer:
- A single LoAF for "the" frame
- A list of extra LoAF, for Input Delay

...but we try to constrain that list to our best guess of what actually affected the interaction, based on platform and version, and can start with the single most recent LoAF that is `event.startTime <= loaf.endTime < event.processingStart` or something.

But I would also be fine with just a single list, especially in reverse chronological order.

src/attribution/onINP.ts
```diff
@@ -25,38 +28,224 @@ import {
...
+  // The first interaction entry may not have a target defined, so use the
+  // first one found in the entry list.
+  // TODO: when the following bug is fixed just use `firstInteractionEntry`.
+  // https://bugs.chromium.org/p/chromium/issues/detail?id=1367329
+  const firstEntryWithTarget = metric.entries.find((entry) => entry.target);
+
+  // Since entry durations are rounded to the nearest 8ms, taking the average
+  // of a group of entries that we know were rendered in the same animation
+  // frame should give a better estimate of the "true" render time.
+  const AveragePresentationTime =
```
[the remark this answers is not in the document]
mmocny: I think midpoint of the min and max will be the most likely to be closest.

Basically, every presentation time value is a random value within the range of [-4,+4].

If many values are close to +4 and a single value is close to -4, the real presentation time is still guaranteed to be in the middle and isn't more likely to group with the majority...

(That said, I think this doesn't matter. You might even want to just make it pretty and round off the presentation time to some multiple of 8 or something.)

src/attribution/onINP.ts
mmocny: Nit: Is spread operator less supported? `Math.max(...nextPaintTimeCandidates)`

README.md
tunetheweb: > The loaf that intersects with processingStart / processingEnd I would consider "the" LoAF for this Interaction, and it feels uniquely important vs just a list of all of them.

I don't get why that's the important one rather than the longest one?

Like if you had a 200ms input delay (e.g. due to a long rAF) and then 10ms event handler I don't see why the event handler LoAF is "the" LoAF for the interaction as apposed to the one that actually caused the delay?

Though I do agree the event handler one is likely to be the longest one in most cases so maybe a bit of a moot point.

> I would personally prefer:
>
> - A single LoAF for "the" frame
> - A list of extra LoAF, for Input Delay

I definitely prefer a single list. It's all complicated enough and think most people are likely to just pull out the script info rather than do any deep diving in to the other LoAF info anyway. So one list just makes it less complicated.

src/attribution/onINP.ts
philipwalton: I didn't use the spread operator here (or any other iterator stuff, e.g. for...of) because then when the bundle is transpiled to ES5 it'll include tons of Babel helper code bloat.

At some point in the future I think we can decide to not transpile to ES5, but I'd want to be intentional about that change and make sure we're not going to be breaking anyone.

README.md
philipwalton: > The one single preceding LoAF might well be useful for debugging input delay.
> 
> However, an long string of preceding LoAFs that overlap startTime, I think are more likely to be red-herrings.

In all of the testing I did for this, I never once saw more than 2 LoAF entries reported in the array. Can there really be  a string of long preceding LoAFs? If so, do you have any suggestions for how I can repro that, because I'd also want to test that scenario.

mmocny: I think that you can only have 1 single "Main thread input delay" LoAF and 1 single "Main thread processing" LoAF, and sometimes those are even the same LoAF.

But you can in theory have variable length extra LoAF for non-main thread input delay.  These LoAFs are red-herring and should just be ignored, but if you naively use startTime as the filter, you will include them.

So the question becomes, how can you force non-main thread input delay?

---

One way is the 300ms non-mobile-viewport tap-delay thing.  You click event will have a really large non-main input delay.  If you happen to have LoAFs on main, those would get reported-- but these aren't the cause of the delay.

Another way on MacOS that is somewhat consistent is to use native drag & drop (highlight and drag text, or drag an image) which causes browser-side jank often delaying event dispatch.  Again, if you happen to also have LoAFs at that time, those would overlap but wouldn't really be the cause.

More generally, you cannot force these types of delays consistently, because that only happens with system jank.  And we try hard not to Jank browser main thread, but it does still happen all the time.

philipwalton: I suppose another option could be to keep this as an array, but expose a max of 1 that overlaps with processing and then 1 that overlaps with input delay (the last one, chronologically). So a max length of 2 total.

As a meta comment, at this point I'd rather pick an API shape that is flexible for future additions and unlikely to require breaking changes if we want to add anything. I think there's a higher chance that we later decide that individual LoAF entries doesn't give developers everything they need than if we offer an array of intersecting entries.

src/attribution/onINP.ts
```diff
@@ -25,38 +31,228 @@ import {
...
+    }
+  });
+
+  // Queue cleanup of entries that are not part of any INP candidates.
+  if (idleHandle < 0) {
+    idleHandle = whenIdle(cleanupEntries);
+  }
+};
+
+const cleanupEntries = () => {
```
philipwalton, line 146: @mmocny if you have time to review this function, that would be helpful. This function isn't part of the public API, but if any part of this logic fails (or is incorrect) then it could result in issues, so I'd love a second pair of eyes on it.

In short, this function tries to remove references to `event` and LoAF entries that are no longer needed (so we're not just storing everything in memory), but there's the potential that I could remove a reference to something that will later been needed and then it won't get reported in the attribution object (or worse, there will be an error because other code will try to reference a property of an object that no longer exists).

src/types/inp.ts
```diff
@@ -31,31 +31,86 @@ export interface INPMetric extends Metric {
...
+  nextPaintTime: DOMHighResTimeStamp;
+
   /**
-   * The time when the user interacted for the event corresponding to INP.
-   * This time will match the `timeStamp` value of the `event` dispatched.
+   * The type of interaction. This will be either 'pointer' or 'keyboard'
+   * since those are the only types of interactions considered for INP.
    */
-  eventTime?: number;
+  interactionType: string;
```
tunetheweb, line 65: This is back to being `pointerup` rather than `pointer`.

So we should either fix that or change the text here. Going to comment some more on this further down.

README.md
```diff
@@ -877,31 +877,86 @@ interface FIDAttribution {
...
-  eventType?: string;
+  nextPaintTime: DOMHighResTimeStamp;
+
   /**
-   * The `PerformanceEventTiming` entry corresponding to INP.
+   * The type of interaction. This will be either 'pointer' or 'keyboard'
+   * since those are the only types of interactions considered for INP.
    */
-  eventEntry?: PerformanceEventTiming;
+  interactionType: string;
```
tunetheweb, line 911: Same comment as above.

src/attribution/onINP.ts
```diff
@@ -25,38 +31,215 @@ import {
...
+    longAnimationFrameEntries.map((loaf) => loaf.startTime + loaf.duration),
+  );
+
+  const nextPaintTime = Math.max.apply(Math, nextPaintTimeCandidates);
+
+  (metric as INPMetricWithAttribution).attribution = {
+    interactionTarget: getSelector(
+      firstEntryWithTarget && firstEntryWithTarget.target,
+    ),
+    interactionType: firstEntry.name,
```
tunetheweb, line 233: So I've been playing with this and don't think picking the first one is ideal.

Most events are likely to be on the `click` event rather than `pointerup` but at the moment it's reporting `pointerup` cause that comes first.

We could either:
1. Pick the last one (which likely works for pointer/click but pay not for others).
2. Have a lookup to `pointer`/`keyboard` as was the original intent (and still in comments, see feedback above).
3. Have some complicated lookup to order in processing order.

I think I'm veering towards 2. since it is reporting the WHOLE `pointer/`keyboard` interaction and not just the specific event handler.

README.md
philipwalton: Good catch, thanks!

src/types/inp.ts
philipwalton: Good catch, thanks! I updated the comment as well as the type (to be more explicit).

src/attribution/onINP.ts
philipwalton: In the previous discussion, I thought you and Michal were arguing that there could be cases where a `click` interaction happens without either a keyboard or pointer, and that's what convinced me to change it back.

I could change it again and just include "click" as a fallback, though I thought it was simpler to just stick with the existing behavior of reporting the event type (which is also less likely to be a divergence in people's analytics data).

tunetheweb: Michal said that it's possible, and agree we shouldn't exclude `click`.

However, I'm saying it's weird to see the type reported as `pointerup` when that has no event handler, and `click` has a long event handler. And I think that's quite a common scenario that both will be in same frame and it'll be the `click` one that really causes the long INP.

And the truth is we aren't just reporting on `pointerup` - we're reporting on the whole pointer interaction (`pointerup` AND `click`). So maybe the type shouldn't claim to be more specific than it is?

So I'm saying we should do something like this (or this idea in better JavaScript):

```js
interactionType = firstEntry.name.replace('click', 'pointer').replace('up', '').replace('down', '');
```

src/onINP.ts
```diff
@@ -160,34 +77,20 @@ export const onINP = (onReport: INPReportCallback, opts?: ReportOpts) => {
     let metric = initMetric('INP');
     let report: ReturnType<typeof bindReporter>;
 
+    const durationThreshold = opts!.durationThreshold ?? 40;
```
tunetheweb, line 80:
```suggestion
    const durationThreshold = opts!.durationThreshold ?? DEFAULT_DURATION_THRESHOLD;
```

test/e2e/onINP-test.js
```diff
@@ -435,33 +454,67 @@ describe('onINP()', async function () {
...
-      assert(interactionIDsMatch(inp1.entries));
-      assert(inp1.entries[0].interactionId > 0);
+      assert(allEntriesPresentTogether(inp1.entries));
       assert.match(inp1.navigationType, /navigate|reload/);
 
-      const clickEntry = inp1.entries.find((e) => e.name === 'click');
-      assert.equal(inp1.attribution.eventTarget, 'html>body>main>h1');
-      assert.equal(inp1.attribution.eventType, clickEntry.name);
-      assert.equal(inp1.attribution.eventTime, clickEntry.startTime);
+      assert.equal(inp1.attribution.interactionTarget, 'html>body>main>h1');
```
tunetheweb, line 460: Could we assert that a LoAF exists (for browser that support LoAF)? Don't think it's necessary to check the entries but would be good to confirm it's there and not null.

tunetheweb (approved):
LGTM with two outstanding questions.

Awesome work here!

src/attribution/onINP.ts
tunetheweb: Changing my mind on this now as per https://github.com/GoogleChrome/web-vitals/pull/442#discussion_r1539967625

(just in case anyone ever comes back to this at some point!)

src/attribution/onINP.ts
brendankenny: I think I agree with Barry.

In DevTools I don't mind the grouping as `Pointer` and `Keyboard`. That's generally helpful, though I haven't really considered the `click` corner case.

As Barry said, though, I find it distracting when I get told something's e.g. a slow `pointerup` and the page doesn't have a `pointerup` listener and I have to either go back to my knowledge of order of events or break out DevTools and look at a trace.

It's the same issue with those events that get an interactionId and those that don't, too, like if there's a super heavy event listener on `mousedown` but a tool reports only `pointerdown` which has no listener attached.

Seems better to just attribute the category of listeners with all that in mind?

src/lib/getNavigationEntry.ts
```diff
@@ -18,7 +18,7 @@ export const getNavigationEntry = ():
   | PerformanceNavigationTiming
   | undefined => {
   return (
-    window.performance &&
+    self.performance &&
```
brendankenny, line 21: oh, you're actually changing it in code as well, in which case `self` makes more sense for old browser support. Buy why now? I'm so curious.

src/attribution/onINP.ts
philipwalton: I updated this to just be 'pointer' or 'keyboard' (where `click` becomes pointer regardless of how it is initiated). Turns out `click` events initiated from keyboards are weird and have lots of issues, and it's non-trivial to determine whether or not a `click` was initiated from a keyboard event because they have different `interactionId` values.

Doing this appears to be consistent with DevTools, though an alternative approach would be to assume that any time a `click` event is the first event dispatched in an animation frame, it was most likely initiated via keyboard (though I'm not sure if that's 100% reliable).

src/lib/getNavigationEntry.ts
philipwalton: I standardized on `self` because both were used throughout the project. Though looking at it closer now, I see `self` was only ever used in test files, so maybe it doesn't matter?

brendankenny: Ah, yeah, makes sense. Probably doesn't matter, but also fine either way. I had no idea `self` has been around since at least IE6.  I thought it was a web-worker-era addition.

test/e2e/onINP-test.js
philipwalton: Added. I couldn't before since LoAF wasn't stable, and I didn't really want to add flags to the tests (in case that had unintended consequences).

PR #445, opened 2024-03-15
add navigationTypes Description to explain kebab case application

gilbertococchi:
Please review the following PR to explain that Navigation Timing API Entries would be slightly different in the WebVitals.js implementation due to kebab case convention.

### DIFF
--- a/src/types/base.ts
+++ b/src/types/base.ts
@@ -77,6 +77,7 @@ export interface Metric {
    * This will be the value returned by the Navigation Timing API (or
    * `undefined` if the browser doesn't support that API), with the following
    * exceptions:
+   * - all entries are converted to hypens (e,.g. `back_forward` is returned as `back-forward`)
    * - 'back-forward-cache': for pages that are restored from the bfcache.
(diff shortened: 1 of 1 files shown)


src/types/base.ts
```diff
@@ -77,6 +77,7 @@ export interface Metric {
    * This will be the value returned by the Navigation Timing API (or
    * `undefined` if the browser doesn't support that API), with the following
    * exceptions:
+   * - all entries are converted to hypens (e,.g. `back_forward` is returned as `back-forward`)
```
tunetheweb, line 80:
```suggestion
   * - all entries are converted to hyphens (e.g. `back_forward` is returned as
   *   `back-forward`).
```

tunetheweb (changes requested):
Can you also make the same change in base.ts?

philipwalton:
Thanks @gilbertococchi, I wanted to tweak the language of this slightly, so I added this here: https://github.com/GoogleChrome/web-vitals/commit/06b0a518096aefd29add3e5f860c0eb700248314

PR #453, opened 2024-03-31
Rename `TTFBAttribution` fields from `*Time` to `*Duration`

philipwalton:
This PR continues the renaming work from #450 and updates the `TTFBAttribution` interface with the same changes—all properties whose name ended in `*Time` have been renamed to `*Duration`, in order to make it more clear that these are durations of time rather than timestamps.

```diff
interface TTFBAttribution {
-  waitingTime: number;
+  waitingDuration: number;

-  dnsTime: number;
+  dnsDuration: number;

-  connectionTime: number;
+  connectionDuration: number;

-  requestTime: number;
+  requestDuration: number;
}
```

**⚠️ Important:** this is a breaking change for anyone using the attribution build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1028,21 +1028,21 @@ interface TTFBAttribution {
    * DNS lookup begins. This includes redirects, service worker startup, and
    * HTTP cache lookup times.
    */
-  waitingTime: number;
+  waitingDuration: number;
   /**
    * The total time to resolve the DNS for the current request.
    */
-  dnsTime: number;
+  dnsDuration: number;
   /**
    * The total time to create the connection to the requested domain.
    */
-  connectionTime: number;
+  connectionDuration: number;
   /**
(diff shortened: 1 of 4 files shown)


tunetheweb (changes requested):
Change itself looks good but can we update README and CHANGELOG?

Would also be nice to add redirect time as per #395

philipwalton:
> Change itself looks good but can we update README and CHANGELOG?

Doh! Not sure how I forgot to update the README. Updated!

BTW, I've generally only been updated the CHANGELOG before doing a release, but I plan to do that as soon as these changes are in.

> Would also be nice to add redirect time as per #395

Yeah, I'm open to that, but I think it'd be worth discussing (and also doing as a separate PR).

rviscomi (approved):

PR #454, opened 2024-04-01
Add an Upgrading to v4 doc

philipwalton:
This PR adds an "Upgrading to v4" doc that lists the full set of breaking changes from v3 in a single place, to make it easier for folks to upgrade.

Note: this is different from the CHANGELOG, which does list all changes but only lists them by PR title. The new upgrading doc makes it so folks can easily upgrade without needing to read through the PR descriptions/discussions.

### DIFF
--- a/docs/upgrading-to-v4.md
+++ b/docs/upgrading-to-v4.md
@@ -1,52 +1,59 @@
 # Upgrading to v4
 
-This document lists the full set of breaking changes that need to be addressed when upgrading from version 3.x to 4.x, across both the "standard" and "attribution" builds (see [build options](/#build-options) for details). For a list of all changes see the [CHANGELOG](/CHANGELOG.md).
+This document lists the full set of changes between version 3 and version 4 that are relevant to anyone wanting to upgrade to the new version. This document groups changes into "breaking changes" and "new features" across both the "standard" and "attribution" builds (see [build options](/#build-options) for details).
 
-Version 4 is currently in beta, you can install it via the `next` tag on npm:
+Note: version 4 is currently in beta, you can install it via the `next` tag on npm:
 
 ```sh
 npm install web-vitals@next
 ```
 
-## Standard build
+## ⚠️ Breaking changes
 
-### General
+### Standard build
 
-- The "base+polyfill" build has been removed (see [#435](https://github.com/GoogleChrome/web-vitals/pull/435))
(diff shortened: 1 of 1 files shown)


docs/upgrading-to-v4.md
```diff
@@ -0,0 +1,48 @@
+# Upgrading to v4
```
rviscomi, line 1: Also worth mentioning that it's available now in beta, and how to get it?

rviscomi (approved):

docs/upgrading-to-v4.md
```diff
@@ -0,0 +1,48 @@
...
+- Previously deprecated `getXXX()` functions have been removed (see [#435](https://github.com/GoogleChrome/web-vitals/pull/435))
+
+### `INPMetric`
+
+- `entries`
+  - Updated to only include entries with matching `interactionId` and `duration` values (previously it was only matching `interactionId` values, see [#442](https://github.com/GoogleChrome/web-vitals/pull/442)).
+
+## Attribution build
+
+### `INPAttribution`
```
rviscomi, line 22: In the CHANGELOG you also mention "Add INP breakdown timings and LoAF attribution" is a breaking change. Should that be documented here as well?

docs/upgrading-to-v4.md
philipwalton: Fair point. Added.

docs/upgrading-to-v4.md
philipwalton: My intention there was to make it clear that that PR included breaking changes (the renaming/removing of attribution properties); however, the addition of the breakdown timings and LoAF attribution itself is not a breaking change (which I can see how that's confusing).

I wanted to keep the "Upgrading to v4" doc scoped as narrowly as possible to just the changes someone would need to make if they wanted to upgrade to the new version without necessarily adding any new features. Do you think it'd be better to include the new features as well?

rviscomi: Either way SGTM. I guess it does seem useful for this doc to cover both breaking changes and new features for anyone interested in upgrading. Otherwise it could be easy to miss that this only covers breaking changes. Maybe different sections might make it clearer? Something like this:

## ⚠️ Breaking changes

...

## 🚀 New features

...

philipwalton: Added.

PR #457, opened 2024-04-03
Fix unit tests broken in v4

tunetheweb:
Unit tests in v4 are broken since merging #442 as it exports a function using `self` which doesn't exist in node contexts.

### DIFF
--- a/src/lib/whenIdle.ts
+++ b/src/lib/whenIdle.ts
@@ -17,7 +17,9 @@
 import {onHidden} from './onHidden.js';
 import {runOnce} from './runOnce.js';
 
-const rIC = self.requestIdleCallback || self.setTimeout;
+// Used by unit test so expose via globalThis with fallback for unsupporting browsers
+const rIC =
+  globalThis.requestIdleCallback || globalThis.setTimeout || self.setTimeout;
 
 /**
  * Runs the passed callback during the next idle period, or immediately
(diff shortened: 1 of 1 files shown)


philipwalton:
Gah, can't believe I didn't notice the failing unit tests. This could also impact anyone running the code in node SSR, so we should fix ASAP.

Rather than using `globalThis` (which will error in some older browsers), I think it would be better to just move the `const rIC = self.requestIdleCallback || self.setTimeout;` line within the `whenIdle()` function. This is what we do for cases that reference `document` in other parts of the code.

src/lib/whenIdle.ts
```diff
@@ -17,7 +17,10 @@
 import {onHidden} from './onHidden.js';
 import {runOnce} from './runOnce.js';
 
-const rIC = self.requestIdleCallback || self.setTimeout;
+// Used by unit test so expose via globalThis with fallback for unsupporting browsers
+const rIC = globalThis
```
philipwalton, line 21: Rather than using `globalThis` (which will error in some older browsers), I think it would be better to keep this line as it was but move it within the `whenIdle()` function (to line 27 in the original file).

This is similar to what we do for cases that reference `document` in other parts of the code.

tunetheweb: Ah clever. That seems to work. Done.

philipwalton (approved):

PR #458, opened 2024-04-03
Split `waitingDuration` to make it easier to understand redirect delays

### LINKED ISSUE #395: Add redirect time to TTFB breakdown
Redirects can cause considerable impact to LCP times. We should add this to TTFB attribution.

Note [concerns have been raised that this is a cross-origin leak](https://github.com/w3c/navigation-timing/issues/160) and so may be removed from Web APIs, but if that happens that will need to lead to discussions on the wider impact to Core Web Vitals (can you measure LCP accurately without this or do redirects have to move outside of LCP then?)

Also what to do with the current `waitingTime` measure which currently includes redirect time? Do leave as is to avoid making a breaking change? Or change in the next major release to subtract redirect time from this. Redirect time is also only part of what it measures and from the [PerformanceNavigationTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming) there's also unload time before it, and service worker and HTTP Cache time after it. So should they all be included and waitTime be full deprecated (but remain as is for now)?

tunetheweb:
Fixes #395 

This splits `waitingDuration` (formerly known as `waitingTime` in v3) into two:
- `waitingDuration` - which the time until (`workerStart` or `fetchStart`). This will mostly be redirect time but I've avoided explicitly calling that as [`redirectStart/End` can't be used](https://github.com/w3c/navigation-timing/issues/160) and it's easy to see when testing instances where looking at `workerStart` or `fetchStart` still leaves a non-0 amount of time that is not redirects.
- `cacheDuration` which is the time from then until `domainLookupStart`. This will mostly be HTTP cache time, except for service-worker controlled sites when it will be service worker time.

See PerformanceNavigationTiming breakdown for more context:
image

As part of this I've also moved any delays between `connectEnd` and `requestStart` from being attributed to `connectionDuration` (i.e before the delay) to `requestDuration`  (i.e after the delay). This helps avoiding attributing this delay to `connectionDuration` for service-worker controlled sites when this should be 0, but without adding an extra breakdown that would not be that useful.


test/e2e/onTTFB-test.js
```diff
@@ -261,8 +261,19 @@ describe('onTTFB()', async function () {
 
       const navEntry = ttfb.entries[0];
       assert.strictEqual(
-        ttfb.attribution.waitingDuration,
-        navEntry.domainLookupStart,
+        ttfb.attribution.redirectDuration,
+        Math.max(
```
philipwalton, line 265: Nit: I don't think `Math.max()` is needed here (or anywhere you're not subtracting `activationStart`).

philipwalton (changes requested):

src/types/ttfb.ts
```diff
@@ -31,11 +31,16 @@ export interface TTFBMetric extends Metric {
  */
 export interface TTFBAttribution {
   /**
-   * The total time from when the user initiates loading the page to when the
-   * DNS lookup begins. This includes redirects, service worker startup, and
+   * The total time spent in redirects before the current page processes the
+   * request. Npote in future this may only include same-origin redirects.
```
philipwalton, line 35:
```suggestion
   * request. Note in future this may only include same-origin redirects.
```

docs/upgrading-to-v4.md
```diff
@@ -36,7 +36,7 @@ npm install web-vitals@next
 
 #### `TTFBAttribution`
 
-- **Renamed** `waitingTime` to `waitingDuration` (see [#453](https://github.com/GoogleChrome/web-vitals/pull/453)).
+- **Split** `waitingTime` into `redirectDuration` and `cacheDuration` ([#458](https://github.com/GoogleChrome/web-vitals/pull/458))
```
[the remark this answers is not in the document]
tunetheweb: Done.

src/attribution/onTTFB.ts
```diff
@@ -42,7 +51,12 @@ const attributeTTFB = (metric: TTFBMetric): void => {
     );
 
     (metric as TTFBMetricWithAttribution).attribution = {
-      waitingDuration: dnsStart,
+      // Set redirectend to be based on fetchStart and workerStart to get most
+      // accurate redirect time available now.
+      // Note this may change in future. See
+      // https://github.com/w3c/navigation-timing/issues/160
+      redirectDuration: redirectEnd || cacheStart,
```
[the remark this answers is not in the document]
tunetheweb: My thinking here was that if `redirectEnd` is present (i.e. for same origin-redirects), then using that is a more accurate measure of the actual redirect time (and avoids a 1ms time you see when it's not present). But maybe it's more confusing to have it based on different end times in different circumstances.

test/e2e/onTTFB-test.js
tunetheweb: Fixed.

src/attribution/onTTFB.ts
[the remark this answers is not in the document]
tunetheweb: Ah you're not saying not to use that logic. Just to include it all in `cacheStart`.

OK I've done that, but renamed `cacheStart` to `redirectEnd` to make it more obvious why we are using that.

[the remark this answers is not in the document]
tunetheweb: Actually I changed my mind and moved it back to `cacheStart` since the other times were all `start***` times.

src/attribution/onTTFB.ts
```diff
@@ -24,10 +24,22 @@ import {
 } from '../types.js';
 
 const attributeTTFB = (metric: TTFBMetric): void => {
-  if (metric.entries.length) {
+  // Only attribute if you have a non-zero responseStart value and a
+  // navigationEntry entry.
+  // E.g. For cross-origin redirects, Safari has a workerStart time but no
+  // TTFB time, so do not provide attribution for that.
+  if (metric.entries.length && metric.entries[0].responseStart) {
```
philipwalton, line 31: I believe the logic here already includes a check for valid `responseStart` values.

src/attribution/onTTFB.ts
```diff
@@ -24,10 +24,22 @@ import {
...
     const navigationEntry = metric.entries[0];
     const activationStart = navigationEntry.activationStart || 0;
 
+    // Measure from workerStart or fetchStart so any service worker startup
+    // time is included in cacheDuration (which also includes other sw time
+    // anyway, that cannot be accurately split out cross-browser).
+    const waitEnd = Math.max(
+      (navigationEntry.workerStart || navigationEntry.fetchStart) -
```
philipwalton, line 39: Maybe it makes more sense to put the SW startup time in "waiting" WDYT?

philipwalton (changes requested):

src/attribution/onTTFB.ts
tunetheweb: Oh good spot that the main metric checks this with `isInvalidTimestamp`. Removed in https://github.com/GoogleChrome/web-vitals/pull/458/commits/d2fb0aed6189985f1509886d3819961e34a3247e

src/attribution/onTTFB.ts
tunetheweb: As per the comment above this, I preferred it in `cacheDuration` as then get a better redirect time in `waitingDuration`.

Also all the other service worker work is in `cacheDuration` so nice to haver them together (given we can't separate them out cleanly as discussed offline).

But I do agree it's technically "waiting time".

philipwalton (approved):

PR #460, opened 2024-04-04
Improved testing options

tunetheweb:
Improved version of #286 

Changes:
- Allows you to run: `npm run test:e2e -- --browser=chrome --metric=TTFB` for example to limit tests to a particular browser and/or metric.
- Adds a precommit test for any stray `it.only`'s, `depends.only`'s, or `browser.debug`'s accidentally left around while testing.
- Adds instructions to contributing file.
- Moves Code of Conduct and Contributing files to more commonly known locations to make them easier to find. 
- Adds a unit test github action (and also updates github actions versions while at it).


nucliweb (approved):
👏

philipwalton:
@tunetheweb I made a few tweaks to the code that I thought make things a bit clearer. Let me know if you have any issues with these changes, otherwise feel free to merge.

philipwalton (approved):

tunetheweb:
@philipwalton , I'll see your one array and raise you some more :-)

tunetheweb:
I also added a unit test GitHub action.

I tried the e2e test but couldn't get it working. Even restricting it to Chrome required a number of extra options (` args: ['headless', 'disable-gpu', 'remote-debugging-pipe'],`) and still fails CLS tests (though the others pass). Let's leave for now and might revisit later. Apparently it can be [a bit flakey anyway](https://stackoverflow.com/questions/48450594/selenium-timed-out-receiving-message-from-renderer), plus our tests are a little flakey themselves, so maybe not a good fit for CI.

tunetheweb:
@philipwalton happy to merge? Or wanna have one last look after latest changes?

test/e2e/onCLS-test.js
```diff
@@ -460,23 +463,25 @@ describe('onCLS()', async function () {
     assert.match(cls1.navigationType, /navigate|reload/);
 
     await clearBeacons();
-    await triggerLayoutShift();
-
     await stubForwardBack();
-    await beaconCountIs(1);
```
tunetheweb, line 466: @philipwalton I don't understand this logic and I think it was causing the tests to be flakey.

Why trigger a layout shift before the back/forward? Doesn't that mean it should really be for the previous page (before the "forward")? In which case we'll have another CLS reported for the original page (depending if bigger or not), then a 0-CLS reported for the "forward"? Yet you're only checking 1 CLS below.

IT seems to work a lot of the time locally, but not on CI. I suspect the "depending if bigger or not" is causing issues.

I've simplified this to:
- Load page with layout shifts. Check they are reported.
- Do a forward/back. Check 0 CLS is reported for the "forward" page navigation.
- Do a layout shift and then a visibility change. Check non-0 CLS with 1 layout shift is report for the "back" navigation.

philipwalton: > @philipwalton I don't understand this logic and I think it was causing the tests to be flakey.
>
> Why trigger a layout shift before the back/forward? Doesn't that mean it should really be for the previous page (before the "forward")?

Yes, it is for the previous visit. In this test there are 3 visits: 1 for the original navigation and then 2 after a bfcache restore. That's why there are 3 `cls` variables all of which have different IDs (and the test asserts that they are all different).

The reason I wanted to have 3 visits (specifically 2 bfcache restores) is to test that the library doesn't "stop listening" after the first bfcache restore.

AFAICT, all your changes do is remove the layout shifts from the 2nd visit (which is fine; we do have a test to ensure that visits with no CLS are reported as 0, but more tests don't hurt). It's not clear to me why that would reduce flakiness though, since in both cases the logic to trigger a shift and then a navigation is the same. Roughly:

```js
await clearBeacons();
await triggerLayoutShift();

await stubVisibilityChange('hidden'); // or stubBackForward(), which is also a visibility change.
await beaconCountIs(1);
```

tunetheweb: OK issue was the layoutshifts weren't always finshing before the stubBackForward ran so sometimes they'd bleed into the next page and not report in the previous pages's CLS.

Added a `await nextFrame();` to `triggerLayoutShift` to avoid this happening.

test/e2e/onCLS-test.js
```diff
@@ -662,11 +662,14 @@ describe('onCLS()', async function () {
     if (!browserSupportsCLS) this.skip();
 
     await navigateTo('/test/cls?hidden=1');
+    // Wait for a frame to be painted.
+    await nextFrame();
 
     await stubForwardBack();
+    await firstContentfulPaint();
```
philipwalton, line 669: Is this needed? It feels a bit weird to await FCP for the original page *after* a navigation. I get that the original navigation was to a page that didn't paint, but I'm not sure how obvious that is.

test/e2e/onCLS-test.js
```diff
@@ -662,11 +662,14 @@ describe('onCLS()', async function () {
     if (!browserSupportsCLS) this.skip();
 
     await navigateTo('/test/cls?hidden=1');
+    // Wait for a frame to be painted.
+    await nextFrame();
```
philipwalton, line 666: Do you know why the `nextFrame()` is needed here? Given that the page is loaded in the hidden state, I wonder if waiting for the load event would be better?

You could try `navigateTo('/test/cls?hidden=1', {readyState: 'complete'})` and see if that helps.

philipwalton (changes requested):

test/e2e/onCLS-test.js
tunetheweb: I was still seeing two shifts after this so hoped that might help.

However I changed:

```js
    assert.strictEqual(cls.entries.length, 1);
```

to:

```
    assert(cls.entries.length >= 1);
```

And that seemed to be the better fix anyway. Though don't think it should really be necessary.

Anyway cleaned this up in https://github.com/GoogleChrome/web-vitals/pull/460/commits/3c31c6872d7fe45c151b63c36d118b905f7fc276

test/e2e/onCLS-test.js
tunetheweb: Good idea, done!

tunetheweb: Actually let me see if that fixes the need for the `>=` mentioned above...

philipwalton (approved):

PR #468, opened 2024-04-25
Allow reportAllChanges for LCP for late loaded script

tunetheweb:
Currently if `reportAllChanges=true` but web-vitals is loaded late, then it doesn't report all LCP changes as we take the last one.

This can be confusing and make you think only one LCP entry was emitted when multiple ones were. It can be useful to know that multiple entries are emitted and this works for all the other metrics.

I also removed some "// Greater than the image load delay." comments that weren't true.

### DIFF
--- a/src/onLCP.ts
+++ b/src/onLCP.ts
@@ -58,6 +58,12 @@ export const onLCP = (onReport: LCPReportCallback, opts?: ReportOpts) => {
     let report: ReturnType<typeof bindReporter>;
 
     const handleEntries = (entries: LCPMetric['entries']) => {
+      // If reportAllChanges is set then call this function for each entry
+      // As otherwise only want to emit the last one.
+      if (opts!.reportAllChanges && entries.length > 1) {
+        entries.forEach((entry) => handleEntries([entry]));
+      }
+
       const lastEntry = entries[entries.length - 1] as LargestContentfulPaint;
       if (lastEntry) {
         // Only report if the page wasn't hidden prior to LCP.
--- a/src/onLCP.ts
+++ b/src/onLCP.ts
@@ -59,7 +59,7 @@ export const onLCP = (onReport: LCPReportCallback, opts?: ReportOpts) => {
 
     const handleEntries = (entries: LCPMetric['entries']) => {
       // If reportAllChanges is set then call this function for each entry,
-      // As otherwise only want to emit the last one.
+      // otherwise only consider the last one.
       if (!opts!.reportAllChanges) {
         entries = entries.slice(-1);
       }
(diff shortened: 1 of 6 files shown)


tunetheweb:
Not sure what's gling on with the Firefox test. Unrelated to this PR and seems it's using v127.0a1 instead of v125.0.1 (the last successful run 2 days ago) so a dependency upgrade and seems to be ahead of stable now. Will keep digging...

tunetheweb:
OK I think it's just a case of Firefox Nightly being a bit broken at the mo. Will try rerunning this tomorrow.

src/onLCP.ts
```diff
@@ -58,6 +58,12 @@ export const onLCP = (onReport: LCPReportCallback, opts?: ReportOpts) => {
     let report: ReturnType<typeof bindReporter>;
 
     const handleEntries = (entries: LCPMetric['entries']) => {
+      // If reportAllChanges is set then call this function for each entry
+      // As otherwise only want to emit the last one.
+      if (opts!.reportAllChanges && entries.length > 1) {
+        entries.forEach((entry) => handleEntries([entry]));
```
philipwalton, line 64: I think using a recursive call here make this code a bit hard to read.

Another way to do this would be to modify the `entries` array to ignore all but the last entry if `reportAllChanges` is false, and then do a `forEach()` over the logic below:

```
if (!opts.reportAllChanges) {
  entries = entries.slice(-1);
}
```

philipwalton (changes requested):

src/onLCP.ts
tunetheweb: Yeah I was trying to make the minimum number of calls so this, relatively rare, case had minimal impact on the more common not reportAllChanges case, but above is cleaner. Will do that.

tunetheweb: Done.

src/onLCP.ts
```diff
@@ -58,24 +58,26 @@ export const onLCP = (onReport: LCPReportCallback, opts?: ReportOpts) => {
     let report: ReturnType<typeof bindReporter>;
 
     const handleEntries = (entries: LCPMetric['entries']) => {
-      const lastEntry = entries[entries.length - 1] as LargestContentfulPaint;
-      if (lastEntry) {
+      // If reportAllChanges is set then call this function for each entry
```
philipwalton, line 61:
```suggestion
      // If reportAllChanges is set then call this function for each entry,
```

src/onLCP.ts
```diff
@@ -58,24 +58,26 @@ export const onLCP = (onReport: LCPReportCallback, opts?: ReportOpts) => {
...
+      // As otherwise only want to emit the last one.
```
philipwalton, line 62:
```suggestion
      // otherwise only consider the last one.
```

philipwalton (approved):

PR #464, opened 2024-04-13
Add jsdelivr and cdnjs as an alternative CDNs

tunetheweb:
Given the recent [unpkg.com outage](https://www.theverge.com/2024/4/12/24128276/open-source-unpkg-cdn-down) we should have backup CDNs in our README. See also #463 

Will merge with v4 branch and update after merged.

Issue #463, opened 2024-04-12
CDN JS - Web server is returning an unknown error


smallma:
CDN js got an error message: Web server is returning an unknown error.

https://unpkg.com/web-vitals@3/dist/web-vitals.iife.js

![image](https://github.com/GoogleChrome/web-vitals/assets/429250/988cfb47-cdec-4847-a86a-061158a7c00d)

tunetheweb:
This is the same for every package on unpkg so seems to be an issue at their end, and nothing that we can do unfortunately. The contact details for unpkg are: https://twitter.com/unpkg

tunetheweb:
Closing as seems to be back now.

smallma:
maybe you can provide a stable cdn link or js where to download for CDN user. Thanks.

PR #469, opened 2024-04-28
Alt alternative CDN options to README

tunetheweb:
Given the recent [unpkg.com outage](https://www.theverge.com/2024/4/12/24128276/open-source-unpkg-cdn-down) we should have backup CDNs in our README. See also https://github.com/GoogleChrome/web-vitals/issues/463

(v4 version of previous PR: #464)

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,9 +110,9 @@ See [Send attribution data](#send-attribution-data) for usage examples, and the 
 
 The recommended way to use the `web-vitals` package is to install it from npm and integrate it into your build process. However, if you're not using npm, it's still possible to use `web-vitals` by requesting it from a CDN that serves npm package files.
 
-The following examples show how to load `web-vitals` from [unpkg.com](https://unpkg.com):
+The following examples show how to load `web-vitals` from [unpkg.com](https://unpkg.com/browse/web-vitals/). It is also possible to load this from [jsDelivr](https://www.jsdelivr.com/package/npm/web-vitals), and [cdnjs](https://cdnjs.com/libraries/web-vitals).
 
-_**Important!** The [unpkg.com](https://unpkg.com) CDN is shown here for example purposes only. `unpkg.com` is not affiliated with Google, and there are no guarantees that the URLs shown in these examples will continue to work in the future._
+_**Important!** The [unpkg.com](https://unpkg.com), [jsDelivr](https://www.jsdelivr.com/), and [cdnjs](https://cdnjs.com) CDNs are shown here for example purposes only. `unpkg.com`, `jsDelivr`, and `cdnjs` are not affiliated with Google, and there are no guarantees that loading the library from those CDNs will continue to work in the future. Self-hosting the built files rather than loading from the CDN is better for security, reliability, and performance reasons._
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -110,9 +110,9 @@ See [Send attribution data](#send-attribution-data) for usage examples, and the
...
 
-The following examples show how to load `web-vitals` from [unpkg.com](https://unpkg.com):
+The following examples show how to load `web-vitals` from [unpkg.com](https://unpkg.com/browse/web-vitals/). It is also possible to load this from [jsDelivr](https://www.jsdelivr.com/package/npm/web-vitals), and [cdnjs](https://cdnjs.com/libraries/web-vitals).
 
-_**Important!** The [unpkg.com](https://unpkg.com) CDN is shown here for example purposes only. `unpkg.com` is not affiliated with Google, and there are no guarantees that the URLs shown in these examples will continue to work in the future._
+_**Important!** The [unpkg.com](https://unpkg.com), [jsDelivr](https://www.jsdelivr.com/), and [cdnjs](https://cdnjs.com) CDNs are shown here for example purposes only. `unpkg.com`, `jsDelivr`, and `cdnjs` are not affiliated with Google, and there are no guarantees that loading the library from those CDNs will continue to work in the future. Self-hosting the built files rather than loading from the CDN is better for security, reliability, and performance reasons._
```
philipwalton, line 115: Nit: while I agree this is generally true, in this particular case I'm not sure "performance" is a reason not to use a CDN to load this library.

philipwalton (approved):

README.md
tunetheweb: While it's true that since this library shouldn't be on the critical path, it can still cause performance issues if, for example, it is blocked from loading if the CDN is unresponsive or fails to load. I think, unless you use `async`, the subsequent scripts could hang and fail to run until that times out?

Plus the connection setup costs (e.g. to download HTTP certs) can take bandwidth better used elseswhere on low bandwidth connections. But think that's a minor point.

So to me there is no performance gain from using a CDN for this one library (since shared caching is not a thing anymore), and only a performance downside (even if small).

But could generalise to something like this if you prefer:

> In general, self-hosting libraries rather than loading from a JavaScript CDN is better for security, reliability, and performance reasons.

philipwalton: I'd keep the recommendation specific to this library, so I think what you have is fine.

> it can still cause performance issues if, for example, it is blocked from loading if the CDN is unresponsive or fails to load. I think, unless you use `async`, the subsequent scripts could hang and fail to run until that times out?

Right, I was referring to using the CDN with the code in the examples, which shouldn't have any real impact on performance.

The main reason I made the comment is that someone might open an issue asking what the performance impact of using a CDN is (given the warning in the README), and I'm not sure I'd have a good answer.

tunetheweb: Well I don’t think it needs to be specific to those simple examples. I think I’d be comfortable saying that subsequent scripts can be blocked. And point to the unpkg outage link in this PR or something like: https://csswizardry.com/2019/05/self-host-your-static-assets/

PR #471, opened 2024-05-01
Small type clean up for v4

brendankenny:
Relatively light type clean up for v4. Things are looking pretty good and I ended up not changing that much, which is great. For many use cases there won't be breaking changes here in practice (e.g. if you already had a `LCPMetric` specifically, the name of the base interface or a union it might belong to doesn't matter).

It's possible some things still went too far for some sensibilities; let me know and I can just roll those back. Once this is looking good, I'll update the README entries around these.

Changes:
- Switch `MetricType` for `Metric` (as originally in #359)

   Most of the time as a user you just want to know which of the union of possible metric types you have, you don't want the combined type in the old `Metric`. If you do somehow end up with a generic `Metric`, it's now a discriminated union, so checking e.g. `metric.name === 'LCP'` will narrow the type to `LCPMetric`. The old `Metric` is still available as `MetricBase` (open for name bike shedding)
- Ditch `CLSReportCallback`/`CLSReportCallbackWithAttribution`, etc. Maybe controversial, but IMO these end up as too many named types for things users will usually never explicitly type, and the only important part is the parameter type, so let's just use that directly.
- slight clean up of `unattributedOn*` methods

   e.g. `unattributedOnCLS` was supposedly calling back with a `metric` of type `CLSMetricWithAttribution`, which isn't true. `metric` is a `CLSMetric` and only becomes a `CLSMetricWithAttribution` after `attributeCLS` is called. Still a type assertion `as CLSMetricWithAttribution` in there, but it better reflects what's really going on. We could have the `attribute*` methods return the newly attributed object to avoid the extra type assertion, but I wasn't sure if that was too big a change.
- A few scattered removed unnecessary type assertions, e.g. `as PerformanceEventTiming` when the compiler already knows they're `PerformanceEventTiming`s


brendankenny:
> We could have the attribute* methods return the newly attributed object to avoid the extra type assertion, but I wasn't sure if that was too big a change.

I like this way better, actually, so I added it. Separate commit so it's easy to remove, though.

src/attribution/onFCP.ts
```diff
@@ -19,42 +19,40 @@ import {getLoadState} from '../lib/getLoadState.js';
...
-const attributeFCP = (metric: FCPMetric): void => {
   if (metric.entries.length) {
     const navigationEntry = getNavigationEntry();
     const fcpEntry = metric.entries[metric.entries.length - 1];
 
     if (navigationEntry) {
       const responseStart = navigationEntry.responseStart;
-      if (isInvalidTimestamp(responseStart)) return;
+      // TODO(bckenny): this is wrong.
+      if (isInvalidTimestamp(responseStart)) return metricWithAttribution;
```
brendankenny, line 34: I didn't notice this until making this change, but since `attribution` is required, isn't this a bug to not have any `attribution` defined in this case?

src/attribution/onLCP.ts
```diff
@@ -22,18 +22,19 @@ import {
...
-const attributeLCP = (metric: LCPMetric) => {
+const attributeLCP = (metric: LCPMetric): LCPMetricWithAttribution => {
+  const metricWithAttribution = metric as LCPMetricWithAttribution;
+
   if (metric.entries.length) {
     const navigationEntry = getNavigationEntry();
 
     if (navigationEntry) {
       const responseStart = navigationEntry.responseStart;
-      if (isInvalidTimestamp(responseStart)) return;
```
brendankenny, line 36: Similar in this case, `attribution` is required but none is added here

src/attribution/onFCP.ts
tunetheweb: That is interesting how did the cast work below if this is required?:

```js
(metric as FCPMetricWithAttribution).attribution = {
```

Surely when the initial case, and before `attribution` is set it's a `FCPMetricWithAttribution` type but without an `attribution`?

Just curious btw rather than saying you're not right here...

src/types/base.ts
```diff
@@ -60,7 +60,7 @@ export interface Metric {
    * The array may also be empty if the metric value was not based on any
    * entries (e.g. a CLS value of 0 given no layout shifts).
    */
-  entries: (PerformanceEntry | LayoutShift)[];
+  entries: PerformanceEntry[] | LayoutShift[];
```
philipwalton, line 63:
```suggestion
  entries: PerformanceEntry[];
```

Now that all of the other metric interfaces set their own `entires` type, I think this can be simplified.

src/types/base.ts
```diff
@@ -14,14 +14,14 @@
...
-import type {TTFBMetric} from './ttfb.js';
+import type {CLSMetric, CLSMetricWithAttribution} from './cls.js';
+import type {FCPMetric, FCPMetricWithAttribution} from './fcp.js';
+import type {FIDMetric, FIDMetricWithAttribution} from './fid.js';
+import type {INPMetric, INPMetricWithAttribution} from './inp.js';
+import type {LCPMetric, LCPMetricWithAttribution} from './lcp.js';
+import type {TTFBMetric, TTFBMetricWithAttribution} from './ttfb.js';
 
-export interface Metric {
+export interface MetricBase {
```
philipwalton, line 24: Given that this interface definition is used in the README documentation, my preference would be to keep the name the same.

What's the rationale for changing it?

philipwalton (changes requested):
Thanks for taking a look at this @brendankenny! For the most part these changes LGTM. The main thing missing right now is to copy the type changes to the relevant parts of the README documentation.

src/attribution/onFCP.ts
brendankenny: > Surely when the initial case, and before `attribution` is set it's a `FCPMetricWithAttribution` type but without an `attribution`?

Exactly. The cast/[type assertion](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions) overrides the type checker, so the compiler believes it's an `FCPMetricWithAttribution` and so will tell you `attribution` is defined, but it might not be. Type assertions are an escape hatch when you know something is correct (or it won't matter for this particular code) but the compiler can't tell for whatever reason. The downside is that the human half of that has to catch any mistakes because the compiler is coming into it with intentionally incorrect assumptions.

Typescript in general isn't great at typing objects that evolve in shape over time like this. Usually you'd do something like make `attribution` optional in type-land or you could create a new object (`{...metric, attribution}`) in value-land and it would track that, but that brings in JS we don't want to run in this case.

I think either we want optional `attribution` for these two cases (if there really is no attribution for them) or use the same kind of fallback empty attribution like is used a little farther down.

src/attribution/onFCP.ts
```diff
@@ -19,42 +19,40 @@ import {getLoadState} from '../lib/getLoadState.js';
...
-  FCPMetric,
-  FCPMetricWithAttribution,
-  FCPReportCallback,
-  FCPReportCallbackWithAttribution,
-  ReportOpts,
-} from '../types.js';
+import {FCPMetric, FCPMetricWithAttribution, ReportOpts} from '../types.js';
+
+const attributeFCP = (metric: FCPMetric): FCPMetricWithAttribution => {
+  const metricWithAttribution = metric as FCPMetricWithAttribution;
```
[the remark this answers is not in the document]
brendankenny: It does have the same problem, just keeps the type lies highly localized and so in theory easier to keep an eye on :)

Your proposal sounds good to make that even better. The early returns of a value (instead of undefined) were sometimes awkward for some of these anyways, so better all around.

src/attribution/onFCP.ts
brendankenny: > Ahhh, good catch

@philipwalton should these get the default `attribution` objects?

src/types/base.ts
brendankenny: > What's the rationale for changing it?

To make it clearer that this is a base interface used for the properties shared between the metrics, not a type you should be encountering or declaring a value to be. The union-of-metrics type is something you may encounter, though (I meant to rename s`MetricType`/`Metric` but I forgot).

My proposal was going to be remove it from the README and have the individual metrics documented instead with the base properties inlined into each. Writing that out now, it's probably a bit verbose.

Could maybe be documented like

```ts
interface CLSMetric {
  name: 'CLS';
  entries: LayoutShift[];
}

interface FCPMetric {
  name: 'FCP';
  entries: PerformancePaintTiming[];
}

// ...
```
Then something like "All metrics share the following properties:"
```ts
interface MetricBase { // (or whatever)
  /**
   * The current value of the metric.
   */
  value: number;

  // ...current `Metric` with `name` and `entries` removed
}
```

brendankenny: Actually, playing with it, I don't think the `Metric` or `MetricType` names are that important on their own, so I'm happy leaving them unchanged. I will try out that adjustment to the README, though.

src/attribution/onFCP.ts
philipwalton: I think they should be an empty object in these cases. I think it would be confusing if the attribution property itself was required on some metrics but optional on others.

brendankenny: > I think they should be an empty object in these cases. I think it would be confusing if the attribution property itself was required on some metrics but optional on others (and since in v3 it was required on all metrics, I think we should stick with that).

Yeah, that's what I was assuming.

src/attribution/onLCP.ts
```diff
@@ -22,19 +22,22 @@ import {
...
+    timeToFirstByte: 0,
+    resourceLoadDelay: 0,
+    resourceLoadDuration: 0,
+    elementRenderDelay: metric.value,
+  };
 
+  const navigationEntry = getNavigationEntry();
+  if (metric.entries.length && navigationEntry) {
+    const responseStart = navigationEntry.responseStart;
+    if (!isInvalidTimestamp(responseStart)) {
```
brendankenny, line 40: almost tempted to break to label here :) Moving the `getNavigationEntry` out seems ok since I assume the `!metric.entries.length` case is very rare.

src/attribution/onCLS.ts
```diff
@@ -33,26 +32,31 @@ const getLargestLayoutShiftSource = (sources: LayoutShiftAttribution[]) => {
...
         };
-        return;
       }
     }
   }
-  // Set an empty object if no other attribution has been set.
-  (metric as CLSMetricWithAttribution).attribution = {};
+
+  // Cast to attribution metric so it can be populated.
+  const metricWithAttribution = metric as CLSMetricWithAttribution;
```
philipwalton, line 57: What about doing something like this to avoid having to typecast?

```ts
const metricWithAttribution: CLSMetricWithAttribution = Object.assign(
  metric,
  {attribution},
);

return metricWithAttribution;
```

brendankenny:
> The main thing missing right now is to copy the type changes to the relevant parts of the README documentation

PTAL

src/attribution/onCLS.ts
brendankenny: Oh, did I misunderstand the need to maintain object identity for the metrics? e.g. so Set will work with multiple callbacks for a metric

src/attribution/onLCP.ts
brendankenny: > Now that we're adding `!` before `isInvalidTimestamp()` it reads awkwardly, and TBH I think it could be refactored into the `getNavigationEntry()` function since it's always used in that context.
> 
> Did you configure your fork to let me push to it in the PR? (Or could you just give me access?) I'd be happy to experiment with some options in this PR.

Makes sense. I just enabled pushes to this PR (also you should add me to `web-vitals` :)

src/attribution/onCLS.ts
philipwalton: This is maintaining object identity, no?

The main reason I was suggesting this is to appease the TypeScript gods and not use type assertions. I don't think it's functionally better than what you originally had.

src/attribution/onLCP.ts
philipwalton: Added! (I didn't realize you weren't already there.)

src/attribution/onCLS.ts
brendankenny: Ugh, yeah, I need to read less quickly (I guess I was anchored on the `Object.assign({}, ...)` spread replacement). Annoying we need to run that line but great to get rid of the type assertions altogether.

brendankenny: Updated!

philipwalton:
@brendankenny I made some updates to your updates (and merged in @tunetheweb's test fixes). Let me know if you have any concerns with any of the changes I made.

tunetheweb (approved):
LGTM with some README link suggestions

philipwalton:
Actually I ended up just removing all the links to the source types because it looks like they no longer work with GitHub's new file viewer UI.

philipwalton (approved):

Issue #472, opened 2024-05-02
Firefox e2e tests flakey


tunetheweb:
Looks in Firefox Nightly FCP is often quite slow in our e2e tests.

For example [this run](https://github.com/GoogleChrome/web-vitals/actions/runs/8923549678/job/24508034241) has this when I turn on debugging:

```
[0-1] reports the correct value after the first paint {
[0-1]   name: 'FCP',
[0-1]   value: 53[15](https://github.com/GoogleChrome/web-vitals/actions/runs/8923549678/job/24508034241#step:6:16),
[0-1]   rating: 'poor',
[0-1]   delta: 5315,
[0-1]   entries: [
[0-1]     {
[0-1]       name: 'first-contentful-paint',
[0-1]       entryType: 'paint',
[0-1]       startTime: 5315,
[0-1]       duration: 0
[0-1]     }
[0-1]   ],
[0-1]   id: 'v4-[17](https://github.com/GoogleChrome/web-vitals/actions/runs/8923549678/job/24508034241#step:6:18)14651853858-9340183650873',
[0-1]   navigationType: 'navigate'
[0-1] }

A rerun often solves it as you can see later in that log:

```
[0-1] reports the correct value after the first paint {
[0-1]   name: 'FCP',
[0-1]   value: 72,
[0-1]   rating: 'good',
[0-1]   delta: 72,
[0-1]   entries: [
[0-1]     {
[0-1]       name: 'first-contentful-paint',
[0-1]       entryType: 'paint',
[0-1]       startTime: 72,
[0-1]       duration: 0
[0-1]     }
[0-1]   ],
[0-1]   id: 'v4-171465[18](https://github.com/GoogleChrome/web-vitals/actions/runs/8923549678/job/24508034241#step:6:19)59617-3386553824299',
[0-1]   navigationType: 'navigate'
[0-1] }
```

But it doesn't seem to solve it for these tests which have been continually failing:
- FCP - 'reports if the page is restored from bfcache'
- LCP - 'reports the correct value on hidden (reportAllChanges === false)'

I hoped it was just some nightly flakiness but doesn't seem to be getting resolved so we might want to comment these out for now for Firefox.

I can't reproduce this on MacOS, nor on linux (Chromebox) so it only seems to happen in GitHub Actions. Possibly related to there not being a display there?

tunetheweb:
Raised a bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1894695

[...]

PR #473, opened 2024-05-02
Turn off flakey Firefox tests

tunetheweb:
Makes progress on #472 by commenting out the frequently failing tests.

### DIFF
--- a/test/e2e/onFCP-test.js
+++ b/test/e2e/onFCP-test.js
@@ -175,8 +175,7 @@ describe('onFCP()', async function () {
     assert.strictEqual(fcp1.value, fcp1.delta);
     // Temp fix to address Firefox flakiness.
     // See https://github.com/GoogleChrome/web-vitals/issues/472
-    console.log(':' + browser.browserName + ':');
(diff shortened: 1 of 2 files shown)


tunetheweb:
> Do you think it's better to comment out particular failures or just disable FF tests until this is fixed? My concern is it will be very easy to forget they've been commented out.

I did just comment out the asserts that failed (the rating checks). Annoyingly the LCP one that's failing is in the `assertStandardReportsAreCorrect` function used by a lot of the tests so it lost that assert for all of them but I preferred that to pulling that code out for that one test, or disabling the whole test.

I still think that's better to keep most of the Firefox tests working. I plan on leaving #472 open until this is fixed by Firefox as a reminder (hence why it's "makes progress on" rather than "fixes" in the initial comment which would have closed it).

Ultimately I'm hoping this is a short term issue and will be fixed on next version, but getting bored of seeing the failures and worried we will miss a REAL failure because of this.

philipwalton:
Gotcha. Keeping the issue open as a reminder to address it eventually SGTM.

philipwalton (approved):

Issue #335, opened 2023-03-29
Target in FID and INP is null


markusBurda:
Hi guys,

Im using your nice script to measuer web vitals on our website. To make optimization easier, we are also persisting the target-elemtent. It works fine with CLS and LCP but in FID and INP I get "null" for the target Element as you can see in this screenshot:
![image](https://user-images.githubusercontent.com/18657665/228434467-45bf4e16-4181-41d1-9938-4634655cf966.png)

This is how I use it:
`
        <script type="module">
            import {onCLS, onFID, onLCP, onINP, onFCP, onTTFB} from "https://unpkg.com/web-vitals@3/dist/web-vitals.attribution.js?module";

            function getIdOrNodeNamePlusClassList(node) {
                return (node && node.nodeType !== 9) ? (node.id ?
                    "#" + node.id :
                    node.nodeName.toLowerCase() +
                    ((node.className && node.className.length) ?
                        "." + Array.from(node.classList.values()).join(".") :
                        "")) : "";
            }

            function getLargestLayoutShiftEntry(entries) {
              return entries.reduce((a, b) => a && a.value > b.value ? a : b);
            }

            function getLargestLayoutShiftSource(sources) {
              return sources.reduce((a, b) => {
                return a.node && a.previousRect.width * a.previousRect.height >
                    b.previousRect.width * b.previousRect.height ? a : b;
              });
            }

            function getCausingElementIdentifier(name, entries = []) {
                // In some cases there won't be any entries (e.g. if CLS is 0,
                // or for LCP after a bfcache restore), so we have to check first.
                if (entries.length && name !== "FCP" && name !== "TTFB") {
                    if (name === "LCP") {
                        const lastEntry = entries[entries.length - 1];
                        return getIdOrNodeNamePlusClassList(lastEntry.element);
                    } else if (name === "FID" || name === "INP") {
                        const firstEntry = entries[0];
                        return getIdOrNodeNamePlusClassList(firstEntry.target);
                    } else if (name === "CLS") {
                        const largestEntry = getLargestLayoutShiftEntry(entries);
                        if (largestEntry && largestEntry.sources && largestEntry.sources.length) {
                            const largestSource = getLargestLayoutShiftSource(largestEntry.sources);
                            if (largestSource) {
                                return getIdOrNodeNamePlusClassList(largestSource.node);
                            }
                        }
                    }
                }
                return "";
            }

            function sendRUMData({name, value, attribution, entries}) {
[...]

PR #477, opened 2024-05-09
Save INP target early for when removed from DOM

### LINKED ISSUE #456: Expose INP target element as first class citizen in attribution
### Background

We use the INP target element instance for gaining additional insights on it upon `onINP` callback.  
The info we're getting is mostly the element location in DOM tree and its relation to the parent elements, which, in turn, helps us understand the element's attribution to specific application/widget within the website.  

### The issue

As of `web-vitals@4.0.0-beta.1` we have to get the target by accessing `metric.entries` and choosing the right `entry.target`. For this to be aligned with what `web-vitals` expose as INP target selector we have to use the [same logic](https://github.com/GoogleChrome/web-vitals/blob/c963a5723a78bd7eec51bb9e50283aafc23b3388/src/attribution/onINP.ts#L211-L215) as in `web-vitals`. Today it takes the first entry with a non-null target, because of a [Chrome bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1367329). Once the bug is solved the logic for target element selection will probably change to get the first entry (`metric.entries[0].target`) and then we'll have to change it in our code as well. 
If fact every time this logic changes (even though unlikely), we'll have to update our code as well, which makes it a leaky abstraction. 

### Proposed solution

Expose INP target *element* as a first class citizen in INP attribution.  
This way the alignment between INP target selector and INP target will be abstracted away from the user and if/when the logic changes it will only have to change inside `web-vitals`.

tunetheweb:
Makes progress on #335 
Fixes #456 

When a target element is removed from the DOM it cannot be looked up to get the CSS selector.

There are instances when the `first-input` and `pointer-down` events provide a target element, but the later `pointerup` and `click` events do not as the element has been removed. This is even more apparent in v4 where [we wait to attribute INP until idle](https://github.com/GoogleChrome/web-vitals/blob/157d1a80d3140896f0f84211ed818d7c2fb84a3e/src/attribution/onINP.ts#L288).

This PR saves a target reference for each event with an interaction id, so it can be retrieved later if the target is then null.

This will not help if the element is removed before the first event entry from that interactionId is reported so is not a full solution, but will help in some cases.

### DIFF
--- a/test/e2e/onINP-test.js
+++ b/test/e2e/onINP-test.js
@@ -651,7 +651,7 @@ describe('onINP()', async function () {
       assert.equal(inp1.attribution.interactionTarget, 'html>body>main>h1');
     });
 
-    it('reports the interaction target when entry removed after event callback', async function () {
+    it('reports the interaction target when target is removed from the DOM', async function () {
       if (!browserSupportsINP) this.skip();
 
       await navigateTo('/test/inp?attribution=1&mouseup=100&click=50', {
@@ -661,8 +661,9 @@ describe('onINP()', async function () {
       const button = await $('#reset');
       await simulateUserLikeClick(button);
 
-      // Wait a bit to ensure events are reported
-      await browser.pause(1000);
+      await nextFrame();
+
+      // Remove the element after the interaction.
       await browser.execute('document.querySelector("#reset").remove()');
 
       await stubVisibilityChange('hidden');
@@ -672,7 +673,6 @@ describe('onINP()', async function () {
 
       assert.equal(inp.attribution.interactionType, 'pointer');
       assert.equal(inp.attribution.interactionTarget, '#reset');
-      assert(inp.attribution.interactionTargetElement);
     });
 
     it('includes LoAF entries if the browser supports it', async function () {
--- a/test/e2e/onINP-test.js
+++ b/test/e2e/onINP-test.js
@@ -651,6 +651,29 @@ describe('onINP()', async function () {
       assert.equal(inp1.attribution.interactionTarget, 'html>body>main>h1');
     });
 
+    it('reports the interaction target when entry removed after event callback', async function () {
+      if (!browserSupportsINP) this.skip();
+
+      await navigateTo('/test/inp?attribution=1&mouseup=100&click=50', {
+        readyState: 'interactive',
+      });
+
+      const button = await $('#reset');
+      await simulateUserLikeClick(button);
+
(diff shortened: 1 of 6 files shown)


nucliweb (approved):

mmocny (approved):
Thanks for this change so quick.

gilbertococchi:
Thanks for fixing this Barry!
Lots of Target nulls due to CMP UI Click should go away with this change!

tunetheweb:
Don't get too excited just yeet. I'm having dropping making this work in the general case. Not sure how it works on your site to be honest :-(

gilbertococchi:
> Don't get too excited just yet. I'm having dropping making this work in the general case. Not sure how it works on your site to be honest :-(

Thanks Barry, you know me well :)

I just did test out this change on few CMP scenarios and it seems work correctly, happy to test out more scenarios as well today.

tunetheweb:
@philipwalton with the latest change to keep the element, I think it makes sense to expose this as requested in #456 so I added that in https://github.com/GoogleChrome/web-vitals/pull/477/commits/663461df255f0a17f91cba26626e3d0654fb4f18. Technically another change to the API but as it's an addition I don't think we need another release candidate for this. WDYT?

philipwalton (approved):
@tunetheweb LGTM with a few changes from me (that we discussed offline). Feel free to merge if you're happy with my changes.

test/e2e/onINP-test.js
```diff
@@ -651,6 +651,30 @@ describe('onINP()', async function () {
...
+      if (!browserSupportsINP) this.skip();
+
+      await navigateTo('/test/inp?attribution=1&mouseup=100&click=50', {
+        readyState: 'interactive',
+      });
+
+      const button = await $('#reset');
+      await simulateUserLikeClick(button);
+
+      await nextFrame();
```
tunetheweb, line 664: Can we be sure that event timing will have run by the next frame?

philipwalton: I don't think we can be sure, but it worked 100% of the time in my testing, and there are 2 retries if it fails.

My preference would be to use something like this rather than an arbitrary timeout. The most robust solution would be to add a PerformanceObserver and wait for the entries to dispatch, but I think what we have now is sufficient (and we can update it if we notice flakiness).

tunetheweb: Fair enough. It's only a test so our problem if it doesn't work.

PR #481, opened 2024-05-13
Only apply Firefox workarounds in GitHub Actions

tunetheweb:


test/e2e/onLCP-test.js
```diff
@@ -717,7 +718,10 @@ const assertFullReportsAreCorrect = (beacons) => {
 
   // Temp fix to address Firefox flakiness.
   // See https://github.com/GoogleChrome/web-vitals/issues/472
-  if (browser.capabilities.browserName !== 'firefox') {
+  if (
```
philipwalton, line 721: I think this logic is incorrect.

Right now I think it will only run the assertion locally in Firefox.

philipwalton (changes requested):

test/e2e/onLCP-test.js
tunetheweb: Fixed

philipwalton (approved):

PR #487, opened 2024-05-23
INP attribution: allow for LoAFs observed before interaction events

### LINKED ISSUE #485: LoAF missed for some INP events
I ran across this issue while playing with the demo for the [Measuring INP codelab](https://codelabs.developers.google.com/measuring-inp). After trying to find a minimal repro, it still depends on a number of _very_ specific circumstances, but
- I ran into it while doing typical INP debugging of a fairly normal page and
- it's possible my repro is only locally minimal and there are more common ways to induce the behavior

so it still seems worth fixing.

### tl;dr
LoAFs can sometimes arrive from the PerfObserver before the interaction events they contain. If a `requestIdleCallback` fires between the PerfObserver callbacks, a LoAF can be discarded and won't be able to be paired with the interaction events when they come in later.

From out-of-band discussion, a reasonable fix would be to not discard any LoAFs that have a `startTime` _after_ any existing event's `processingEnd` (or `startTime` + `duration`?). I'll open a PR to address this.

---

This turned out to be relatively simple, but I spent a decent amount of time debugging because it didn't appear to be simple at first, so I'm going to write down way too much about it for posterity. Feel free to skip the rest :)

### To reproduce
- Visit https://nosy-first-plaster.glitch.me/
- As mentioned on the page, enable mobile device emulation and 4x CPU slowdown
- Click on the hamburger button

The demo does two things:
- when the hamburger button is clicked it gets a `:hover` style (since this is emulated mobile) and the hamburger SVG stroke starts a CSS color transition
- the `click` event triggers a 400ms busy task in a `requestAnimationFrame` callback

Things that are necessary but I don't totally understand:
- the hamburger button needs `touch-action: manipulation` (originally included because I had copy/pasted some Adam Argyle styles into the page). I thought that was just useful for preventing DTZ, but the page also has a proper mobile viewport set so that shouldn't be a factor regardless. Does that property interact with mobile `:hover` in some way? Speaking to @mmocny, setting this may affect the timing/order of event dispatch, so it may just be a matter of PerfObserver timing, not a difference in behavior
- 4x CPU slowdown appears necessary. I assume because it affects the `requestIdleCallback` underlying `whenIdle` triggering LoAF cleanup, but it's unclear if it just makes it easier to repro and isn't necessary, or if there's other task ordering/triggering that could have the same effect without DevTools.

### Behavior
Usually two INP events will be logged to console (`reportAllChanges` is set to `true`). On my machine:
- Sometimes there will be two identical INPs logged.

   What's actually happening is two _different_ INPs are reported from the unattributed `onINP`, but the `whenIdle` on the callback in the attributed `onINP` is causing the first report to not be made (and logged to console) until after the `metric` has been updated for the second report callback. Since it's the same object being updated and passed to the two callbacks, the same values are logged twice. In this case the attribution will have a LoAF, though, so everything is good (I don't know if it's worth doing anything about the slightly confusing apparent double logging...maybe debounce when `reportAllChanges` is true? Abort the idle task when a new report comes in? Or is that not "report _all_ changes" anymore?)

- Much more often, a short INP is logged with just a `pointerdown` event in `entries`, then a little later a longer INP comes in with the ~400ms `click` event. **Neither of these INP values will have an attributed LoAF attached**.

I can very reliably reproduce the second case, which is why I noticed this in the first place.

Looking into what's happening, it appears that
- the `pointerdown` comes in
- a `LoAF` comes in that starts _after_ the `pointerdown`
- [`queueCleanup`'s `whenIdle` fires](https://github.com/GoogleChrome/web-vitals/blob/b32786bf6baa82593a4cae12226f8e81172b8407/src/attribution/onINP.ts#L148-L153). The LoAF [doesn't overlap with the `pointerdown`, so is discarded](https://github.com/GoogleChrome/web-vitals/blob/b32786bf6baa82593a4cae12226f8e81172b8407/src/attribution/onINP.ts#L184-L194)
- the `click` event comes in, which would have overlapped with the `LoAF`
- With no overlapping `LoAF` for attribution now, `longAnimationFrameEntries` is left empty.

<details>
<summary><strong>Example run of bug</strong></summary>

This can be difficult to debug because recording via the performance panel and even console logging seem to affect when the `whenIdle` callbacks fire. I ended up using a global array to push log messages to and then logged everything at once at the end.

Here's a trace with the different PerformanceObserver callbacks added as performance marks and the LoAF as a measure:
annotated DevTools trace showing order of bug
https://trace.cafe/t/Q3usrDo9C7

Console output:
- new events (at 1930.9):
   | name | startTime | processingStart | processingEnd | end | duration
   | -- | -- | -- | -- | -- | --
   | `'pointerdown'` | 1813.5 | 1834.6 | 1834.6 | 1837.5 | 24

- new LoAFs (at 2354.3):
   | startTime | end | duration
   | -- | -- | --
   | 1935.3 | 2347.3 | 412

- LoAFs discarded (at 2356.9):
   | startTime | end | duration
   | -- | -- | --
   | 1935.3 | 2347.3 | 412

- web-vitals: INP of 24ms

- new events (at 2371.8):
   | name | startTime | processingStart | processingEnd | end | duration
   | -- | -- | -- | -- | -- | --
   | `'mouseover'` | 1909.4 | 1941.1 | 1941.1 | 2357.4 | 448
   | `'mousedown'` | 1909.4 | 1942.1 | 1942.1 | 2357.4 | 448
   | `'mouseup'` | 1909.4 | 1943 | 1943 | 2357.4 | 448
   | `'click'` | 1909.4 | 1943 | 1943.9 | 2357.4 | 448

- web-vitals: INP of 448ms
</details>

brendankenny:
fixes #485

As described there, I added an extra condition to keep any LoAFs around that are _after_ any existing known interactions, in case those interactions then come in later.

I then realized that would leave the LoAF collection unbounded, since a page can have a bunch of slow animation frames without the user interacting. I added a `MAX_UNPAIRED_LOAFS` of 20, so it will only keep the 20 most recent LoAFs after any known user interaction events. The number isn't hugely important, that just allows a one second delay in event-timing reporting even if the page somehow has filled the time with the shortest (50ms) long animation frames possible. LoAF entries are fairly light weight, so that's not so bad.

As I was implementing this, however, I realized that even without this change the `pendingLoAFs` array can grow without bound on a page with e.g. animations and a busy main thread, since LoAFs are appended as they come in but clean up is only scheduled by user interactions. Seems like we should fix? @philipwalton @tunetheweb 

My only idea right now is to split the event timing and LoAF cleanups into two separate functions, and run each as their respective entries come in, since that's the only time each list of entries grows.

### DIFF
--- a/src/attribution/onINP.ts
+++ b/src/attribution/onINP.ts
@@ -42,7 +42,7 @@ interface pendingEntriesGroup {
  * The maximum number of LoAFs to retain on cleanup that don't have any
  * overlapping events.
  */
-const MAX_UNPAIRED_LOAFS = 20;
+const MAX_PENDING_ENTRIES = 50;
 
 // A PerformanceObserver, observing new `long-animation-frame` entries.
 // If this variable is defined it means the browser supports LoAF.
@@ -65,6 +65,9 @@ const pendingEntriesGroupMap: Map<number, pendingEntriesGroup> = new Map();
 // are removed.
 let previousRenderTimes: number[] = [];
 
+// The `processingEnd` time of most recently-processed event, chronologically.
+let latestProcessingEnd: number;
+
 // A WeakMap so you can look up the `renderTime` of a given entry and the
 // value returned will be the same value used by `pendingEntriesGroupMap`.
 const entryToRenderTimeMap: WeakMap<
@@ -84,7 +87,8 @@ let idleHandle: number = -1;
  * Adds new LoAF entries to the `pendingLoAFs` list.
  */
 const handleLoAFEntries = (entries: PerformanceLongAnimationFrameTiming[]) => {
-  entries.forEach((entry) => pendingLoAFs.push(entry));
+  pendingLoAFs = pendingLoAFs.concat(entries);
+  queueCleanup();
 };
 
 // Get a reference to the interaction target element in case it's removed
@@ -110,6 +114,8 @@ const saveInteractionTarget = (entry: PerformanceEventTiming) => {
 const groupEntriesByRenderTime = (entry: PerformanceEventTiming) => {
   let renderTime = entry.startTime + entry.duration;
   let previousRenderTime;
+
+  latestProcessingEnd = Math.max(latestProcessingEnd, entry.processingEnd);
 
   // Iterate of all previous render times in reverse order to find a match.
   // Go in reverse since the most likely match will be at the end.
@@ -149,6 +155,8 @@ const groupEntriesByRenderTime = (entry: PerformanceEventTiming) => {
   if (entry.interactionId || entry.entryType === 'first-input') {
     entryToRenderTimeMap.set(entry, renderTime);
   }
+
+  queueCleanup();
 };
 
 const queueCleanup = () => {
@@ -173,7 +181,7 @@ const cleanupEntries = () => {
   // events are dispatched out of order. When this happens they're generally
   // only off by a frame or two, so keeping the most recent 50 should be
   // more than sufficient.
-  previousRenderTimes = previousRenderTimes.slice(-50);
+  previousRenderTimes = previousRenderTimes.slice(-1 * MAX_PENDING_ENTRIES);
 
   // Keep all render times that are part of a pending INP candidate or
   // that occurred within the 50 most recently-dispatched animation frames.
@@ -187,10 +195,9 @@ const cleanupEntries = () => {
     if (!renderTimesToKeep.has(key)) pendingEntriesGroupMap.delete(key);
   });
 
-  // Keep all pending LoAF entries that
-  // - intersect with entries in the newly cleaned up `pendingEntriesGroupMap`
-  // - occur after all existing pendingEntriesGroups, in case a LoAF came in
-  //   before concurrent entries.
+  // Keep all pending LoAF entries that either:
+  // 1) intersect with entries in the newly cleaned up `pendingEntriesGroupMap`
+  // 2) occur after the most recently-processed event entry.
   const loafsToKeep: Set<PerformanceLongAnimationFrameTiming> = new Set();
   pendingEntriesGroupMap.forEach((group) => {
     getIntersectingLoAFs(group.startTime, group.processingEnd).forEach(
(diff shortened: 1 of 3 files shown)


brendankenny:
Oh, also I could not get a test for this working reliably. Inherently flaky and subject to future Chrome scheduler changes anyways. I verified the issue repro no longer fails, but looking for other ideas on this. Time to bust out some unit tests?

tunetheweb:
> As I was implementing this, however, I realized that even without this change the `pendingLoAFs` array can grow without bound on a page with e.g. animations and a busy main thread, since LoAFs are appended as they come in but clean up is only scheduled by user interactions. Seems like we should fix? @philipwalton @tunetheweb

Maybe we should do this clean up in `handleLoAFEntries` then (behind a `whenIdle` call)?
At the moment we queue a clean up interactions when new interactions come in, so we should also queue LoAFs clean ups when they come in.

philipwalton (approved):
Thanks for taking a look at this @brendankenny, and for the detailed repro. It was very helpful for testing this!

I took Barry's suggestion and made a few other minor tweaks to the code, but other than that LGTM.

src/attribution/onINP.ts
```diff
@@ -38,6 +38,12 @@ interface pendingEntriesGroup {
   entries: PerformanceEventTiming[];
 }
 
+/**
+ * The maximum number of LoAFs to retain on cleanup that don't have any
+ * overlapping events.
```
tunetheweb, line 43: Should we explain the rational for picking 50? We explain it below but maybe better pulling that comment up here?

Also this is not just "The maximum number of LoAFs", it's also the maximum number of Rendering times so maybe change this comment a little because of that?

Alternatively, mqake this just for LoAFs and add `LOAF` to the const name

src/attribution/onINP.ts
```diff
@@ -167,7 +181,7 @@ const cleanupEntries = () => {
   // events are dispatched out of order. When this happens they're generally
   // only off by a frame or two, so keeping the most recent 50 should be
   // more than sufficient.
```
tunetheweb, line 183: This is the comment I'm suggest now belongs with the MAX_PENDING_ENTRIES definition.

tunetheweb (changes requested):
LGTM but some nits on comments.

philipwalton:
Updated, and also moved the `whenIdle()` call to within the main `onINP()` function (as mentioned [here](https://github.com/GoogleChrome/web-vitals/issues/485#issuecomment-2153166655)) to ensure that attribution is always processed synchronously with the INP dispatch.

tunetheweb (approved):

PR #488, opened 2024-05-27
Fix LICENSE (2020 -> 2024)

kurnakovv:


tunetheweb:
The year is when copyright is enforced from (i.e. the year or first publishing).

PR #493, opened 2024-06-10
fix cleanup loop over pending LoAFs

brendankenny:
introduced in d41f3bc in #487


philipwalton (approved):
(🤦 ... I could have sworn I tested this!)

brendankenny:
Well it works for the first LoAF, and I think that's all we've ever observed it happen to :)

PR #495, opened 2024-06-11
Small simplification refactor of INP attribution

brendankenny:
I wanted to do this while the code was still fresh in my mind. There are no expected functional changes here, just code simplification.

Currently, `previousRenderTimes` is used for easier lookup into event-timing-entry groups in `pendingEntriesGroupMap` and for quick event grouping by render time, but it's also not quite the same set of render times in `pendingEntriesGroupMap` (the times and groups grow together as events come in, but the map actually contains all the events referenceable by `previousRenderTimes` _and_ groups with events from `longestInteractionList`). The results in more bookkeeping and (speaking as someone recently getting their head around this code :) it can make some of this logic harder to follow because effects aren't always collocated.

This PR proposes getting rid of `previousRenderTimes` so that there are only two main data structures, an array tracking LoAFs and an array tracking groups of event timing entries. I think this makes it a little simpler to understand how events are grouped as they come in and which event groups are being preserved during cleanup, without a runtime or clarity tradeoff elsewhere.

`latestProcessingEnd` and the entry WeakMap are still here as secondary data sources because they're convenient ways to skip extra work and they're both only ever updated in a single place (the WeakMap now maps directly to entry groups, though).

### DIFF
--- a/src/attribution/onINP.ts
+++ b/src/attribution/onINP.ts
@@ -35,12 +35,13 @@ interface pendingEntriesGroup {
   startTime: DOMHighResTimeStamp;
   processingStart: DOMHighResTimeStamp;
   processingEnd: DOMHighResTimeStamp;
+  renderTime: DOMHighResTimeStamp;
   entries: PerformanceEventTiming[];
 }
 
-// The maximum number of previous frames for which data is kept in regards to.
+// The maximum number of previous frames for which data is kept.
 // Storing data about previous frames is necessary to handle cases where event
-// and LoAF entries are dispatched out or order, and so a buffer of previous
+// and LoAF entries are dispatched out of order, and so a buffer of previous
 // frame data is needed to determine various bits of INP attribution once all
 // the frame-related data has come in.
 // In most cases this out-of-order data is only off by a frame or two, so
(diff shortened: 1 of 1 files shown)


src/attribution/onINP.ts
```diff
@@ -35,12 +35,13 @@ interface pendingEntriesGroup {
   startTime: DOMHighResTimeStamp;
   processingStart: DOMHighResTimeStamp;
   processingEnd: DOMHighResTimeStamp;
+  renderTime: DOMHighResTimeStamp;
```
brendankenny, line 38: technically the code below could use `group.entries[0].startTime + group.entries[0].duration` and not add a new property here, but it seemed better to put this here as a convenience

src/attribution/onINP.ts
```diff
@@ -109,54 +104,55 @@ const saveInteractionTarget = (entry: PerformanceEventTiming) => {
...
+    // assume they were part of the same frame and add it to the group.
+    if (Math.abs(renderTime - potentialGroup.renderTime) <= 8) {
+      group = potentialGroup;
       group.startTime = Math.min(entry.startTime, group.startTime);
       group.processingStart = Math.min(
         entry.processingStart,
         group.processingStart,
       );
       group.processingEnd = Math.max(entry.processingEnd, group.processingEnd);
+      group.renderTime = Math.max(group.renderTime, renderTime);
```
brendankenny, line 133: `previousRenderTimes` was never updated before like this, it was just the renderTime of whichever event was dispatched first, but it felt weird to not do this 🤷

brendankenny: Or is it important that it's the first event's render time so groups are all entries with renderTime within 8ms of the first entry, not entries where they're all within 8ms of some other entry in that group? I can't really imagine a scenario where that could be a problem, but maybe if it's on a 144Hz monitor...

brendankenny: Talking to @mmocny this could indeed (theoretically!) be problematic, so dropped it and maintaining the current behavior of using the renderTime of the first dispatched event in a group

src/attribution/onINP.ts
philipwalton: Yeah, new property is better I think.

src/attribution/onINP.ts
```diff
@@ -180,35 +175,29 @@ const cleanupEntries = () => {
...
+    return longestInteractionGroups.includes(group);
   });
 
   // Keep all pending LoAF entries that either:
-  // 1) intersect with entries in the newly cleaned up `pendingEntriesGroupMap`
+  // 1) intersect with entries in the newly cleaned up `pendingEntriesGroups`
   // 2) occur after the most recently-processed event entry.
   const loafsToKeep: Set<PerformanceLongAnimationFrameTiming> = new Set();
-  pendingEntriesGroupMap.forEach((group) => {
+  for (let i = 0; i < pendingEntriesGroups.length; i++) {
```
philipwalton, line 193: Didn't like a `forEach()` right above a regular for loop, eh? :)

You could alternatively do something like this to keep it one line (and which is common in the Google JS style guide):

```
for (let i = 0, group; group = pendingEntriesGroups[i]; i++) {
```

philipwalton (approved):
Thanks @brendankenny! Overall these changes LGTM andI think do simplify things.

src/attribution/onINP.ts
philipwalton: Yeah, I agree dropping it is the right call.

PR #505, opened 2024-07-17
Fix interaction count after bfcache restore

BenoitZugmeyer:
While reviewing the INP implementation, I noticed what I believe is a bug in INP/`interactions.ts`. Currently, `prevInteractionCount` is always `0`, but based on the comment
```js
// Used to store the interaction count after a bfcache restore, since p98
// interaction latencies should only consider the current navigation.
```
I believe `prevInteractionCount` should be set to `getInteractionCount()` when reset, so the next time `getInteractionCountForNavigation()` is called it should return `0` instead of the count of interactions that happened before the bfcache restore.

### DIFF
--- a/src/lib/interactions.ts
+++ b/src/lib/interactions.ts
@@ -51,7 +51,7 @@ const getInteractionCountForNavigation = () => {
 };
 
 export const resetInteractions = () => {
-  prevInteractionCount = 0;
+  prevInteractionCount = getInteractionCount();
   longestInteractionList.length = 0;
   longestInteractionMap.clear();
 };


philipwalton:
Thanks @BenoitZugmeyer, I believe you're right, and that matches how the code looked in v3.5.2: https://github.com/GoogleChrome/web-vitals/blob/v3.5.2/src/onINP.ts#L247-L249

Looks like this got changed in the [refactor for v4](https://github.com/GoogleChrome/web-vitals/pull/442/files)...

philipwalton:
This has been released in `v4.2.2`.

Issue #486, opened 2024-05-23
Facing issue while exporting `web-vitals/attribution` using  `browserify`


Archish27:
Unable to build using `browserify` when using

```js
const {
  onCLS,
  onLCP,
  onFID,
  onINP,
  onFCP,
  onTTFB,
} = require("web-vitals/attribution");
```

Getting error while building browserify -
```
./node_modules/.bin/browserify -d  -t [ babelify --presets [ @babel/preset-env ] ] index.js > dist/bundle.js

/home/archish/code/agent-browser/node_modules/web-vitals/attribution.js:18
export * from './dist/web-vitals.attribution.js';
^
ParseError: 'import' and 'export' may appear only with 'sourceType: module'
```

tunetheweb:
There's some suggestions here which seem to work from a quick test: https://stackoverflow.com/questions/51023231/npm-browserify-import-and-export-may-appear-only-with-sourcetype-module.

Alternatively you can use `} = require("web-vitals/dist/web-vitals.attribution.iife.js");` or `} = require("web-vitals/dist/web-vitals.attribution.umd.cjs");` in your code.

But ultimately this is a browserify issue and not a web-vitals issue as far as I can tell so closing this.

PR #475, opened 2024-05-03
Disable more Firefox tests

tunetheweb:
Thought it was just a few tests that failed but apparently all of them are susceptible so let's just comment all of them out.

PR #507, opened 2024-08-01
Improve test flakiness

tunetheweb:
Makes the following improvements:
- Reverts temp workarounds from #473, #475, #476, #486 as no longer required as  discussed in: https://bugzilla.mozilla.org/show_bug.cgi?id=1893921
- Skips prerender on unsupported browser (seemed to be causing flakiness for FCP tests on Safari).
- Adds an extra delay until `interactive` for the LCP image after hidden test as LCP doesn't always seem to be reported without this in all browsers.
- Adds an extra delay for the FCP initial hidden tests as FCP doesn't always seem to be reported without this in Safari.
- Adds extra config to skip the Search engine choice pop up.


test/e2e/onLCP-test.js
```diff
@@ -282,6 +264,9 @@ describe('onLCP()', async function () {
     // Wait for a frame to be painted.
     await browser.executeAsync((done) => requestAnimationFrame(done));
 
+    // Wait a bit to ensure beacon gets a chance to be sent
+    await browser.pause(1000);
```
philipwalton, line 268: Can you explain why this is needed? (At minimum it seems like the comment is incorrect, since there's already code to wait for the beacon below.)

tunetheweb: No I can't! Still digging but was gonna ask if you have any ideas, cause this has tests has definitely been flakey with no beacon being sent, and this pause seems to fix it.

Was wonder if the move to only reporting LCP `whenIdle` made this more flakey on CI. And if there's a better way. Can update comment to that if think it is that.

tunetheweb: Maybe instead I could change line above from this:

```js
await navigateTo('/test/lcp?imgDelay=0&imgHidden=1');
```
to this:
```js
await navigateTo('/test/lcp?imgDelay=0&imgHidden=1', {readyState: 'interactive'});
```
?

philipwalton: Yeah, you could try the latter to see if that fixes things. I believe the code to wait for a frame to paint is meant to ensure the page is visible before hiding it, but it's possible FF fires rAF before that?

tunetheweb: It was actually Chrome that was flakey with this one. This PR has expended beyond it's initial scope and has since been renamed!

tunetheweb: That seems to have worked and no downside to that I don't think, and seems better. So let's go with that.

test/e2e/onFCP-test.js
```diff
@@ -136,6 +122,10 @@ describe('onFCP()', async function () {
 
     await navigateTo('/test/fcp?hidden=1', {readyState: 'interactive'});
 
+    // Wait a bit to ensure an initial "frame" happens to avoid Safari
+    // flakiness with this test if switching visibility as soon as interactive.
+    await browser.pause(100);
+
```
tunetheweb, line 128: This is another one you might have issues with @philipwalton , but seems to help.

wdio.conf.cjs
```diff
@@ -82,6 +82,7 @@ module.exports.config = {
     if (browserName === 'chrome') {
       capability['goog:chromeOptions'] = {
         excludeSwitches: ['enable-automation'],
+        args: ['disable-search-engine-choice-screen'],
```
tunetheweb, line 85: This is included in puppeteer 21.2.1 so could skip this and wait for that to come through. It doesn't cause issues, other than being annoying when testing locally.

test/e2e/onFCP-test.js
tunetheweb: Checked and even with `complete` Safari doesn't always finish loading web-vitals.js before changing the readyState. So the hardcoded pause seems to be the only thing that makes it more stable. Updated comment to reflect this.

philipwalton:
I reverted the checks that would skip some tests in browsers that don't support prerender. I don't think the prerender-stubbing code was the cause of the flakiness, I think it was mainly the issue we discovered with the `navigateTo()` method.

IMO it only makes sense to skip the check if we're *actually* triggering a prerender load somehow. If we're stubbing it then I think running the tests in all browsers is still useful to check that out assumptions of how things work are interoperable.

philipwalton:
I've run the tests three times in a row now and they've all passed each time.

PR #512, opened 2024-08-06
Fix missing LoAF entries in INP attribution

### LINKED ISSUE #510: INP attribution helper latestProcessingEnd never initialized
[`latestProcessingEnd`](https://github.com/GoogleChrome/web-vitals/blob/5b91b766031c435e44dda912be75f685d77fe073/src/attribution/onINP.ts#L66) is declared without a value, so it's initially undefined.

When [comparing](https://github.com/GoogleChrome/web-vitals/blob/5b91b766031c435e44dda912be75f685d77fe073/src/attribution/onINP.ts#L116) the undefined value with `entry.processingEnd`, the result is always `NaN`.

And finally, when the `NaN` value is [compared](https://github.com/GoogleChrome/web-vitals/blob/5b91b766031c435e44dda912be75f685d77fe073/src/attribution/onINP.ts#L207) with `loaf.startTime` the result is always false, so the LoAF entries are always kept.

### LINKED ISSUE #511: Missing LoAF attributions for INP due to violating internal invariants
#487 introduced a bug that we didn't notice that reduces (often significantly) the number of INPs that get LoAF attribution.

The extra loop added to keep LoAFs that happened after all known events

https://github.com/GoogleChrome/web-vitals/blob/5b91b766031c435e44dda912be75f685d77fe073/src/attribution/onINP.ts#L201-L210

works as expected and is only additive to the LoAFs kept for potential attribution before #487. The problem is that it adds to the `loafsToKeep` set as a second batch of LoAFs, after [the first loop through the LoAFs](https://github.com/GoogleChrome/web-vitals/blob/5b91b766031c435e44dda912be75f685d77fe073/src/attribution/onINP.ts#L192-L200).

When the new array of LoAFs is created from this set (`pendingLoAFs = Array.from(loafsToKeep)`), the `pendingLoAFs` array is now ordered by how values were inserted into `loafsToKeep`. This is bad because `pendingLoAFs` is assumed to be in `startTime` order by `getIntersectingLoAFs`, which will quit looking for intersecting LoAFs early if it thinks it's reached the end of possible intersections:

https://github.com/GoogleChrome/web-vitals/blob/5b91b766031c435e44dda912be75f685d77fe073/src/attribution/onINP.ts#L233-L235

All the correct LoAFs are in `pendingLoAFs`, but `getIntersectingLoAFs` gives up before seeing them. I believe this tends to get worse for busier or longer-lived pages because LoAFs that happened after any known interaction events get put last in this array initially, then they get paired with some event group as events come in so get put into the first batch of LoAFs in the `loafsToKeep` Set. But they're now (potentially) out of order in that batch, spreading the disorder throughout the `pendingLoAFs` array.

Fix: don't do that :) e.g. sort or filter from the existing sorted array.

brendankenny:
fixes #510, fixes #511

This change ensures that the order of `pendingLoAFs` is maintained.

Technically this should also fix potential misordering issues from before #487. Since `pendingEntriesGroups` isn't necessarily in order, the overlapping LoAFs may not be as well, which would also break the ordering invariant. No idea how often this was a problem, though, if ever.

One lingering question: how much can count on the LoAFs coming in order from the PerformanceObserver? That was an assumption in 4.0.0 as well so nothing new, but worth following up on.

### DIFF
--- a/src/attribution/onINP.ts
+++ b/src/attribution/onINP.ts
@@ -63,7 +63,7 @@ let pendingLoAFs: PerformanceLongAnimationFrameTiming[] = [];
 let pendingEntriesGroups: pendingEntriesGroup[] = [];
 
 // The `processingEnd` time of most recently-processed event, chronologically.
-let latestProcessingEnd: number;
+let latestProcessingEnd: number = 0;
 
 // A WeakMap to look up the event-timing-entries group of a given entry.
 // Note that this only maps from "important" entries: either the first input or
@@ -188,7 +188,7 @@ const cleanupEntries = () => {
 
   // Keep all pending LoAF entries that either:
   // 1) intersect with entries in the newly cleaned up `pendingEntriesGroups`
-  // 2) occur after the most recently-processed event entry.
+  // 2) occur after the most recently-processed event entry (for up to MAX_PREVIOUS_FRAMES)
   const loafsToKeep: Set<PerformanceLongAnimationFrameTiming> = new Set();
   for (let i = 0; i < pendingEntriesGroups.length; i++) {
     const group = pendingEntriesGroups[i];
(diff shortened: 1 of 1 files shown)


src/attribution/onINP.ts
```diff
@@ -63,7 +63,7 @@ let pendingLoAFs: PerformanceLongAnimationFrameTiming[] = [];
 let pendingEntriesGroups: pendingEntriesGroup[] = [];
 
 // The `processingEnd` time of most recently-processed event, chronologically.
-let latestProcessingEnd: number;
+let latestProcessingEnd: number = 0;
```
brendankenny, line 66: I blame tsc for allowing this to be `number` without an initial value :P

philipwalton: Yeah, a bit surprising that it didn't catch this.

philipwalton (approved):
LGTM, thanks!

PR #518, opened 2024-08-12
Sort the classes that appear in attribution selectors to reduce cardinality

### LINKED ISSUE #514: Identical abstract selectors can be returned as distinct strings
## The Issue

At the moment, the `web-vitals` attribution builds can return distinct selector strings for elements that have semantically identical selectors. For example, in the image below, you’ll see two selectors that are identical in the abstract/query sense, but distinct as strings:

exact_same_selector_but_with_different_class_order

## A Proposal

In the construction of the selector string, sort the element class names before joining them. This sorting would ensure that the selectors reported above are consolidated and identical. The drawback is that the as-is order of classes could (in theory) help a developer delineate between elements that have the exact same classes but in a different order, but my guess is that that drawback is outweighed by the benefit of cleaner selector aggregation.

tunetheweb:
Fixes #514 

Adding this to v5 branch due reasons given in https://github.com/GoogleChrome/web-vitals/issues/514#issuecomment-2277072435

### DIFF
--- a/src/lib/getSelector.ts
+++ b/src/lib/getSelector.ts
@@ -25,6 +25,7 @@ export const getSelector = (node: Node | null | undefined, maxLen?: number) => {
   let sel = '';
 
   try {
+    console.log('Hi Barry, using sorted selector code');
     while (node && node.nodeType !== 9) {
       const el: Element = node as Element;
       const part = el.id
@@ -34,7 +35,7 @@ export const getSelector = (node: Node | null | undefined, maxLen?: number) => {
           el.classList.value &&
           el.classList.value.trim() &&
           el.classList.value.trim().length
-            ? '.' + el.classList.value.trim().replace(/\s+/g, '.')
+            ? '.' + el.classList.value.trim().split(' ').sort().join('.')
             : '');
       if (sel.length + part.length > (maxLen || 100) - 1) return sel || part;
       sel = sel ? part + '>' + sel : part;
--- a/src/lib/getSelector.ts
+++ b/src/lib/getSelector.ts
@@ -29,10 +29,7 @@ export const getSelector = (node: Node | null | undefined, maxLen?: number) => {
       const el: Element = node as Element;
       const part = el.id
         ? '#' + el.id
-        : getName(el) +
-          (el.classList?.length
-            ? '.' + Array.from(el.classList).sort().join('.')
-            : '');
+        : getName(el) + Array.from(el.classList).sort().join('.');
       if (sel.length + part.length > (maxLen || 100) - 1) return sel || part;
       sel = sel ? part + '>' + sel : part;
       if (el.id) break;
--- a/src/lib/getSelector.ts
+++ b/src/lib/getSelector.ts
@@ -21,21 +21,24 @@ const getName = (node: Node) => {
     : name.toUpperCase().replace(/^#/, '');
 };
 
-export const getSelector = (node: Node | null | undefined, maxLen?: number) => {
+const MAX_LEN = 100;
+
+export const getSelector = (node: Node | null | undefined) => {
   let sel = '';
 
   try {
-    while (node && node.nodeType !== 9) {
+    while (node?.nodeType !== 9) {
       const el: Element = node as Element;
       const part = el.id
         ? '#' + el.id
-        : getName(el) +
-          (el.classList.length
-            ? '.' + Array.from(el.classList).sort().join('.')
-            : '');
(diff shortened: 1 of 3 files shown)


src/lib/getSelector.ts
```diff
@@ -34,7 +34,7 @@ export const getSelector = (node: Node | null | undefined, maxLen?: number) => {
           el.classList.value &&
           el.classList.value.trim() &&
           el.classList.value.trim().length
-            ? '.' + el.classList.value.trim().replace(/\s+/g, '.')
+            ? '.' + el.classList.value.trim().split(' ').sort().join('.')
```
rviscomi, line 37: If we're dropping IE 11 support, we could do something like this, which has the added benefit of deduplicating classes:

```suggestion
            ? '.' + Array.from(el.classList).sort().join('.')
```

I also wonder if there's any reason not to check for el.classList.length, instead of checking the length of the trimmed value.

tunetheweb: Yes I think this works and allows simplifying it a lot more. Which I've done in https://github.com/GoogleChrome/web-vitals/pull/518/commits/c0481e7ffc8ded7a828bde4ee6914ba51e030216

src/lib/getSelector.ts
```diff
@@ -30,11 +30,8 @@ export const getSelector = (node: Node | null | undefined, maxLen?: number) => {
       const part = el.id
         ? '#' + el.id
         : getName(el) +
-          (el.classList &&
-          el.classList.value &&
-          el.classList.value.trim() &&
-          el.classList.value.trim().length
-            ? '.' + el.classList.value.trim().replace(/\s+/g, '.')
+          (el.classList?.length
```
philipwalton, line 33: The `.classList` method will always exist in baseline browsers, and its not necessary to check if it's empty because it will just result in an empty array if it is. So you can just do this instead:

```js
getName(el) + Array.from(el.classList).sort().join('.')
```

tunetheweb: Ah I missed the check for `node` not being `null` (or `undefined`) above and only looked at the top function call (which said it could be one of those).

Fixed now.

rviscomi: Don't we still want the `'.'` before the first class name?

tunetheweb: Ah but only if the `classList` is not empty! That's why we need that check. Reverting the last change.

tunetheweb:
We really need unit tests for this...

philipwalton:
> We really need unit tests for this...

I was just going to add that we should have a test for this behavior.

src/lib/getSelector.ts
philipwalton: Barry, you're giving up at code golf too easily! You could move the `.join()` call to an outer array to handle the case Rick pointed out:

```js
[getName(el), ...[...el.classList].sort()].join('.')
```

tunetheweb: Your JS developers were so preoccupied with whether they could do it in one line... they didn't stop to think if they should.

philipwalton: 😎

brendankenny: the change is replacing the year-old

https://github.com/GoogleChrome/web-vitals/blob/9b932519b16f72328c6d8e9814b811f1bc1a0bb5/src/lib/getSelector.ts#L33-L37

and is still a nested ternary, so let's not pretend it's Hemingway or something :P

philipwalton (changes requested):
Coming back to this PR. I think there are still two things that need to be address before we can merge:

1. Address the [open comment](https://github.com/GoogleChrome/web-vitals/pull/518/files/c0481e7ffc8ded7a828bde4ee6914ba51e030216#r1747631036) to shorten/simplify the selector generation logic.
2. Update the tests to include some cases where there are classes out of order that get properly ordered when the tests are run.

philipwalton (approved):

philipwalton:
I decided to just add them because I wanted to look at this code for some other v5 PRs as well.

src/lib/getSelector.ts
tunetheweb: Oh you were serious about this! Seems more obscure to figure out what's going on but ok.

philipwalton: > Seems more obscure to figure out what's going

This is probably subjective based on who's looking at it. Personally I find that calling `.join()` on a list of items is easier to understand than string concatenation.

PR #520, opened 2024-08-12
Update uniqueID-to-v5

tunetheweb:
For v5 release


tunetheweb:
Is there still value in including the major version number in this? Or should we just remove it?

philipwalton:
The original point of this string was to be able to track version usage of web-vitals in the wild (via HTTP Archive). I think it's still useful for that, so I'd prefer to keep it in.

tunetheweb:
Yeah I recently used this to to identify old beacons so agree let's keep it.

philipwalton (approved):

PR #521, opened 2024-08-14
chore: optimize sessionValue check in onCLS.ts

szc807414589:
I am using version 4.2.1 and sometimes report TypeError
image
image


tunetheweb:
This was fixed in #536 instead.

Issue #159, opened 2021-06-02
getEntriesByName is not supported in some versions of Opera


jroebu14:
Hi there,

We noticed that upgrading from v1.1.1 to v1.1.2 broke our application for our users on data saving mode in the Opera Mini browser.

We believe it was caused by this change https://github.com/GoogleChrome/web-vitals/commit/4dfde193e0c7f7d1dbb002a33c92128fa6ad54c1

It seems that `performance.getEntriesByName` is [not supported in some version of Opera](https://caniuse.com/?search=getEntriesByName).

So that we can upgrade the web-vitals package in our application, would you recommend we look for a polyfill for `performance.getEntriesByName` or should we contribute a fix?

Thanks

philipwalton:
The fix for this was released in [v2.0.1](https://github.com/GoogleChrome/web-vitals/blob/master/CHANGELOG.md#v201-2021-06-02), so you'll have to do a major version update to get it, but hopefully that's not a problem?

There was minimal breaking changes in the v2 update, mostly just the change to how CLS is calculated which you'll probably want to update to get anyway.

jroebu14:
Thanks for your help, @philipwalton. Yeh I don't think upgrading to v2 will be an issue 🙂

PR #525, opened 2024-08-29
Support baseline widely available browsers

tunetheweb:
Change `web-vitals` builds to support [Baseline Widely Available](https://web.dev/baseline)—that is features available in the major browsers 30 months ago or more. This represents between 95.93%-98.69% of desktop browsers, and between 97.32%-99.15% of mobile browsers, according to [RUM Archive](https://rumarchive.com/insights/#baseline).

Users wishing broader support can import and/or build the library themselves targeting more browsers but at the cost of larger bundle size for transpiled syntax, and with many of the features of web-vitals.js either unavailable, only partially available, or available with bugs in older browsers.


rollup.config.js
```diff
@@ -24,8 +24,18 @@ const configurePlugins = ({module}) => {
...
+             * We support Baseline Widely Available (-30 months). This only
+             * affects the built packages in /dist and available through unpkg
+             * and other CDNs. Developers importing or building this library
+             * can support other versions based on their own user bases.
+             *
+             * Last updated: Sep 2024.
+             */
             targets: {
-              browsers: ['ie 11'],
+              'safari': '13.1',
```
philipwalton, line 36:
```js
{
  targets: [
    // Match browser supporting "Baseline Widely Available" features:
    // https://web.dev/baseline
    'chrome >0 and last 2.5 years',
    'edge >0 and last 2.5 years',
    'safari >0 and last 2.5 years',
    'firefox >0 and last 2.5 years',
    'and_chr >0 and last 2.5 years',
    'and_ff >0 and last 2.5 years',
    'ios >0 and last 2.5 years',
  ],
},
```

philipwalton (changes requested):
Can you add something to the description to explain the change? Also, we'll have to update the "browser support" section of the README to remove mentions of IE9.

Then (either in this PR or a follow-up, though maybe this PR makes more sense?) we should make the syntax changes we want to make to modernize the code, e.g. optional chaining, for..of, nullish coalescing, etc.

tunetheweb:
> Can you add something to the description to explain the change? Also, we'll have to update the "browser support" section of the README to remove mentions of IE9.
> 
> Then (either in this PR or a follow-up, though maybe this PR makes more sense?) we should make the syntax changes we want to make to modernize the code, e.g. optional chaining, for..of, nullish coalescing, etc.

Done and done.

Build | Old bytes | New bytes | % Diff
-- | -- | -- | --
web-vitals.attribution.iife.js | 12487 | 11468 | 8.16%
web-vitals.attribution.js | 12489 | 11473 | 8.14%
web-vitals.attribution.umd.cjs | 12678 | 11659 | 8.04%
web-vitals.iife.js | 7208 | 6581 | 8.70%
web-vitals.js | 7197 | 6570 | 8.71%
web-vitals.umd.cjs | 7399 | 6772 | 8.47%

src/attribution/onINP.ts
```diff
@@ -117,9 +117,7 @@ const groupEntriesByRenderTime = (entry: PerformanceEventTiming) => {
 
   // Iterate over all previous render times in reverse order to find a match.
   // Go in reverse since the most likely match will be at the end.
-  for (let i = pendingEntriesGroups.length - 1; i >= 0; i--) {
-    const potentialGroup = pendingEntriesGroups[i];
-
+  for (const potentialGroup of pendingEntriesGroups) {
```
philipwalton, line 120: Ahh, we probably can't make this change because the old code iterated in reverse.

philipwalton:
@tunetheweb I made some updates, can you re-review?
@brendankenny can you review as well for a sanity check?

tunetheweb:
LGTM

philipwalton (approved):

src/attribution/onCLS.ts
```diff
@@ -53,12 +53,7 @@ const attributeCLS = (metric: CLSMetric): CLSMetricWithAttribution => {
     }
   }
 
-  // Use Object.assign to set property to keep tsc happy.
-  const metricWithAttribution: CLSMetricWithAttribution = Object.assign(
-    metric,
-    {attribution},
-  );
-  return metricWithAttribution;
+  return {...metric, attribution};
```
brendankenny, line 56: this returns a new object, though?

brendankenny (commented):
I'm still looking, but this seems to be the only possible issue so far, so flushing the review early

src/onINP.ts
```diff
@@ -63,21 +63,14 @@ export const INPThresholds: MetricRatingThresholds = [200, 500];
...
 ) => {
   // Return if the browser doesn't support all APIs needed to measure INP.
-  if (
-    !(
-      'PerformanceEventTiming' in self &&
-      'interactionId' in PerformanceEventTiming.prototype
-    )
-  ) {
+  const eventTimingProto = globalThis.PerformanceEventTiming?.prototype;
+  if (!(eventTimingProto && 'interactionId' in eventTimingProto)) {
```
brendankenny, line 70: this is a funny one

brendankenny: `Object.hasOwn()` would probably be worse because you'd still need a fallback in the `undefined` case. :shipit:

src/onINP.ts
```diff
@@ -93,11 +86,13 @@ export const onINP = (
...
       whenIdle(() => {
-        entries.forEach(processInteractionEntry);
+        for (const entry of entries) {
+          processInteractionEntry(entry);
+        }
 
         const inp = estimateP98LongestInteraction();
 
-        if (inp && inp.latency !== metric.value) {
+        if (inp?.latency !== metric.value) {
```
brendankenny, line 95: seems a little risky but I guess `metric.value` should never be undefined...

src/attribution/onCLS.ts
```diff
@@ -25,11 +25,11 @@ import {
 } from '../types.js';
 
 const getLargestLayoutShiftEntry = (entries: LayoutShift[]) => {
-  return entries.reduce((a, b) => (a && a.value > b.value ? a : b));
+  return entries.reduce((a, b) => (a?.value > b.value ? a : b));
```
brendankenny, line 28: how would this ever happen?

src/attribution/onCLS.ts
```diff
@@ -25,11 +25,11 @@ import {
...
 };
 
 const getLargestLayoutShiftSource = (sources: LayoutShiftAttribution[]) => {
-  return sources.find((s) => s.node && s.node.nodeType === 1) || sources[0];
+  return sources.find((s) => s?.node?.nodeType === 1) || sources[0];
```
brendankenny, line 32:
```suggestion
  return sources.find((s) => s.node?.nodeType === 1) || sources[0];
```
yah?

src/attribution/onFCP.ts
```diff
@@ -51,12 +51,7 @@ const attributeFCP = (metric: FCPMetric): FCPMetricWithAttribution => {
     }
   }
 
-  // Use Object.assign to set property to keep tsc happy.
-  const metricWithAttribution: FCPMetricWithAttribution = Object.assign(
-    metric,
-    {attribution},
-  );
-  return metricWithAttribution;
+  return {...metric, attribution};
```
brendankenny, line 54: same question about new metric object returned

src/attribution/onINP.ts
```diff
@@ -283,12 +281,7 @@ const attributeINP = (metric: INPMetric): INPMetricWithAttribution => {
     loadState: getLoadState(firstEntry.startTime),
   };
 
-  // Use Object.assign to set property to keep tsc happy.
-  const metricWithAttribution: INPMetricWithAttribution = Object.assign(
-    metric,
-    {attribution},
-  );
-  return metricWithAttribution;
+  return {...metric, attribution};
```
brendankenny, line 284: etc

src/attribution/onLCP.ts
```diff
@@ -37,7 +37,8 @@ const attributeLCP = (metric: LCPMetric): LCPMetricWithAttribution => {
     const navigationEntry = getNavigationEntry();
     if (navigationEntry) {
       const activationStart = navigationEntry.activationStart || 0;
-      const lcpEntry = metric.entries[metric.entries.length - 1];
+      // The `metric.entries.length` check ensures there will be an entry.
+      const lcpEntry = metric.entries.at(-1)!;
```
brendankenny, line 41: annoying that this is different behavior and more annoying that complaining about it just exposes the lie of square bracket access not including `undefined` in the type :)

src/attribution/onLCP.ts
```diff
@@ -83,12 +84,7 @@ const attributeLCP = (metric: LCPMetric): LCPMetricWithAttribution => {
     }
   }
 
-  // Use Object.assign to set property to keep tsc happy.
-  const metricWithAttribution: LCPMetricWithAttribution = Object.assign(
-    metric,
-    {attribution},
-  );
-  return metricWithAttribution;
+  return {...metric, attribution};
```
brendankenny, line 87: same

src/attribution/onTTFB.ts
```diff
@@ -73,12 +73,7 @@ const attributeTTFB = (metric: TTFBMetric): TTFBMetricWithAttribution => {
     };
   }
 
-  // Use Object.assign to set property to keep tsc happy.
-  const metricWithAttribution: TTFBMetricWithAttribution = Object.assign(
-    metric,
-    {attribution},
-  );
-  return metricWithAttribution;
+  return {...metric, attribution};
```
brendankenny, line 76: last one

brendankenny (approved):
LGTM! Mostly a couple of questions and clarifications

src/lib/getNavigationEntry.ts
```diff
@@ -16,9 +16,7 @@
 
 export const getNavigationEntry = (): PerformanceNavigationTiming | void => {
   const navigationEntry =
-    self.performance &&
-    performance.getEntriesByType &&
-    performance.getEntriesByType('navigation')[0];
+    globalThis.performance?.getEntriesByType?.('navigation')[0];
```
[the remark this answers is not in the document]
philipwalton: Yeah, it's is baseline widely available, but we'd previously fixed these bugs, so that's why I was keeping it in:
- https://github.com/GoogleChrome/web-vitals/issues/185
- https://github.com/GoogleChrome/web-vitals/issues/159

src/attribution/onCLS.ts
philipwalton: Ahh, yes, thanks. I forgot I originally didn't want to do that!

src/lib/getNavigationEntry.ts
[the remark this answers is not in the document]
tunetheweb: I wonder if it’s worth testing this branch in Opera Mini to see if there’s going to be any other similar issues?

src/attribution/onCLS.ts
philipwalton: I added a comment to ensure we don't forget in the future.

src/attribution/onCLS.ts
philipwalton: Hmmm, I think you're right that it can't happen? My assumption is I originally thought that since we weren't passing an initial value as the second param, then it would be undefined, but looking at the reduce docs (and when testing) it seems like that's not true.

src/attribution/onCLS.ts
philipwalton: Yeah, I don't think `s` will ever be undefined.

src/attribution/onLCP.ts
philipwalton: Yeah, I was surprised to see that the behavior was different. But on the plus side the warning actually made be double check that there weren't cases where it could be undefined.

src/lib/getNavigationEntry.ts
```diff
@@ -28,9 +26,8 @@ export const getNavigationEntry = (): PerformanceNavigationTiming | void => {
   // https://github.com/GoogleChrome/web-vitals/issues/162
   // https://github.com/GoogleChrome/web-vitals/issues/275
   if (
-    navigationEntry &&
-    navigationEntry.responseStart > 0 &&
-    navigationEntry.responseStart < performance.now()
+    navigationEntry?.responseStart > 0 &&
+    navigationEntry?.responseStart < globalThis.performance?.now()
```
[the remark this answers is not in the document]
philipwalton: Yeah, I guess the behavior is actually different from the previous code. I think adding a comment to point out that comparing `undefined ` to 0 will always be false is fine.

src/onINP.ts
philipwalton: Yeah, I tried a few iterations of this and I couldn't come up with anyone I liked. This one at least minified the best :)

src/onINP.ts
philipwalton: Yeah, `metric.value` should also be a number ... but looking at this again, you're right that it is subtly different and maybe it's better not to use optional chaining in these cases? WDYT?

src/lib/getNavigationEntry.ts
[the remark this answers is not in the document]
philipwalton: Actually I think I changed my mind based on your other comment below. Reverted this to how it was before.

src/onINP.ts
philipwalton: Decided to just revert this since the original code minifies better anyway...

src/lib/getNavigationEntry.ts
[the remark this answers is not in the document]
philipwalton: So, I just tested this in Opera Mini (with extreme data saver mode on) and discovered some interesting things.

- Looks like `performance.now()` is supported, so I guess they must have added that since I last tested.
- ES6 is not supported, so if we're going to be publishing ES6 then we probably don't need to worry about other quirks of Opera Mini on extreme data saver mode
- web-vitals v3 runs without errors, but v4 doesn't work for some reason (not sure what).

When I don't turn on extreme data saver mode, all of the code runs just fine since Opera uses Chromium.

Given all of this, and our intention to support Baseline Widely Available, I think it's fine to start accessing `globalThis.performance` without conditional check.

Anyone have concerns?

philipwalton (approved):

PR #527, opened 2024-08-30
Cap LCP load duration to LCP time

### LINKED ISSUE #509: Update LCP attribution to handle the first frames of animated content
As of [Chrome 116](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/speed/metrics_changelog/2023_08_lcp.md), the LCP definition accounts for the first painted frame of an animated image or video. However, due to a bug in the rollout, this change has not actually taken effect in the LCP API yet.

When fixed, this change means that the LCP time can occur _before_ the LCP resource has finished loading. The LCP phases in the attribution object would need to be updated to handle these cases, specifically resource load duration and render delay, which are both calculated based on the [`lcpResponseEnd`](https://github.com/GoogleChrome/web-vitals/blob/main/src/attribution/onLCP.ts#L66-L74) time. In the current implementation, the sum of the LCP phases would exceed the total LCP time.

The attribution code could detect when this happens by checking to see if the LCP time falls within the load duration phase. When detected, load duration could end at the first byte time of the resource response (`lcpResourceEntry.responseStart`), and render delay set from that moment to the LCP time. Alternatively, if `responseStart` isn't available, for example if the resource is cross-origin and there's no TAO, the load duration could be capped at LCP time and render delay set to 0.

tunetheweb:
Fixes #509 

Videos can have LCP time before finished download so the `resourceLoadDuration` should be capped to LCP time.

I can't see an easy way to add a test case for this, but think the change is fairly innocuous.


src/attribution/onLCP.ts
```diff
@@ -54,9 +54,13 @@ const attributeLCP = (metric: LCPMetric): LCPMetricWithAttribution => {
...
-      const lcpResponseEnd = Math.max(
-        lcpRequestStart,
-        lcpResourceEntry ? lcpResourceEntry.responseEnd - activationStart : 0,
+      const lcpResponseEnd = Math.min(
+        // Cap at LCP time (videos continue downloading after LCP for example)
+        lcpEntry.startTime - activationStart,
+        Math.max(
+          lcpRequestStart,
+          lcpResourceEntry ? lcpResourceEntry.responseEnd - activationStart : 0,
+        ),
```
brendankenny, line 63: Besides handling the video case, clamping on LCP does feel like a more correct constraint if trying to fix up a broken ordering of these numbers (probably a different argument to be made to not fix illogical timings and just report them as an error instead, but conversation for another day :)

Note that it makes the `Math.max` for `lcpRenderTime` below redundant (it'll always be `lcpEntry.startTime - activationStart`), so that should be removed, but treating the reported LCP as a fixed point rather than having the subpart breakdown not add up to LCP seems like a better thing to do.

The constraints still aren't quite right, though. e.g. an LCP before `lcpRequestStart` will win out now, placing `lcpResponseEnd` and `lcpRenderTime` before `lcpRequestStart`, leading to a negative load duration.

It might be helpful to pull out some intermediate variables to make this easier to reason about and to read? e.g. define `rawLcpRequestStart` and `rawLcpResponseEnd` (or named however) directly, and then apply the max/mins to ensure they're ordered correctly.

src/attribution/onLCP.ts
```diff
@@ -54,9 +54,13 @@ const attributeLCP = (metric: LCPMetric): LCPMetricWithAttribution => {
...
       );
       const lcpRenderTime = Math.max(
```
brendankenny, line 65: I just noticed this uses `lcpEntry.startTime - activationStart`, but `metric.value` is also clamped to 0 so it's never negative:

https://github.com/GoogleChrome/web-vitals/blob/9b932519b16f72328c6d8e9814b811f1bc1a0bb5/src/onLCP.ts#L74

Should `lcpRenderTime` do the same thing (I assume we want the subparts to add up to LCP)? Or just use `metric.value` directly

brendankenny (commented):
> I can't see an easy way to add a test case for this, but think the change is fairly innocuous.

How hard would it be to add a video LCP test?

tunetheweb:
> > I can't see an easy way to add a test case for this, but think the change is fairly innocuous.
> 
> How hard would it be to add a video LCP test?

Well it's not just that. We'd need to load a video, and draw the LCP from the first frame, but ensure the video is still downloading after that LCP. That sounds incredibly racy. Plus the implementation in Chrome seems to be broken at the moment which Ian is looking into.

Plus need to consider other browsers (e.g. Firefox), which don't support video yet (I don't think) so might be OK, but another complication...

src/attribution/onLCP.ts
tunetheweb: > Note that it makes the `Math.max` for `lcpRenderTime` below redundant (it'll always be `lcpEntry.startTime - activationStart`), so that should be removed

Good spot! Done.

> but treating the reported LCP as a fixed point rather than having the subpart breakdown not add up to LCP seems like a better thing to do.

I'm not getting what you mean?

> The constraints still aren't quite right, though. e.g. an LCP before lcpRequestStart will win out now, placing lcpResponseEnd and lcpRenderTime before lcpRequestStart, leading to a negative load duration.

I can't see how that would happen though? Or are you thinking about a video when the poster image is loaded before the video? But in that case the LCP resource should be the poster image.

> It might be helpful to pull out some intermediate variables to make this easier to reason about and to read? e.g. define rawLcpRequestStart and rawLcpResponseEnd (or named however) directly, and then apply the max/mins to ensure they're ordered correctly.

Yeah could be made clearer. Will park this for now while we thrash out the actual implementation.

src/attribution/onLCP.ts
tunetheweb: In theory you can never get an LCP entry before an `activationStart` (you need to activate to get a paint, to get LCP), though not sure that's guaranteed for non-TAO cross-origin images... 🤔

I don't think we can use `metric.value` that's a duration and we want a timestamp. Unless we do:
`elementRenderDelay = metric.value - resourceLoadDuration - resourceLoadDelay - timeToFirstByte`
But not sure that's any clearer.

src/attribution/onLCP.ts
brendankenny: > > but treating the reported LCP as a fixed point rather than having the subpart breakdown not add up to LCP seems like a better thing to do.
> 
> I'm not getting what you mean?

before it could bump up `lcpRenderTime` to a later point, even if after LCP, so it seems better to keep `lcpResponseEnd` at `lcpRenderTime` or below, as your change does.

> > The constraints still aren't quite right, though. e.g. an LCP before lcpRequestStart will win out now, placing lcpResponseEnd and lcpRenderTime before lcpRequestStart, leading to a negative load duration.
> 
> I can't see how that would happen though? Or are you thinking about a video when the poster image is loaded before the video? But in that case the LCP resource should be the poster image.

No idea how it could happen, I'm just saying that's what the code says it can do :) Presumably we're clamping to each value because we think it could happen.

src/attribution/onLCP.ts
brendankenny: After this change, `lcpRenderTime` and `metric.value` are the same value as long as they're ≥ 0. I'm trying to clarify if `lcpRenderTime` should ever be negative or should it always match `metric.value`.

tunetheweb: Oh yes LCP is a timestamp. Getting confused with the INP change going on at the same time (when `metric.value` is not a timestamp!).

Yes it should always be the same. So agree it can be set to `metric.value`

src/attribution/onLCP.ts
tunetheweb: > Presumably we're clamping to each value because we think it could happen.

Yeah but we have reasons why we think they could happen.

> It might be helpful to pull out some intermediate variables to make this easier to reason about and to read? e.g. define rawLcpRequestStart and rawLcpResponseEnd (or named however) directly, and then apply the max/mins to ensure they're ordered correctly.

Let this for now as `lcpResponseEnd` is similar to `lcpRequestStart` above.

philipwalton:
> Well it's not just that. We'd need to load a video, and draw the LCP from the first frame, but ensure the video is still downloading after that LCP. That sounds incredibly racy.

I'm not sure if it's possible in Express, but it seems like we should be able to create a stream object that reads the video file on disk and adds an arbitrary setTimeout half way through to simulate this case without having to rely on the racy-ness of the network.

tunetheweb:
>I'm not sure if it's possible in Express, but it seems like we should be able to create a stream object that reads the video file on disk and adds an arbitrary setTimeout half way through to simulate this case without having to rely on the racy-ness of the network.

Still depends on Chrome working as it's supposed to though. So point is kinda moot for now.

tunetheweb:
Raised this bug for the Chrome issue: https://issues.chromium.org/issues/364860066

Once that's resolved (and that version of Chrome feeds through to our test suite) we should see how to add a test, but until then I still think we should merge this change in preparatation.

src/attribution/onLCP.ts
```diff
@@ -54,21 +54,21 @@ const attributeLCP = (metric: LCPMetric): LCPMetricWithAttribution => {
...
       );
-      const lcpResponseEnd = Math.max(
-        lcpRequestStart,
-        lcpResourceEntry ? lcpResourceEntry.responseEnd - activationStart : 0,
-      );
-      const lcpRenderTime = Math.max(
-        lcpResponseEnd,
+      const lcpResponseEnd = Math.min(
+        // Cap at LCP time (videos continue downloading after LCP for example)
         lcpEntry.startTime - activationStart,
```
philipwalton, line 59: Should this be `metric.value` as well?

src/attribution/onLCP.ts
```diff
@@ -54,21 +54,21 @@ const attributeLCP = (metric: LCPMetric): LCPMetricWithAttribution => {
...
+        ),
       );
 
       attribution = {
         element: getSelector(lcpEntry.element),
         timeToFirstByte: ttfb,
         resourceLoadDelay: lcpRequestStart - ttfb,
         resourceLoadDuration: lcpResponseEnd - lcpRequestStart,
-        elementRenderDelay: lcpRenderTime - lcpResponseEnd,
+        elementRenderDelay: metric.value - lcpResponseEnd,
```
philipwalton, line 71: I don't remember why I calculated `lcpRenderTime` instead of just using `metric.value`. It's totally possible I just missed that I could, but I'm also wondering if there was another reason...

How much did you check this for edge cases?

tunetheweb: It was @brendankenny that pointed out this could be `metric.value` above. I can't think of any reason why reason why it shouldn't be that, even with edge cases... 🤔

src/attribution/onLCP.ts
tunetheweb: Makes sense. Updated in https://github.com/GoogleChrome/web-vitals/pull/527/commits/0136440a98826bf4081bc401b78fa8ebb3348cc0

brendankenny:
Should this wait until we have a clear picture of what exactly ships (and when) in Chrome?

tunetheweb:
> Should this wait until we have a clear picture of what exactly ships (and when) in Chrome?

I don't see any reason NOT to cap it at the metric value anyway. Especially as I know of at least one reason (that's supposed to have shipped already).

src/attribution/onLCP.ts
philipwalton: 👍

philipwalton (approved):

philipwalton:
Agreed. This seems low risk and may help avoid future bugs.

PR #528, opened 2024-08-30
Cap INP breakdowns to INP duration

### LINKED ISSUE #492: Heads Up on Changes to INP Measurement When Interfered with JS Modal Dialogs
With landing of [EventTimingFallbackToModalDialogStart](https://chromium.googlesource.com/chromium/src/+/main/docs/speed/metrics_changelog/2024_07_inp.md#enable-eventtimingfallbacktomodaldialogstart-by-default), event timing now will treat JS modal dialog showing time as an alternative end point of INP or event duration measurement.

It has following impact on WebVitals.js v4:

* Events will have a short duration (when interfered with JS modal dialogs), which is good
* The Event Timing processing is still large (this is a weirdness in event timing itself)

So now we can have a situation like:
INP: 100ms, input delay: 20ms, processing: 10000ms, presentationDelay: 0ms
This doesn't break web-vitals.js which is good, and it's also technically the truth:)

However, that huge event timing processing is also used to get a LoAF attribution, and it can be something that users will find surprising if it comes from field data, and we might want to signal it somehow.
(Loaf ::Script will have a pauseDuration for dialogs, which could be one strategy to mitigate this)

tunetheweb:
Fixes #492 

Modals (`alert()`, `confirm()`, `print()`) do give feedback to the user, but also block the main thread so the event processing doesn't end when this shows. However, from an INP point of view, the feedback is given and so the duration is set to that timestamp.

I think it should be possible add a test case for this, but wanna agree on the change before going to that effort.


src/attribution/onINP.ts
```diff
@@ -273,7 +280,14 @@ const attributeINP = (metric: INPMetric): INPMetricWithAttribution => {
     longAnimationFrameEntries.map((loaf) => loaf.startTime + loaf.duration),
   );
 
-  const nextPaintTime = Math.max.apply(Math, nextPaintTimeCandidates);
+  // processingEnd can extend beyond duration for modals where we artificially
+  // mark event timing duration as that paint.
+  // See: https://github.com/GoogleChrome/web-vitals/issues/492
+  // So cap to the INP value.
+  const nextPaintTime = Math.min(
```
mmocny, line 287: I'm not sure about this change...

Looking at like 276 above, you are already including `startTime + duration` in the list of paint time candidates, which means its in the list of MAX and then also in the list of MIN.  I think that means its always going to be picked?

(Note: I assume there is no difference between `firstEntry.duration` and `metric.value` but there can be)

Second, the comment anove is:

```js
// Since entry durations are rounded to the nearest 8ms, we need to clamp
// the `nextPaintTime` value to be higher than the `processingEnd` or
// end time of any LoAF entry.
```

I think we are specifically looking for cases where `startTime + duration` is "accidentally" less than one of the alternative paintTimeCandidates (such as processingEnd) when we do rounding.

Now here by using `startTime + duration` when it is less than the alternatives, we are effectively just undoing that?

Might as well just delete this code instead? Or, selectively do it if its "less than 8ms off" or something?

---

Finally, also already capped the processingEnd time in this CL, so its also not useful to look at processingEnd as an alternative to paint time any more (it will always be smaller now).

tunetheweb: > Now here by using startTime + duration when it is less than the alternatives, we are effectively just undoing that?

Urgh! I realised this earlier, and then completely forgot about it and committed it with further.

> Might as well just delete this code instead?

I think we need to do _something_ here, or we'll just move this incorrect breakdown to presentation delay.

> Or, selectively do it if its "less than 8ms off" or something?

Yeah think we need something like this. Or if `processingEnd` is `firstEntry.startTime + metric.value`, then we can 0 out the presentation time.

tunetheweb: Have another look not if you can @mmocny

src/attribution/onINP.ts
```diff
@@ -239,10 +239,16 @@ const getIntersectingLoAFs = (
 
 const attributeINP = (metric: INPMetric): INPMetricWithAttribution => {
   const firstEntry = metric.entries[0];
+  const inpTime = firstEntry.startTime + metric.value;
```
mmocny, line 242: Nit: just call it `endTime` or something

src/attribution/onINP.ts
```diff
@@ -239,10 +239,16 @@ const getIntersectingLoAFs = (
...
   const group = entryToEntriesGroupMap.get(firstEntry)!;
 
   const processingStart = firstEntry.processingStart;
-  const processingEnd = group.processingEnd;
+  // processingEnd can extend beyond duration for modals where we artificially
+  // mark event timing duration as that paint.
```
mmocny, line 247: Nit: "as that paint" comment is confusing (there is no actual paint)

Perhaps just "where duration ends at the time modal dialog started (and provided visual feedback)"

src/attribution/onINP.ts
```diff
@@ -274,6 +280,10 @@ const attributeINP = (metric: INPMetric): INPMetricWithAttribution => {
   );
 
   const nextPaintTime = Math.max.apply(Math, nextPaintTimeCandidates);
+  // If processingEnd has been capped to inpTime then presentationDelay is 0
+  // Else use the nextPaintTime
+  const presentationDelay =
+    processingEnd == inpTime ? 0 : Math.max(nextPaintTime - processingEnd, 0);
```
mmocny, line 286: I think this is a fine temporary solution, but, consider just creating a variable such as "used_fallback_time", set up top when `endTime < processingEnd`, and re-use that everywhere.

(The goal is to expose a clue to event timing for this value, eventually.)

BTW the case where endTime is rounded to within 8ms of processingEnd is almost always another "used_fallback_time" example (i.e. when we actually dont have a paint)

mmocny (approved):
I made some nit comments, but pre-approving.

Probably best to validate carefully...

src/attribution/onINP.ts
tunetheweb: Updated

src/attribution/onINP.ts
tunetheweb: Updated.

src/attribution/onINP.ts
tunetheweb: I'd prefer not to do that variable name. We haven't used a fallback time — Event timing has. We are always using the actual time given by that.

The current wording is similar to the LCP implementation in #527

mmocny: Thats fine.

---

The goal for Event Timing will be to:
- Expose `presentation_time` (possibly as `renderTime` to match other paint timing specs)
- Expose `fallback_time` (might be named something else)
- `duration` will be based on `fallback_time || presentation_time`, which is already is under the hood, but now you will know why.

Also, in hindsight, my suggestion of comparing with `processingEnd` only works if your event listener(s) caused the alert, but wouldn't work if some other task in the same animation frame did. Nor would it work for other cases of fallback_time like when you don't need next paint, or the page unloaded and hit visibilitychange as an alternative end time...

src/attribution/onINP.ts
```diff
@@ -274,6 +281,10 @@ const attributeINP = (metric: INPMetric): INPMetricWithAttribution => {
   );
 
   const nextPaintTime = Math.max.apply(Math, nextPaintTimeCandidates);
+  // If processingEnd has been capped to endTime then presentationDelay is 0
+  // Else use the nextPaintTime
+  const presentationDelay =
+    processingEnd == endTime ? 0 : Math.max(nextPaintTime - processingEnd, 0);
```
philipwalton, line 287: I not sure this quite works, because I think it's possible that `endTime` is earlier than `nextPaintTime` (if the duration value is rounded down to the nearest 8ms), which would mean that the numbers wouldn't add up to the total INP duration.

I think it's safer to do all of the capping at the end rather than at various places through the logic. I also think it's potentially problematic (and confusing) to have both `endTime` and `nextPaintTime` and have those values possible be different things.

philipwalton: Maybe the best solution here is to just do away with the logic that tries to be smart about figuring out more accurate `nextPaintTime` values and just use the values given.

philipwalton: OK, I updated the code to do this. @tunetheweb let me know what you think.

tunetheweb: I like it. Much simpler.

My only comment is to wonder if we should apply the cap in `groupEntriesByRenderTime` instead? Because at the moment `getIntersectingLoAFs(group.startTime, group.processingEnd)` (as just one example) would include LoAFs after INP. For example:

- 10,000 ms: Interaction happens
- 10,016 ms: The interaction processing shows an `alert()`
- 10,030 ms: `alert()` is rendered and INP is capped.
- 11,000 ms: User dismisses the `alert()`

This produces a 1 second LoAF. Which technically is part of the INP interaction, but also technically doesn't matter that much (as INP ended before that), so is noise that doesn't need reporting (or maybe it's good to surface this?).

Capping `group.processingEnd` at the time it's set, would avoid that LoAF being included.

Also I'm not sure if in future people would forget to handle the fact that `group.processingEnd` can extend past INP time. Again capping `group.processingEnd` would mean anyone using `group` wouldn't have to consider that.

philipwalton: The current code gets intersecting LoAFs based on the capped `processingEnd` rather than `group.processingEnd`, so I think it's already doing what you're suggesting (unless I'm misunderstanding you)?

tunetheweb: Oh so it does. The `cleanupEntries` code uses the `group.processingEnd` (which is not bad as meant it'll be kept cause it might be needed). My bad. But goes to show it's kinda confusing.

philipwalton: Yeah, it is confusing...and I'm honestly not sure if it matters which one we choose because I think LoAF does report a long frame in this case, right? So that frame would get returned regardless of which processingEnd value was used?

tunetheweb: Agreed it is a LoAF. And I think it's right to be reported as a LoAF (though potentially not right to include that LoAF in INP attribution).

Anyway, I'm OK with the current implementation.
But I'm also OK with moving the capping to `groupEntriesByRenderTime` if that's less confusing (so you therefore wouldn't need the capping here as well).

WDYT is better?

philipwalton: My main concern with moving the capping to `groupEntriesByRenderTime()` is that we don't currently have the "real" `renderTime` value in that function, because each entry has its own `startTime+duration` value. So to properly cap it there, we'd have to also store the `renderTime` for the group (which is okay, but would add additional code).

tunetheweb: OK let's leave as is then. Happy to merge.

PR #537, opened 2024-10-02
Make event handlers that are only needed to run once, to only run once

### LINKED ISSUE #534: 【Discuss & May Performance Issue】whenIdle add never removed visibilitychange event listener
I was just tried to collect LCP metric for my project，and when I just add `onLCP` for my project, I got an performance issue at the Performance monitor panel：the JS event listeners increased at every operate on the page
and I tried to debug the code，and I got the code [onLCP.ts#L103-L108](https://github.com/GoogleChrome/web-vitals/blob/main/src/onLCP.ts#L103-L108) may case the performance issue
```typescript
['keydown', 'click'].forEach((type) => {
        // Wrap in a setTimeout so the callback is run in a separate task
        // to avoid extending the keyboard/click handler to reduce INP impact
        // https://github.com/GoogleChrome/web-vitals/issues/383
        addEventListener(type, () => whenIdle(stopListening), true);
});
```
1. it add `click` and `keydown` event, and the handler was wrapped with whenIdle
2. whenIdle will call `onHidden` which will add `visibilitychange` listener when the page's visibilityState is not 'hidden', and will never be removed.
3. each time the page (or tab for chrome) visibility change(hidden -> visible), the listeners will execute

click/keydown -> whenIdle process -> add visiblitychange listener（always increase never remove）

for the onINP each time the PerformanceObserve.observe('event') was fired, the handle function（handleEntries）call whenIdle which add a visiblitychange listener

It's a **performance issue** or **designed**?

tunetheweb:
Fixes #534

### DIFF
--- a/src/lib/onHidden.ts
+++ b/src/lib/onHidden.ts
@@ -14,10 +14,14 @@
  * limitations under the License.
  */
 
-export const onHidden = (cb: () => void) => {
-  document.addEventListener('visibilitychange', () => {
-    if (document.visibilityState === 'hidden') {
-      cb();
-    }
-  });
+export const onHidden = (cb: () => void, once: boolean = false) => {
+  document.addEventListener(
+    'visibilitychange',
+    () => {
+      if (document.visibilityState === 'hidden') {
+        cb();
+      }
+    },
+    {once: once},
+  );
 };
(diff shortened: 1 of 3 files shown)


tunetheweb:
Closing in favour of #538

PR #538, opened 2024-10-02
Ensure one-time event listeners only run once

philipwalton:
Fixes https://github.com/GoogleChrome/web-vitals/issues/534.

This PR removes the `onHidden()` utility in favor of manually using `visibilitychange` listeners, which gives more control over adding listeners that just need to run once.


tunetheweb (approved):
Approving, but as discussed I would suggest making it clearer that LCP's `visibilitychange` change event is handled right away (either with a comment, renaming `whenIdle` to explain when it does better, or by moving it back to it's own, separate call).

src/attribution/onINP.ts
```diff
@@ -79,7 +79,7 @@ export const interactionTargetMap: Map<number, Node> = new Map();
 // A reference to the idle task used to clean up entries from the above
 // variables. If the value is -1 it means no task is queue, and if it's
 // greater than -1 the value corresponds to the idle callback handle.
```
tunetheweb, line 81: Can we update this comment?

philipwalton: Good catch!

src/lib/whenIdleOrHidden.ts
```diff
@@ -14,25 +14,19 @@
...
-  cb = runOnce(cb);
   // If the document is hidden, run the callback immediately, otherwise
   // race an idle callback with the next `visibilitychange` event.
   if (document.visibilityState === 'hidden') {
     cb();
   } else {
-    handle = rIC(cb);
-    onHidden(cb);
+    rIC(cb);
+    document.addEventListener('visibilitychange', cb, {once: true});
```
youngBrain1893, line 30: 【discuss】why the cb should be called two times？
- when it's hidden：call it once，immediately
- when it's visible：call it once with rIC and another time when change to hidden

before the change，it‘s wrapped with runOnce will only call once

philipwalton: Thanks for catching this! I've created to https://github.com/GoogleChrome/web-vitals/pull/541 to address this problem.

tunetheweb: Yes. Very nice catch!

PR #549, opened 2024-10-16
Export attributeINP

tunetheweb:
This is useful for those wanting to monitor individual interactions (e.g. DevTools) for those willing to reach into our source for this function to do that.

We won't include in our built files (e.g. dist/web-vitals.js), but will include in our source build files (e.g. `dist/modules/attribution/onINP.js`)

CC: @adamraine

### DIFF
--- a/src/attribution/onINP.ts
+++ b/src/attribution/onINP.ts
@@ -233,6 +233,8 @@ const getIntersectingLoAFs = (
   return intersectingLoAFs;
 };
 
+// Export to allow it to be used from dist if needed
+// See https://github.com/GoogleChrome/web-vitals/pull/549
 export const attributeINP = (metric: INPMetric): INPMetricWithAttribution => {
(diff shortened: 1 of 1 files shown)


philipwalton:
My only concern with this is that technically the exports in `src/attribution/onINP.ts` are part of the public API since that file is part of the [package exports](https://github.com/GoogleChrome/web-vitals/blob/9b932519b16f72328c6d8e9814b811f1bc1a0bb5/package.json#L62-L65).

If we don't want this function to be part of the public API (which I think we don't?) then it'd be better to move this logic into a separate file that `src/attribution/onINP.ts` imports, but which itself is not listed in the package exports.

Alternatively, if DevTools is able to automate adding this in their code, then this change isn't needed (and that would be my preference).

tunetheweb:
This might not be needed so closing for now.

PR #553, opened 2024-10-22
Add guard against getEntriesByType

tunetheweb:
Causing failures when testing Google upgrades.

### DIFF
--- a/src/lib/getNavigationEntry.ts
+++ b/src/lib/getNavigationEntry.ts
@@ -15,7 +15,11 @@
  */
 
 export const getNavigationEntry = (): PerformanceNavigationTiming | void => {
-  const navigationEntry = performance.getEntriesByType('navigation')[0];
+  // JSDOM does not implement getEntriesByType: https://github.com/jsdom/jsdom/issues/3309
+  const navigationEntry =
+    self.performance &&
+    performance.getEntriesByType &&
+    performance.getEntriesByType('navigation')[0];
 
   // Check to ensure the `responseStart` property is present and valid.
   // In some cases a zero value is reported by the browser (for
--- a/src/lib/getNavigationEntry.ts
+++ b/src/lib/getNavigationEntry.ts
@@ -16,10 +16,7 @@
 
 export const getNavigationEntry = (): PerformanceNavigationTiming | void => {
   // JSDOM does not implement getEntriesByType: https://github.com/jsdom/jsdom/issues/3309
-  const navigationEntry =
-    self.performance &&
-    performance.getEntriesByType &&
-    performance.getEntriesByType('navigation')[0];
+  const navigationEntry = performance.getEntriesByType?.('navigation')[0];
 
   // Check to ensure the `responseStart` property is present and valid.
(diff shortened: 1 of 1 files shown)


src/lib/getNavigationEntry.ts
```diff
@@ -15,7 +15,11 @@
  */
 
 export const getNavigationEntry = (): PerformanceNavigationTiming | void => {
-  const navigationEntry = performance.getEntriesByType('navigation')[0];
+  // JSDOM does not implement getEntriesByType: https://github.com/jsdom/jsdom/issues/3309
```
vanderhoop, line 18: Feels sad to accommodate an inadequacy of a test-env-specific library, given that this API is supported in all major browsers: https://caniuse.com/mdn-api_performance_getentriesbytype

tunetheweb: I agree. However it's a guard, rather than a polyfill, or holding ourselves back in using new features or syntac. So I'm comfortable with this. We've a few cases of this in the library.

But let's see if @philipwalton agrees!

brendankenny: Optional chaining could make this not so bad, but in general if you're using jsdom for testing, you'll be mocking out a bunch of stuff anyways and it shouldn't generally be put on the library to do that.

Nothing will work in jsdom anyways, looks like there's [no PerformanceObservers either](https://github.com/jsdom/jsdom/issues/3713), so I wonder if the tester should just short circuit it completely.

tunetheweb: For those we wrap it in a try/catch anyway:
https://github.com/GoogleChrome/web-vitals/blob/v5/src/lib/observe.ts
Hence why they didn’t cause a problem.

For TTFB however we just call it as it doesn’t need a PerformanceObserver.

I don’t think the intent is to test the library in JSDOM, it’s more the whole page is being loaded in that (including this script) and it hits this error.

tunetheweb: Oh and I tried optional chaining initially and still complained `getEntriesByTyoe` wasn’t a function. Because performance does exist but not much under it. Open to ideas on how to implement this better if you know some better optional chaining syntax for that? For now I just reverted back to what we had in v4.

brendankenny: > I don’t think the intent is to test the library in JSDOM, it’s more the whole page is being loaded in that (including this script) and it hits this error.

right, but I guarantee this isn't the only thing broken in jsdom and they already have a "set up the mocks" section somewhere because there's always something (usually several things) that need to be mocked out 

> Open to ideas on how to implement this better if you know some better optional chaining syntax for that?

`performance.getEntriesByType?.('navigation')[0]` should work

tunetheweb: Oh that worked! Thank you. Updated in https://github.com/GoogleChrome/web-vitals/pull/553/commits/16f4a51ad53aa094490ec5f8811361b659d606b4

tunetheweb:
Closing after discussion off line and finding the fix in the offending code that depended on this.

PR #555, opened 2024-10-22
Prep CHANGELOG for v4.2.4

tunetheweb:


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 See the [upgrading to v4](/docs/upgrading-to-v4.md) documentation for a complete list of all API changes in version 4.
 
+### v4.2.4 (2024-10-10)
```
philipwalton, line 5: This should be whatever date we plan on releasing this version.

philipwalton (changes requested):

CHANGELOG.md
tunetheweb: Oh that was a typo and was supposed to be this:

```suggestion
### v4.2.4 (2024-10-22)
```

Think dealing with wacky date formats for a certain, unnamed, country of the world messed me up!

philipwalton: That's why we use ISO-8601 here!

philipwalton (approved):

PR #558, opened 2024-10-29
Add upgrading to v5 doc

tunetheweb:
This may be a little verbose but let me know your thoughts.


rviscomi (approved):

docs/upgrading-to-v5.md
```diff
@@ -0,0 +1,37 @@
...
+- **Changed** the browser support policy to [Baseline Widely available](https://web.dev/baseline) browser support. ([#525](https://github.com/GoogleChrome/web-vitals/pull/525))
+
+##### More details on the Baseline Widely available change:
+
+The pre-built [packages of the `web-vitals` library available on CDNs](/README#load-web-vitals-from-a-cdn) are built with this Baseline Widely available support in mind. This should not be an issue for modern browsers released in the last few years, which are usually required in order to measure the metrics this library is designed for.
+
+For basic use of the library we recommend loading the `web-vitals` library in its own `<script type="module">` tag, regardless of whether loading this library [from a CDN](README.md#from-a-cdn) or hosted locally on your site. This will allow this library to fail without affecting other code if unsupported syntax or features are used on very old browsers.
```
philipwalton, line 18:
```suggestion
We recommend loading the `web-vitals` library separately from your site's main application bundle(s), either via its own `<script type="module">` tag or via your bundlers code splitting feature. This ensures that any errors encountered while loading the library to not impact other code on your site.
```

docs/upgrading-to-v5.md
```diff
@@ -0,0 +1,37 @@
...
+### Standard build
+
+#### General
+
+- **Removed** First Input Delay (FID) support and `onFID()` [as deprecated in v4](./upgrading-to-v4.md#%EF%B8%8F-deprecations). FID has been replaced by INP and this [removal was previously advertised in that announcement post](https://web.dev/blog/inp-cwv-launch#fid_deprecation_timeline). ([#519](https://github.com/GoogleChrome/web-vitals/pull/519))
+- **Changed** the browser support policy to [Baseline Widely available](https://web.dev/baseline) browser support. ([#525](https://github.com/GoogleChrome/web-vitals/pull/525))
+
+##### More details on the Baseline Widely available change:
+
+The pre-built [packages of the `web-vitals` library available on CDNs](/README#load-web-vitals-from-a-cdn) are built with this Baseline Widely available support in mind. This should not be an issue for modern browsers released in the last few years, which are usually required in order to measure the metrics this library is designed for.
```
philipwalton, line 16:
```suggestion
All of the [builds](README#build-options) in `web-vitals` v5 use only [Baseline Widely available](https://web.dev/baseline) APIs, which means they should run without error in all browsers released in the last few years. Note that the Core Web Vitals metrics are only available in modern browsers, so legacy browser support is unnecessary for this library.

If your site needs to support legacy browsers, you can still use the `web-vitals` library without causing errors in those browsers by adhering to the following recommendations:
```

philipwalton (commented):
Made a few language suggestions. Let me know what you think.

docs/upgrading-to-v5.md
```diff
@@ -0,0 +1,37 @@
...
+#### `INPAttribution`, `LCPAttribution`, and `CLSAttribution`
+
+- **Changed** to sort the classes that appear in attribution selectors to reduce cardinality. While not a breaking change in the API, this may result in a short term difference in reports based on the selector during the change over from v4 to v5, but longer term should result in few selectors that are more easily grouped in reporting. ([#518](https://github.com/GoogleChrome/web-vitals/pull/518))
+
+## 🚀 New features
+
+- **Added** additional cappings have been added to LCP ([#527](https://github.com/GoogleChrome/web-vitals/pull/527)) and INP ([#528](https://github.com/GoogleChrome/web-vitals/pull/528)) breakdowns to handle edge cases.
+- **Added** support for generating custom targets in the attribution build ([#585](https://github.com/GoogleChrome/web-vitals/pull/585))
+- **Added** extended INP attribution with extra LoAF information: longest script and buckets ([#592](https://github.com/GoogleChrome/web-vitals/pull/592))
+- **Changed** event listeners to run less often ([#538](https://github.com/GoogleChrome/web-vitals/pull/538))
```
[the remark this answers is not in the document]
tunetheweb: There were reports of member leaks hence why I mentioned this. WDYT?

docs/upgrading-to-v5.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Attribution build
+
+#### `INPAttribution`, `LCPAttribution`, and `CLSAttribution`
+
+- **Changed** to sort the classes that appear in attribution selectors to reduce cardinality. While not a breaking change in the API, this may result in a short term difference in reports based on the selector during the change over from v4 to v5, but longer term should result in few selectors that are more easily grouped in reporting. ([#518](https://github.com/GoogleChrome/web-vitals/pull/518))
+
+## 🚀 New features
+
+- **Added** additional cappings have been added to LCP ([#527](https://github.com/GoogleChrome/web-vitals/pull/527)) and INP ([#528](https://github.com/GoogleChrome/web-vitals/pull/528)) breakdowns to handle edge cases.
```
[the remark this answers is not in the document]
tunetheweb: Yeah I'm OK with that. Struggled to understand whether the upgrading doc was a full changelog for v5.0.0 or merely an upgrading guide. Guess it's more the latter.

```suggestion
```

docs/upgrading-to-v5.md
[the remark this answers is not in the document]
tunetheweb: Guess that's similar to the next item and not an "upgrading" issue. Thought it is a reason to upgrade! :-)

```suggestion
```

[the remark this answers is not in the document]
philipwalton: You mentioned this in another comment, but in my mind the "Upgrade Guide" isn't supposed to be everything that changed, it's just so supposed to be the most important information someone needs to know when upgrading—so basically any breaking changes as well the main new features they might want to add.

docs/upgrading-to-v5.md
tunetheweb: I think we should call out that packages on unpkg and the like are affected more and don't think your suggestions do that strongly enough. WDYT?

docs/upgrading-to-v5.md
philipwalton: I updated the suggestion to make it clearer that I'm we're not recommending that. Do you think this is better / more clear?

docs/upgrading-to-v5.md
philipwalton: Why are they affected more?

docs/upgrading-to-v5.md
tunetheweb: I can't see an update?

docs/upgrading-to-v5.md
tunetheweb: Yeah I guess I'm reading "build options" as more of a way of building it, rather than "pre-built packages available to install from CDNs". But guess it's the same.

And, to your other point, these pre-built options WILL at least be in their own `<script>` (and hopefully `<script type="module">`!) elements.

Let's go with your suggestion then.

docs/upgrading-to-v5.md
philipwalton: Do you see "3 hours ago - edited" next to my name? If so then you can see the changes there.

tunetheweb: Oh so you’re still “recommending” that? So back to my original question:

> Do we really recommend this? ~So we want people to depend on globals?~

Ok with `type=module` you don’t need global. By bad on that. So ignore that bit.

for simple static sites this recommendation makes sense. But for more complex sites, or those using web-vitals as part of a RUM solution the use of the word “recommend” seems a bit strong. I now it’s covered in the next section nut “recommend” here seems to suggest they are holding this wrong.

philipwalton: For complex sites I'm assuming people would `npm install web-vitals` into their dependencies list and then they'd do something like this in the JS code:

```js
import('web-vitals').then(({onCLS, onINP, onLCP}) => {
  // Use onCLS, onINP, and onLCP as normal...
});
```

Doing it this way via dynamic `import()` means most bundlers will code-split out all of the `web-vitals` related code into its own script file that gets loaded separately. This is what I meant by "via your bundlers code splitting feature".

For more simplistic sites, I think we want people to follow the example in the README, like this:

```html
<script type="module">
  import {onCLS, onINP, onLCP} from 'https://unpkg.com/web-vitals@4?module';

  onCLS(console.log);
  onINP(console.log);
  onLCP(console.log);
</script>
```

With both of these, if the `web-vitals` code fails to parse or execute, it won't break any of the other code on the page.

What we **don't** want people to do is something like this:

```js
import {React} from 'react';
import {onCLS, onINP, onLCP} from 'web-vitals';
// ...
```

Because doing it this way would case everything to break in older browsers if the site owner doesn't know how to configure their bundler to transpile and add polyfills for `web-vitals`.

So, do you think my suggested phrasing doesn't convey what I have here above?

philipwalton: For people using RUM solutions, I expect they'd just follow their RUM providers installation instructions. So I don't think we need to worry about that?

tunetheweb: >```js
>import('web-vitals').then(({onCLS, onINP, onLCP}) => {
>  // Use onCLS, onINP, and onLCP as normal...
>});
>```

OK this does seem like the best way. But I don't think it's that common. We don't even give it as an example in the README. And I didn't think of that.

> For more simplistic sites, I think we want people to follow the example in the README...

100% agree. That one was clear.

> What we don't want people to do is something like this:
> ...
> Because doing it this way would cause everything to break in older browsers if the site owner doesn't know how to configure their bundler to transpile and add polyfills for web-vitals.

Yes I agree. But that wasn't clear to me. I wonder how we can make this clearer. Do we need to update the README to show this example? And possibly link to that from this doc.

> For people using RUM solutions, I expect they'd just follow their RUM providers installation instructions. So I don't think we need to worry about that?

For RUM I mean more RUM developers themselves. We know there's a number of them using this site. But the it's the same advice as for complex sites bundling so doesn't need to be called out separately. I just mean they do bundle.

philipwalton: > OK this does seem like the best way. But I don't think it's that common. We don't even give it as an example in the README. And I didn't think of that.
> ...
> Yes I agree. But that wasn't clear to me. I wonder how we can make this clearer. Do we need to update the README to show this example? And possibly link to that from this doc.

Yeah, I think we could update the README to better promote these patterns, but TBH these aren't really specific to the `web-vitals` library. These are best practices that apply to any library that isn't core to the functionally of a site.

To the larger point, though, I don't think we need to promote them in the upgrading guide.

PR #562, opened 2024-11-06
Save a reference to the LCP and CLS targets in case they are removed from the DOM

### LINKED ISSUE #561: LCP Loses Entry Element After SPA Route Change
When using the web-vitals library to track the Largest Contentful Paint (LCP) metric in a Single Page Application (SPA), we encounter an issue where the LCP entry element is lost upon changing routes. This problem arises because the web-vitals library captures the LCP metric based on the elements rendered on the page at the time of measurement. 

In SPAs, content is often loaded dynamically, and when navigating to a new route, the relevant content may not be fully rendered or may change, causing the library to lose track of the largest contentful element. As a result, the reported LCP value may not accurately reflect the user experience, leading to misleading performance metrics.

How to reproduce
1) Create any default SPA (Vite etc)
2) Start listerning onLCP (web-vitals or web-vitals/attribution)
3) Change route before the scroll on any other interaciton

tunetheweb:
Fixes #561

Note this also corrects an inconsistency where lcp selector in the attribution object was labeled `element` rather than `target` (like CLS and INP).


tunetheweb:
Test here: https://disappearing-cwv-node.glitch.me/index.html

Maybe we should add this to the test suite?
And maybe we should do this for CLS too?

src/attribution/onLCP.ts
```diff
@@ -16,14 +16,20 @@
...
+} from '../onLCP.js';
 import {
   LCPAttribution,
   LCPMetric,
   LCPMetricWithAttribution,
   ReportOpts,
 } from '../types.js';
 
+// A reference to the LCP Target node incase it is removed before reporting
+let lcpTargetNode: Node;
```
philipwalton, line 31: Rather than using a single variable reference to a `Node` here, I think it would be better to use a `WeakMap<PerformanceEventTiming, Node>` and associate each entry with its respective target.

That feels more conceptually in line with what you're trying to accomplish in this PR (e.g. for every entry that had a target element at one point in time, make sure you can always get a reference to that target as long as you have a reference to the entry).

philipwalton (changes requested):
> Maybe we should add this to the test suite?

Yes, I think we want a test for this behavior. It should be possible to test.

src/attribution/onLCP.ts
```diff
@@ -65,7 +71,7 @@ const attributeLCP = (metric: LCPMetric): LCPMetricWithAttribution => {
       );
 
       attribution = {
-        element: getSelector(lcpEntry.element),
+        element: getSelector(lcpEntry.element || lcpTargetMap.get(lcpEntry)),
```
philipwalton, line 74:
```suggestion
        element: getSelector(lcpEntry.element ?? lcpTargetMap.get(lcpEntry)),
```
I don't think this will ever make a difference, but may as well use the more appropriate operator for the job.

tunetheweb: I presume it actually makes sense to do this rather than just `lcpTargetMap.get(lcpEntry)` all the time? Would be smaller code to just do that and in theory it should be the same thing but guess the `lcpEntry.element` might be slightly faster? WDYT?

philipwalton: I think this is fine, though I can't think of any cases where just using `lcpTargetMap.get()` would return anything different...

rviscomi, line 74: Should you also delete the entry from the map after this?

tunetheweb:
> Yes, I think we want a test for this behavior. It should be possible to test.

Added tests for LCP, CLS and even the INP scenario added in #477

dmitiiv:
Can you add the same logic for metric -> lcpEntry -> element?

README.md
```diff
@@ -927,9 +927,14 @@ interface INPAttribution {
 ```ts
 interface LCPAttribution {
   /**
-   * The element corresponding to the largest contentful paint for the page.
+   * The selector of the element corresponding to the largest contentful paint
```
philipwalton, line 930:
```suggestion
   * A selector identifying the element corresponding to the largest contentful paint
```

For consistency with other JSDoc comments.

src/attribution/onCLS.ts
```diff
@@ -16,14 +16,21 @@
...
+  entryPreProcessingCallbacks,
+} from '../onCLS.js';
 import {
   CLSAttribution,
   CLSMetric,
   CLSMetricWithAttribution,
   ReportOpts,
 } from '../types.js';
 
+// A reference to the layout shift target node incase it is removed before reporting
```
philipwalton, line 30:
```suggestion
// A reference to the layout shift target node in case it's removed before reporting.
```

src/attribution/onLCP.ts
```diff
@@ -16,14 +16,20 @@
...
+  entryPreProcessingCallbacks,
+} from '../onLCP.js';
 import {
   LCPAttribution,
   LCPMetric,
   LCPMetricWithAttribution,
   ReportOpts,
 } from '../types.js';
 
+// A reference to the LCP Target node incase it is removed before reporting
```
philipwalton, line 30:
```suggestion
// A reference to the LCP target node in case it's removed before reporting.
```

src/types/lcp.ts
```diff
@@ -31,9 +31,14 @@ export interface LCPMetric extends Metric {
  */
 export interface LCPAttribution {
   /**
-   * The element corresponding to the largest contentful paint for the page.
+   * The selector of the element corresponding to the largest contentful paint
```
philipwalton, line 34:
```suggestion
   * A selector identifying the element corresponding to the largest contentful paint
```

src/types/lcp.ts
```diff
@@ -31,9 +31,14 @@ export interface LCPMetric extends Metric {
...
+   * for the page.
    */
   element?: string;
+  /**
+   * The element corresponding to the largest contentful paint for the page.
+   */
+  elementNode?: Node;
```
philipwalton, line 41: I understand the desire to keep the API for `element` the same, but it does feel weird to call this a "node" when it's always going to be an element (and when the comment itself calls it an "element").

What do you think about using `target` and `targetElement` instead (for consistency with other interfaces)? I know that would be a breaking change, but a major version is the time to make such a change...

tunetheweb: Yeah I think it's time to take the plunge on this one! Done.

Though, personally I never liked the name "target" outside of interactions (which you "target" with your mouse or keyboard). But oh well. Consistency is more important.

tunetheweb:
@dmitiiv there's a request not to save the full element (see #580) but instead just the target due to the amount of memory that can be taken up. This makes reporting the element back kinda pointless (since it can be got from the lcp entry).

Would the selector be sufficient for you without the element?

dmitiiv:
Yes, I've seen the issue. Removing the element (without saving) makes the fix almost useless for the purpose of obtaining information about the target element in a Single Page Application (SPA). It is critical to keep statistics without losing the details of the problem. However, if we still have an element selector, it could allow us to safely retain the necessary information. 
Briefly, yes its not that was expected but better if we don't have even a selector

tunetheweb:
Yes we would still look to save the element selector (and in fact this would be the better selector since it would be calculated while the element is still part of the DOM, rather than once it's been removed).

src/types/lcp.ts
tunetheweb: Reverting this, as no no longer saving the Node (so no need to report the node separately from the entry node).

philipwalton:
There are a number of tradeoffs being made here, balancing different needs from various users of this library.

I think the root of all these issues is that currently the library has its own internal `getSelector()` function that users cannot modify, so some users have asked for the library to [expose the underlying DOM element](https://github.com/GoogleChrome/web-vitals/pull/479) as a solution. This gives users more flexibility, but it comes at a cost of requiring the library to store strong references to these DOM elements for all users, even users that don't need this functionality (because the default selector logic works fine for them).

I wonder if instead of the library making decisions about these tradeoffs, it would be better to give users full control, in the form of letting them customize the target selector generation logic.

Perhaps something like an option to pass their own `getTargetSelector` function:

```
onLCP(metricCallback, { 
  getTargetSelector: (element) => {
    // Users can add whatever logic they want to generate a selector,
    // some other identifier, or even just return the element itself.
    return myCustomStringifyLogic(element);
  }
});
```

This option would be available for all metric functions in the attribution build.

* * *

Regarding whether to run the `getSelector()` logic at reporting time or at the time when the entry is dispatched, there are also tradeoffs here. I don't think the `getSelector()` is slow or anything, but it does have to traverse up the DOM tree for each element, which (depending on the size of the DOM and the number of page interactions, layout shifts, etc.) could result in a lot of processing. That's the main reason I originally choose to generate the selector at reporting time.

While we *could* make this change and always calculate at entry-dispatch time, any users who need this behavior can already get it by using the `reportAllChanges` option, so this change would impose the slower behavior on all users, even those who don't need it.

To be clear, I'm not necessarily opposed to this change, I just want to make sure we're thinking about all of the trade offs before making a decision that impacts all users of the library.

tunetheweb:
> Removing the element (without saving) makes the fix almost useless for the purpose of obtaining information about the target element in a Single Page Application (SPA). It is critical to keep statistics without losing the details of the problem. However, if we still have an element selector, it could allow us to safely retain the necessary information.
Briefly, yes its not that was expected but better if we don't have even a selector

@dmitiiv and alternative being proposed in #585 would allow you to generate your own selector, in case you have better details to return from the Node than the ones the library chooses. The function (whether the default `getSelector` function or your custom override) would be run and be saved for later at as soon as the performance observer fires so the element has the best chance of being still in the DOM. Would that suffice? Or is there a reason you'd still prefer the node to be returned? If so, can you explain the use case?

dmitiiv:
@tunetheweb Yes, we process Nodes by extracting HTML and CSS from them for further display in the UI, in order to visually represent performance issues for developers. And of course, it is more convenient to receive the node directly. However, having the ability to generate a selector (getSelector()) will allow us to find a solution. Thank you for the question. If the selector is preserved after the node is deleted, then saving the styles and HTML on our side by the selector and matching them with a specific metric object seems like a workable option.

PR #566, opened 2024-11-07
Add interoperable TTFB and `documentDuration` to measure Early Hints better

### LINKED ISSUE #565: Add interoperable TTFB and ability to measure document duration.
For sites using Early Hints there are two "TTFB"s depending on your definition of "first bytes":

- **When the Early Hint bytes are received**: Many argue these _are_ the "first" bytes of the response.
- **When the actual document bytes are received**: Many argue this has always been the understanding of TTFB and this should not change just because Early Hints have been deployed. We don't count many other responses (TCP, redirects...etc) as "first bytes" so Early Hints should not be counted either.

In Chrome 115, [`firstInterimResponseStart`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/firstInterimResponseStart) was introduced as a way of measuring Early Hints response start, and `responseStart` was moved out to be the the document bytes were delivered.

While done with the best intentions and to address the above issue where the majority agreed with the second definition and wanted to return to this, the moving of `responseStart` turned out to be a bad idea as it [created an interop issue](https://docs.google.com/presentation/d/1BU9vWK7e0LSe-scNwCWk_qPghoYfGm7U3SZBF5TMRIk/edit#slide=id.p) between those browsers that supported `firstInterimResponseStart` and had moved `responseStart` (Chrome 115+) and those that had not made this change yet (everyone else, including Safari, Firefox, Chrome < 115, tooling like Lighthouse, CrUX...etc).

We [recently agreed in the WebPerf WG](https://github.com/w3c/resource-timing/issues/345#issuecomment-2450475244) to resolve the interop issue by revering `responseStart` to it's older definition (so first bytes, whether Early Hints or document) and adding a `finalResponseHeadersStart` timing to measure the document first bytes.

Chrome intends to [implement this change from Chrome 132](https://chromestatus.com/feature/5158830722514944) (behind a flag initially and then by default from 133).

We should update the `web-vitals` library to:
1. consolidate on the interop-compatible definition of TTFB (first bytes, whether Early Hints or document). This can be done ahead of the change in Chrome 132.
2. add support for "Document Duration" to TTFB attribution, for those that want to measure this time between the two definitions (note this time will be AFTER TTFB so is a little weird to include in TTFB breakdown, but I think this is the best place for it despite that).

As 1. will be a breaking change we will need a major release. As we have one in flight at the moment (v5), I'm aiming to add it to that rather than wait for v6, despite the fact a release candidate has already been released.

tunetheweb:
Fixes #565

### DIFF
--- a/README.md
+++ b/README.md
@@ -1010,6 +1010,13 @@ export interface TTFBAttribution {
    */
   requestDuration: number;
   /**
+   * The total time from the first byte of the response was received. Until the
+   * first byte of the document was received. This will only be non-zero for
+   * servers using Early Hints and where browsers support sending this additional
+   * timing. This time is after the TTFB metric.value.
+   */
+  documentDuration: number;
+  /**
    * The `navigation` entry of the current page, which is useful for diagnosing
    * general page load issues. This can be used to access `serverTiming` for
    * example: navigationEntry?.serverTiming
--- a/src/onTTFB.ts
+++ b/src/onTTFB.ts
@@ -71,14 +71,19 @@ export const onTTFB = (
     const navigationEntry = getNavigationEntry();
 
     if (navigationEntry) {
+      // Form Chrome 115 until Chrome 132 (with flags), Chrome reported
+      // responseStart as the document bytes, rather than Early Hint bytes.
+      // Prefer the Early Hint bytes (firstInterimResponseStart) for
+      // consistency with other browers, if non-zero
+      const responseStart =
+        navigationEntry.firstInterimResponseStart ||
+        navigationEntry.responseStart;
+
       // The activationStart reference is used because TTFB should be
       // relative to page activation rather than navigation start if the
       // page was prerendered. But in cases where `activationStart` occurs
       // after the first byte is received, this time should be clamped at 0.
-      metric.value = Math.max(
-        navigationEntry.responseStart - getActivationStart(),
-        0,
-      );
+      metric.value = Math.max(responseStart - getActivationStart(), 0);
 
       metric.entries = [navigationEntry];
       report(true);
--- a/src/types.ts
+++ b/src/types.ts
@@ -53,9 +53,12 @@ declare global {
     durationThreshold?: number;
   }
 
-  // https://wicg.github.io/nav-speculation/prerendering.html#performance-navigation-timing-extension
   interface PerformanceNavigationTiming {
+    // https://wicg.github.io/nav-speculation/prerendering.html#performance-navigation-timing-extension
     activationStart?: number;
+    // Early Hints support
+    firstInterimResponseStart?: number;
+    finalResponseHeadersStart?: number;
   }
 
   // https://wicg.github.io/event-timing/#sec-performance-event-timing
--- a/test/views/ttfb.njk
+++ b/test/views/ttfb.njk
@@ -27,11 +27,37 @@
   <script>
     // Set the blocking values based on query params if present.
     const params = new URLSearchParams(location.search);
+    const navEntry = performance.getEntriesByType('navigation')[0];
 
     if (params.has('responseStart')) {
-      const navEntry = performance.getEntriesByType('navigation')[0];
       Object.defineProperty(navEntry, 'responseStart', {
         value: Number(params.get('responseStart')),
+      });
+    }
+
+    function block(blockingTime) {
+      const startTime = performance.now();
+      while (performance.now() < startTime + blockingTime) {
+        // Block...
+      }
+    }
+
+    if (params.has('earlyHintsDelay')) {
+      const earlyHintsDelay = Number(params.get('earlyHintsDelay'))
+      const responseStart = navEntry.responseStart;
+      // Block for delay time—to avoid the library seeing future timestamps,
+      // and so not reporting a TTFB at all as it's invalid.
+      block(Number(params.get('earlyHintsDelay')));
+      Object.defineProperties(navEntry, {
+        'firstInterimResponseStart': {
+          value: Number(responseStart),
+        },
+        'finalResponseHeadersStart': {
+          value: Number(responseStart + earlyHintsDelay),
+        },
+        'responseStart': {
+          value: Number(responseStart + earlyHintsDelay),
+        }
       });
     }
   </script>
--- a/src/attribution/onTTFB.ts
+++ b/src/attribution/onTTFB.ts
@@ -30,6 +30,7 @@ const attributeTTFB = (metric: TTFBMetric): TTFBMetricWithAttribution => {
     dnsDuration: 0,
     connectionDuration: 0,
     requestDuration: 0,
+    documentDuration: 0,
   };
 
   if (metric.entries.length) {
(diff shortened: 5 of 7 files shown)


README.md
```diff
@@ -1009,6 +1009,13 @@ export interface TTFBAttribution {
    * processing time.
    */
   requestDuration: number;
+  /**
+   * The total time from the first byte of the response was received. Until the
+   * first byte of the document was received. This will only be non-zero for
+   * servers using Early Hints and where browsers support sending this additional
+   * timing. This time is after the TTFB metric.value.
+   */
+  documentDuration: number;
```
yoavweiss, line 1018: isn't that the first byte of the headers of the document?

tunetheweb: It's a duration. So the time between EH first bytes and document first bytes.

But yes it's basically:
`finalResponseHeadersStart` - `firstInterimResponseStart`
(with some extra handling for Chrome 115 - 132)

yoavweiss: so it's the time between EH first byte and the document's *headers* first byte, right?

tunetheweb: Correct. `documentHeadersDuration` or `documentResponseDuration` was a little long winded. But I'm up for a better name if we have it!

yoavweiss: I think the name is fine, but the docs may have room for clarifications

tunetheweb: WDYT:

```
  /**
   * The total time, after TTFB, after which the document started to be
   * received. This will only be non-zero for servers using Early Hints and
   * where browsers support sending this additional timing. This is the time
   * between the Early Hints first response (TTFB) and when the actual document
   * response started with its HTTP headers.
   */
```

yoavweiss: I'd go with "document's HTTP headers" in the first line, instead of "document". Otherwise, looks great!

Montiwa11:
85cb9f0ab8de1f3768afa69581f4ab6a8e88fac1

Montiwa11 (approved):

test/views/ttfb.njk
```diff
@@ -27,13 +27,54 @@
...
     }
+
+    function block(blockingTime) {
+      const startTime = performance.now();
+      while (performance.now() < startTime + blockingTime) {
+        // Block...
+      }
+    }
+
+    if (params.has('earlyHintsDelay')) {
```
philipwalton, line 47: Instead of stubbing this on the client, why not implement it on the server? Node.js [supports early hints](https://nodejs.org/api/http.html#responsewriteearlyhintshints-callback), and I was able to get it to work pretty easily by using the `response.writeEarlyHints()` method in the `server.js` file (which uses Express, but Express is just sugar on top of Node's HTTP server implementation, so every Express `Request` and `Response` objects are actually node `Request` and `Response` objects).

philipwalton (changes requested):

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - **[BREAKING]** Remove the deprecated `onFID()` function ([#519](https://github.com/GoogleChrome/web-vitals/pull/519))
 - **[BREAKING]** Change browser support policy to Baseline Widely Available ([#525](https://github.com/GoogleChrome/web-vitals/pull/525))
 - **[BREAKING]** Sort the classes that appear in attribution selectors to reduce cardinality ([#518](https://github.com/GoogleChrome/web-vitals/pull/518))
+- **[BREAKING]** Add interoperable TTFB to measure Early Hints consistently ([#566](https://github.com/GoogleChrome/web-vitals/pull/566))
```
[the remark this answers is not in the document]
tunetheweb: Move.

> BTW, I typically wait to update the CHANGELOG until right before releasing a version.

Yeah probably should do that. Just wanna make sure we don't miss anything!

src/attribution/onLCP.ts
```diff
@@ -45,7 +45,15 @@ const attributeLCP = (metric: LCPMetric): LCPMetricWithAttribution => {
           .getEntriesByType('resource')
           .filter((e) => e.name === lcpEntry.url)[0];
 
-      const ttfb = Math.max(0, navigationEntry.responseStart - activationStart);
+      const ttfb = Math.max(
+        0,
+        // From Chrome 115 until, Chrome reported responseStart as the document
```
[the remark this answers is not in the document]
tunetheweb: Yup, was waiting confirmation. Will chase in 133. Updated to that now too.

src/attribution/onLCP.ts
```diff
@@ -45,7 +45,15 @@ const attributeLCP = (metric: LCPMetric): LCPMetricWithAttribution => {
...
+        // bytes, rather than Early Hint bytes. Prefer the Early Hint bytes
+        // (firstInterimResponseStart) for consistency with other browers, if
+        // non-zero
+        navigationEntry.firstInterimResponseStart ||
```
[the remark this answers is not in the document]
tunetheweb: I think it **will** make a difference and will be incorrect if we do this suggestion. `firstInterimResponseStart` is 0 when there are no Early Hints so we don't want to use this. That's what the comment on line 53 is about. Will make it more explicit.

src/onTTFB.ts
```diff
@@ -71,14 +71,19 @@ export const onTTFB = (
     const navigationEntry = getNavigationEntry();
 
     if (navigationEntry) {
+      // From Chrome 115 until, Chrome reported responseStart as the document
+      // bytes, rather than Early Hint bytes. Prefer the Early Hint bytes
+      // (firstInterimResponseStart) for consistency with other browers, if
+      // non-zero
+      const responseStart =
+        navigationEntry.firstInterimResponseStart ||
```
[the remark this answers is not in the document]
tunetheweb: Again we shouldn't do this as we want to ignore `0`s.

src/types.ts
```diff
@@ -53,9 +53,12 @@ declare global {
     durationThreshold?: number;
   }
 
-  // https://wicg.github.io/nav-speculation/prerendering.html#performance-navigation-timing-extension
   interface PerformanceNavigationTiming {
+    // https://wicg.github.io/nav-speculation/prerendering.html#performance-navigation-timing-extension
     activationStart?: number;
+    // Early Hints support
```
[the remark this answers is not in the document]
tunetheweb: Updated to https://w3c.github.io/resource-timing/#dom-performanceresourcetiming-firstinterimresponsestart

test/views/ttfb.njk
tunetheweb: Oh that was a lot easier! Thanks.

tunetheweb:
@philipwalton WDYT about this one?

I know we agreed not to add a `documentStart` timeline (which has been removed from this PR) since that's easily getable by looking at `finalHeadersResponseStart` (about to be launched in Chrome 133), so we _could_ not handle Chrome 115-133 differently since that's due to launch, but normally we try to support things for two years before we consider them no longer required to work around, so should we still merge this?

At the very least I think the Early Hints test is worth saving, but can move that to a separate PR. But it would need to wait for Chrome 133 to land if we don't merge this.

tunetheweb:
Closing this in favour of #609 since Chrome 133 is now well rolled out so seems little point in adding code to support browsers that will be less and less used over time.

PR #574, opened 2024-12-06
LoAF Summary attribution information

### LINKED ISSUE #559: Add more LoAF diagnostics to the INP attribution object
Currently, web-vitals provides the raw LoAF entries in the INP attribution object, which consumers of the library need to parse themselves to extract useful diagnostic info about the interaction. To make the attribution object more readily consumable, web-vitals should pre-process a core set of useful LoAF-based diagnostics.

After experimenting with different INP diagnostics, this is the subset that I've found to be most useful, including some existing metrics:

- Overview of how time during the interaction was spent
  - [x] **INP duration**
  - [x] **INP phase durations**
  - [x] **Interaction time**
  - [ ] Total style and layout duration
  - [ ] Total forced style and layout duration
  - [ ] **Script duration grouped by invoker type and INP phase**
- Details about the slowest script entry during the interaction
  - [ ] **Corresponding INP phase**
  - [ ] **Total duration**
  - [ ] Compile duration
  - [ ] Execution duration
  - [ ] Forced style and layout duration
  - [ ] Pause duration
  - [ ] **Invoker type**
  - [ ] **Invoker name**
  - [ ] **Source URL**
  - [ ] Source function
  - [ ] Source character position
- Details about the interaction itself
  - [x] **Interaction target**
  - [x] **Interaction type**
- Additional metadata
  - [ ] Number of LoAF entries
  - [ ] Number of LoAF script entries

Diagnostics in bold are the ones I think every site owner should be looking at, with the rest considered nice to have. All of the unchecked items represent the ones I'm proposing we add first-class support for in web-vitals. These could be added directly to the attribution object, or contained within some nested object like `loafAttribution`.

web-vitals should only report the durations that occur during the interaction. A script could already be running at the moment a user interacts, but the only relevant time for attribution purposes is the time since the interaction started. Similarly, any scripts that have already finished executing before the interaction should be completely ignored from all INP diagnostics, even if they happen to occur within the same LoAF entry as the interaction.

The attribution object might look something like this:

```js
inp.attribution = {
  loafAttribution: {
    totalLoafEntries: 1,
    totalLoafScriptEntries: 2,
    totalStyleAndLayoutDuration: 30,
    totalForcedStyleAndLayoutDuration: 50,
    totalDurationsPerPhase: {
      inputDelay: {
        'classic-script': 100
      },
      processingDuration: {
        'event-listener': 200
      }
    },
    slowestScript: {
      phase: 'processingDuration',
      duration: 200,
      compileDuration: 10,
      executionDuration: 190,
      forcedStyleAndLayoutDuration: 50,
      pauseDuration: 0,
      invokerType: 'event-listener',
      invoker: 'BUTTON.onclick',
      sourceURL: 'https://example.com/script.js',
      sourceFunctionName: null,
      sourceCharPosition: -1,
      entry: [PerformanceScriptTiming object]
    }
  }
  ...
}
```

tunetheweb:
Fixes #559

Includes object like this:

```js
"longAnimationFrameSummary": {
    "numLongAnimationFrames": 3,
    "numIntersectingScripts": 9,
    "slowestScript": {
      "entry": // Not Shown,
      "subpart": "processingDuration",
      "intersectingDuration": 395,
      "totalDuration": 395,
      "compileDuration": 0,
      "executionDuration": 395,
      "forcedStyleAndLayoutDuration": 144,
      "pauseDuration": 0,
      "invokerType": "event-listener",
      "invoker": "BUTTON.onclick",
      "sourceURL": "",
      "sourceFunctionName": "",
      "sourceCharPosition": -1,
    },
    "totalDurationsPerSubpart": {
        "inputDelay": {
            "user-callback": 28,
            "event-listener": 185
        },
        "processingDuration": {
            "event-listener": 434
        }
    },
    "totalForcedStyleAndLayoutDuration": 144,
    "totalNonForcedStyleAndLayoutDuration": 3144099.700000003,
    "totalScriptDuration": 647
}
```


src/attribution/onINP.ts
```diff
@@ -233,6 +234,78 @@ const getIntersectingLoAFs = (
...
+
+  attribution.longAnimationFrameEntries.forEach((loafEntry) => {
+    totalStyleAndLayout +=
+      loafEntry.startTime + loafEntry.duration - loafEntry.styleAndLayoutStart;
+    loafEntry.scripts.forEach((script) => {
+      const scriptEndTime = script.startTime + script.duration;
+      if (scriptEndTime < interactionTime) {
+        return;
+      }
+      totalScriptTime += script.duration;
```
rviscomi, line 264: Repeating from my other comment, I don't think total script time is needed. But if we did expose it, it should only include the intersection durations.

src/attribution/onINP.ts
```diff
@@ -233,6 +234,78 @@ const getIntersectingLoAFs = (
...
+      numScripts++;
+      totalForcedStyleAndLayout += script.forcedStyleAndLayoutDuration;
+      const blockingDuration =
+        scriptEndTime - Math.max(interactionTime, script.startTime);
+      const invokerType = script.invokerType; //.replace('-', '_');
```
rviscomi, line 269:
```suggestion
      const invokerType = script.invokerType;
```

src/attribution/onINP.ts
```diff
@@ -233,6 +234,78 @@ const getIntersectingLoAFs = (
...
+      loafEntry.startTime + loafEntry.duration - loafEntry.styleAndLayoutStart;
+    loafEntry.scripts.forEach((script) => {
+      const scriptEndTime = script.startTime + script.duration;
+      if (scriptEndTime < interactionTime) {
+        return;
+      }
+      totalScriptTime += script.duration;
+      numScripts++;
+      totalForcedStyleAndLayout += script.forcedStyleAndLayoutDuration;
+      const blockingDuration =
```
rviscomi, line 267: Since we decided not to call this `blockingDuration` externally, we should probably find a less ambiguous name to use internally too.

src/attribution/onINP.ts
```diff
@@ -233,6 +234,78 @@ const getIntersectingLoAFs = (
...
+        scriptEndTime - Math.max(interactionTime, script.startTime);
+      const invokerType = script.invokerType; //.replace('-', '_');
+      let phase = 'processingDuration';
+      if (script.startTime < interactionTime + inputDelay) {
+        phase = 'inputDelay';
+      } else if (
+        script.startTime >
+        interactionTime + inputDelay + processingDuration
+      ) {
+        phase = 'presentation';
```
rviscomi, line 277: Does the type checker really not fail on this? Probably a good enough reason to use an enum instead.

```suggestion
        phase = 'presentationDelay';
```

src/attribution/onINP.ts
```diff
@@ -233,6 +234,78 @@ const getIntersectingLoAFs = (
...
+        return;
+      }
+      totalScriptTime += script.duration;
+      numScripts++;
+      totalForcedStyleAndLayout += script.forcedStyleAndLayoutDuration;
+      const blockingDuration =
+        scriptEndTime - Math.max(interactionTime, script.startTime);
+      const invokerType = script.invokerType; //.replace('-', '_');
+      let phase = 'processingDuration';
+      if (script.startTime < interactionTime + inputDelay) {
```
rviscomi, line 271: Personal preference but I like to make the order of operations more explicitly clear
```suggestion
      if (script.startTime < (interactionTime + inputDelay)) {
```

src/attribution/onINP.ts
```diff
@@ -233,6 +234,78 @@ const getIntersectingLoAFs = (
...
+        phase = 'inputDelay';
+      } else if (
+        script.startTime >
+        interactionTime + inputDelay + processingDuration
```
rviscomi, line 275:
```suggestion
        (interactionTime + inputDelay + processingDuration)
```

src/attribution/onINP.ts
```diff
@@ -233,6 +234,78 @@ const getIntersectingLoAFs = (
...
+    totalStyleAndLayout +=
+      loafEntry.startTime + loafEntry.duration - loafEntry.styleAndLayoutStart;
+    loafEntry.scripts.forEach((script) => {
+      const scriptEndTime = script.startTime + script.duration;
+      if (scriptEndTime < interactionTime) {
+        return;
+      }
+      totalScriptTime += script.duration;
+      numScripts++;
+      totalForcedStyleAndLayout += script.forcedStyleAndLayoutDuration;
```
rviscomi, line 266: I think there's an open question whether we should only be taking the blocking/intersecting portion of _all_ duration metrics. WDYT?

src/attribution/onINP.ts
```diff
@@ -233,6 +234,78 @@ const getIntersectingLoAFs = (
...
+        interactionTime + inputDelay + processingDuration
+      ) {
+        phase = 'presentation';
+      }
+      if (!(phase in phases)) {
+        phases[phase] = {};
+      }
+      if (!(invokerType in phases[phase])) {
+        phases[phase][invokerType] = 0;
+      }
```
rviscomi, line 284: I think this is Baseline Widely available, right?
```suggestion
      phases[phase] ??= {};
      phases[phase][invokerType] ??= 0;
```

src/attribution/onINP.ts
```diff
@@ -233,6 +234,78 @@ const getIntersectingLoAFs = (
...
+      phases[phase][invokerType] += blockingDuration;
+
+      if (blockingDuration > slowestScriptDuration) {
+        slowestScript = script;
+        slowestScriptPhase = phase;
+        slowestScriptDuration = blockingDuration;
```
rviscomi, line 290: This is unused

src/attribution/onINP.ts
```diff
@@ -233,6 +234,78 @@ const getIntersectingLoAFs = (
...
 };
 
+const getLoAFSummary = (attribution: INPAttribution) => {
+  const loafAttribution: LongAnimationFrameSummary = {};
+
+  // Stats across all LoAF entries and scripts.
+  const interactionTime = attribution.interactionTime;
+  const inputDelay = attribution.inputDelay;
+  const processingDuration = attribution.processingDuration;
+  let totalStyleAndLayout = 0;
```
rviscomi, line 244: Renaming for consistency with the external API (also need to update the references)
```suggestion
  let totalNonForcedStyleAndLayout = 0;
```

README.md
```diff
@@ -922,6 +931,62 @@ interface INPAttribution {
...
+    | 'inputDelay'
+    | 'processingDuration'
+    | 'presentationDelay';
+  /**
+   * The total blocking durations in each phase by invoker for scripts that
+   * intersect the INP event.
+   */
+  totalDurationsPerPhase?: Record<
+    'inputDelay' | 'processingDuration' | 'presentationDelay',
+    Record<string, number>
```
rviscomi, line 969: I assume this was kept as a string to leave the option open to group by things other than invokers, otherwise we could use the more specific type
```suggestion
    Record<ScriptInvokerType, number>
```

rviscomi (commented):
Thanks Barry! Left a few comments.

README.md
```diff
@@ -922,6 +931,62 @@ interface INPAttribution {
...
+   * The total non-force (i.e. end-of-frame) style and layout duration from any
+   * Long Animation Frames intercepting INP event.
+   */
+  totalNonForcedStyleAndLayoutDuration?: number;
+  /**
+   * The total duration of Long Animation Frame scripts that intersect the INP
+   * duration. Note, this includes forced style and layout within those
+   * scripts.
+   */
+  totalScriptDuration?: number;
```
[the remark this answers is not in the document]
tunetheweb: As noted in https://github.com/GoogleChrome/web-vitals/issues/559#issuecomment-2524195543 I added this as I thought it might be helpful to differentiate between script and non-script (style and layout) time.

README.md
```diff
@@ -922,6 +931,62 @@ interface INPAttribution {
...
+  /**
+   * The number of Long Animation Frame scripts that intersect the INP event.
+   * NOTE: This may be be less than the total count of scripts in the Long
+   * Animation Frames as some scripts may occur before the interaction.
+   */
+  numScripts?: number;
+  /**
+   * The slowest Long Animation Frame script that intersects the INP event.
+   */
+  slowestScript?: PerformanceScriptTiming;
```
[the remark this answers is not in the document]
tunetheweb: I was looking at your implementation on web.dev, where these are very simple look ups or 1 (or in a few cases 2) values.

We could possibly limit the times interesting to just the INP event (e.g. if clicking halfway through the script execution, then only show half the time) but that could also be confusing if the script name and function is reported without the full duration ("why does script X run so much faster for this user, but their INP is higher?"). I think it's a lot clearer to be honest about what this is.

README.md
```diff
@@ -922,6 +931,62 @@ interface INPAttribution {
...
+   * The total blocking durations in each phase by invoker for scripts that
+   * intersect the INP event.
+   */
+  totalDurationsPerPhase?: Record<
+    'inputDelay' | 'processingDuration' | 'presentationDelay',
+    Record<string, number>
+  >;
+  /**
+   * The total forced style and layout durations as provided by Long Animation
+   * Frame scripts intercepting the INP event.
```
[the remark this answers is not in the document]
tunetheweb: Ta. Fixed.

src/attribution/onINP.ts
tunetheweb: Good idea. Done. Can still discuss whether this will be useful or not, but I think it could be.

src/attribution/onINP.ts
tunetheweb: Ta. done.

src/attribution/onINP.ts
tunetheweb: Oh good spot. No it doesn't fail. Will think about enum.

src/attribution/onINP.ts
tunetheweb: Actually no, [the `+` takes precendence over the `>`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_precedence#table) and so prettier (which runs automatically on commit) removes it.

src/attribution/onINP.ts
tunetheweb: Unfortunately we can't for forced style and layout since we don't have the timings LoAF. Maybe that's a good reason to keep this simple and just take it for all intersecting scripts (which is what it does currently)?

src/attribution/onINP.ts
tunetheweb: Looks to be! Updated.

src/attribution/onINP.ts
tunetheweb: Updated.

README.md
tunetheweb: Oh I'd forgotten they'd been explicitly defined in the spec. Updated.

src/attribution/onINP.ts
tunetheweb: It's used 3 lines above to work out whether this is the slowest script.

src/attribution/onINP.ts
tunetheweb: I agree, but sadly prettier (that we enforce) doesn't :-(

src/attribution/onINP.ts
tunetheweb: Changed it to `interestingScriptDuration`

rviscomi: Did you mean `intersectingScriptDuration`?

src/attribution/onINP.ts
rviscomi: This was related to https://github.com/GoogleChrome/web-vitals/pull/574#discussion_r1876248440 about including the slowest script duration in the summary info.

src/attribution/onINP.ts
tunetheweb: Man I have an inability to spell that properly it seems!!!

README.md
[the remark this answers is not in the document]
tunetheweb: Here's an example (more from LoAF than from INP) where the ability to split by script, style&layout (and forced Style and Layout) was thought to be useful: https://webperformance.slack.com/archives/C04BK7K1X/p1733613959706609?thread_ts=1732204038.067829&cid=C04BK7K1X

src/types/inp.ts
```diff
@@ -24,6 +24,129 @@ export interface INPMetric extends Metric {
...
+   * The total non-force (i.e. end-of-frame) style and layout duration from any
+   * Long Animation Frames intersecting INP interaction.
+   */
+  totalNonForcedStyleAndLayoutDuration?: number;
+  /**
+   * The total duration of Long Animation Frame scripts that intersect the INP
+   * duration. Note, this includes forced style and layout within those
+   * scripts.
+   */
+  totalScriptDuration?: number;
```
brendankenny, line 147: can't some of these properties be required and not make the caller check for undefineds? e.g. shouldn't the default for this  one be 0?

src/attribution/onINP.ts
```diff
@@ -233,6 +235,100 @@ const getIntersectingLoAFs = (
...
+  const inputDelay = attribution.inputDelay;
+  const processingDuration = attribution.processingDuration;
+  let totalNonForcedStyleAndLayoutDuration = 0;
+  let totalForcedStyleAndLayout = 0;
+  let totalScriptTime = 0;
+  let numScripts = 0;
+  let slowestScriptDuration = 0;
+  let slowestScriptEntry!: PerformanceScriptTiming;
+  let slowestScriptPhase: string = '';
+  const phases = {} as Record<string, Record<ScriptInvokerType, number>>;
```
brendankenny, line 252:
```suggestion
  const phases: Record<string, Record<ScriptInvokerType, number>> = {};
```

but you may actually want `const phases: Record<string, Partial<Record<ScriptInvokerType, number>>> = {};`

src/attribution/onINP.ts
```diff
@@ -233,6 +235,100 @@ const getIntersectingLoAFs = (
...
+
+  attribution.longAnimationFrameEntries.forEach((loafEntry) => {
```
brendankenny, line 254: switch to for...of loops?

src/attribution/onINP.ts
```diff
@@ -233,6 +235,100 @@ const getIntersectingLoAFs = (
...
+      if (scriptEndTime < interactionTime) {
+        return;
+      }
+      const intersectingScriptDuration =
+        scriptEndTime - Math.max(interactionTime, script.startTime);
+      totalScriptTime += intersectingScriptDuration;
+      numScripts++;
+      totalForcedStyleAndLayout += script.forcedStyleAndLayoutDuration;
+      const invokerType = script.invokerType;
+      let phase = 'processingDuration';
```
brendankenny, line 268: worth typing `phase` to avoid the cast below

src/attribution/onINP.ts
```diff
@@ -233,6 +235,100 @@ const getIntersectingLoAFs = (
...
+        phase = 'inputDelay';
+      } else if (
+        script.startTime >
+        interactionTime + inputDelay + processingDuration
+      ) {
+        phase = 'presentationDelay';
+      }
+
+      // Define the record if necessary
+      phases[phase] ??= {} as Record<ScriptInvokerType, number>;
```
brendankenny, line 279: cast shouldn't be necessary, should be read from `phases` type. I assume this should be `Partial<Record<ScriptInvokerType, number>>` though, as mentioned above?

src/attribution/onINP.ts
```diff
@@ -233,6 +235,100 @@ const getIntersectingLoAFs = (
...
+  loafAttribution.numLongAnimationFrames =
+    attribution.longAnimationFrameEntries.length;
+  loafAttribution.numIntersectingScripts = numScripts;
+  if (slowestScriptEntry !== null) {
+    const slowestScript: SlowestScriptSummary = {
+      entry: slowestScriptEntry,
+      phase: slowestScriptPhase as
+        | 'inputDelay'
+        | 'processingDuration'
+        | 'presentationDelay',
```
brendankenny, line 302: this is actually

```suggestion
      phase: slowestScriptPhase as
        | 'inputDelay'
        | 'processingDuration'
        | 'presentationDelay'
        | '',
```
?

src/attribution/onINP.ts
```diff
@@ -233,6 +235,100 @@ const getIntersectingLoAFs = (
...
+
+    loafAttribution.slowestScript = slowestScript;
+  }
+  loafAttribution.totalDurationsPerPhase = phases;
+  loafAttribution.totalForcedStyleAndLayoutDuration = totalForcedStyleAndLayout;
+  loafAttribution.totalNonForcedStyleAndLayoutDuration =
+    totalNonForcedStyleAndLayoutDuration;
+  loafAttribution.totalScriptDuration = totalScriptTime;
+
+  return loafAttribution;
```
brendankenny, line 329: `loafSummary` for this variable?

brendankenny (commented):
incidentally, it seems like we're leaning `subparts`, not `phases` for LCP, so we might want to do the same thing here (or object quickly on the LCP side of things :)

src/types/inp.ts
tunetheweb: Yes they can.

src/attribution/onINP.ts
tunetheweb: Changed to `Partial` without the `as`.

src/attribution/onINP.ts
tunetheweb: Done.

src/attribution/onINP.ts
tunetheweb: Removed this typing. It should always be defined if a script exists.

src/attribution/onINP.ts
tunetheweb: Done.

tunetheweb:
> incidentally, it seems like we're leaning subparts, not phases for LCP, so we might want to do the same thing here (or object quickly on the LCP side of things :)

I'm cool with subparts so switched to that throughout.

README.md
```diff
@@ -922,6 +931,62 @@ interface INPAttribution {
...
+   * The slowest Long Animation Frame script that intersects the INP event.
+   */
+  slowestScript?: PerformanceScriptTiming;
+  /**
+   * The INP phase where the longest script ran.
+   */
+  slowestScriptPhase?:
+    | 'inputDelay'
+    | 'processingDuration'
+    | 'presentationDelay';
```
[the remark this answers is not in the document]
tunetheweb: Added this type.

rviscomi (approved):
LGTM! I'll try it out in the next RC.

src/attribution/onINP.ts
```diff
@@ -233,6 +235,95 @@ const getIntersectingLoAFs = (
...
+  let slowestScriptEntry!: PerformanceScriptTiming;
+  let slowestScriptSubpart!: INPSubpart;
+  const subparts: Partial<
+    Record<INPSubpart, Partial<Record<ScriptInvokerType, number>>>
+  > = {};
+
+  for (const loafEntry of attribution.longAnimationFrameEntries) {
+    totalNonForcedStyleAndLayoutDuration +=
+      loafEntry.startTime + loafEntry.duration - loafEntry.styleAndLayoutStart;
+    loafEntry.scripts.forEach((script) => {
```
philipwalton, line 257: Can this be a `for..of` loop?

src/attribution/onINP.ts
```diff
@@ -233,6 +235,95 @@ const getIntersectingLoAFs = (
...
+        subpart = 'inputDelay';
+      } else if (
+        script.startTime >
+        interactionTime + inputDelay + processingDuration
+      ) {
+        subpart = 'presentationDelay';
+      }
+
+      // Define the record if necessary. Annoyingly Typescript doesn't yet
+      // recognise this so need a few `!`s on the next two lines to convinced
```
philipwalton, line 279:
```suggestion
      // recognize this so need a few `!`s on the next two lines to convinced
```

src/types/inp.ts
```diff
@@ -24,6 +24,140 @@ export interface INPMetric extends Metric {
   entries: PerformanceEventTiming[];
 }
 
+export type INPSubpart =
+  | 'inputDelay'
```
philipwalton, line 28:
```suggestion
  | 'input-delay'
```

src/types/inp.ts
```diff
@@ -24,6 +24,140 @@ export interface INPMetric extends Metric {
...
+  | 'processingDuration'
```
philipwalton, line 29:
```suggestion
  | 'processing-duration'
```

src/types/inp.ts
```diff
@@ -24,6 +24,140 @@ export interface INPMetric extends Metric {
...
+  | 'presentationDelay';
```
philipwalton, line 30:
```suggestion
  | 'presentation-delay';
```

src/types/inp.ts
```diff
@@ -24,6 +24,140 @@ export interface INPMetric extends Metric {
...
+export interface SlowestScriptSummary {
+  /**
+   * The slowest Long Animation Frame script that intersects the INP
+   * interaction.
+   */
+  entry: PerformanceScriptTiming;
+  /**
+   * The INP sub-part where the longest script ran.
+   */
+  subpart: INPSubpart; //'inputDelay' | 'processingDuration' | 'presentationDelay';
```
philipwalton, line 48:
```suggestion
  subpart: INPSubpart; // 'input-delay' | 'processing-duration' | 'presentation-delay'
```

README.md
```diff
@@ -891,6 +891,125 @@ interface INPAttribution {
...
+        {
+          /**
+           * The slowest Long Animation Frame script that intersects the INP
+           * interaction.
+           */
+          entry: PerformanceScriptTiming;
+          /**
+           * The INP sub-part where the longest script ran.
+           */
+          subpart: INPSubpart; //'inputDelay' | 'processingDuration' | 'presentationDelay';
```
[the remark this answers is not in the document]
tunetheweb: @brendankenny I think it was you that told me were were moving towards subparts [to mirror LCP](https://web.dev/articles/optimize-lcp#optimal_sub-part_times), is that right?

README.md
```diff
@@ -891,6 +891,125 @@ interface INPAttribution {
...
+           * The sourceFunctionName of the slowest script.
+           */
+          sourceFunctionName?: string;
+          /**
+           * The sourceCharPosition of the slowest script.
+           */
+          sourceCharPosition?: number;
+        }
+      /**
+       * The total blocking durations in each sub-part by invoker for scripts that
```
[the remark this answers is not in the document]
tunetheweb: Yeah this started as "blocking duration" meaning of "intersecting duration" which we wanted to remove to avoid confusion with the traditional 50ms blocking definition. Looks like I forgot to update this here. Should be "total intersecting durations"

README.md
```diff
@@ -891,6 +891,125 @@ interface INPAttribution {
...
+      /**
+       * The total forced style and layout durations as provided by Long Animation
+       * Frame scripts intersecting the INP interaction.
+       */
+      totalForcedStyleAndLayoutDuration: number;
+      /**
+       * The total non-force (i.e. end-of-frame) style and layout duration from any
+       * Long Animation Frames intersecting INP interaction.
+       */
+      totalNonForcedStyleAndLayoutDuration: number;
```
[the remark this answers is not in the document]
tunetheweb: Personally I'd definitely rather have the "nonForced" time (i.e. the end of frame style and layout time) separately and maybe total time and then the forced time can we (somewhat) calculated from that. But agree `totalNonForcedStyleAndLayoutDuration` isn't a great name. Open to ideas here.

The problem is the "total forced" time is not really know since it's only made up on scripts > 50ms (see https://github.com/w3c/long-animation-frames/issues/23) so I don't think it's accurate to say the "total force time".

README.md
```diff
@@ -891,6 +891,125 @@ interface INPAttribution {
...
+      /**
+       * The total duration of Long Animation Frame scripts that intersect the INP
+       * duration. Note, this includes forced style and layout within those
+       * scripts.
+       */
+      totalScriptDuration: number;
```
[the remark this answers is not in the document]
tunetheweb: It's also "totalIntersectingScriptDuration" (where script > 5ms).

README.md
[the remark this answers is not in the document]
brendankenny: Yes, we should definitely unify. I suggested it specifically because we didn't want to simultaneously cement "phases" here in web-vitals and "subparts" over in CrUX, and everyone in the LCP discussions agreed on "subpart".

src/attribution/onINP.ts
tunetheweb: Yup.

README.md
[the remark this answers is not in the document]
tunetheweb: Changed to `totalIntersectingScriptsDuration`. Seems a bit long thought!

README.md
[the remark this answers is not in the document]
tunetheweb: Changed to `totalIntersectingScriptsDuration`. Seems a bit long though!

Also added "and is limited to scripts > 5 milliseconds." to the comment.

README.md
```diff
@@ -927,6 +927,125 @@ interface INPAttribution {
...
+      numIntersectingScripts: number;
+      /**
+       * The number of Long Animation Frames intersecting the INP interaction.
+       */
+      numLongAnimationFrames: number;
+      /**
+       * The slowest Long Animation Frame script that intersects the INP
+       * interaction.
+       */
+      slowestScript: SlowestScriptSummary;
```
philipwalton, line 953: Also, I wonder if `longestScript` would be better and more consistent with other API names (e.g. Long Tasks and Long Animation Frames).

test/e2e/onINP-test.js
```diff
@@ -841,6 +841,54 @@ describe('onINP()', async function () {
 
       const [inp1] = await getBeacons();
       assert(inp1.attribution.longAnimationFrameEntries.length > 0);
+      assert(inp1.attribution.longAnimationFrameSummary != {});
```
philipwalton, line 844: This assertion will never fail because the `longAnimationFrameSummary` object will never be the same object as an object you create inline via `{}`.

tunetheweb: Of course. What's the easiest way to test it's not empty. Like this?

```suggestion
      assert(Object.keys(inp1.attribution.longAnimationFrameSummary.length) !== 0);
```

philipwalton: Yeah, but make sure to call `.length` of the result of the `Object.keys()` call.

```suggestion
      assert(Object.keys(inp1.attribution.longAnimationFrameSummary).length !== 0);
```

src/attribution/onINP.ts
```diff
@@ -260,6 +262,99 @@ export const onINP = (
...
+          scriptEndTime - Math.max(interactionTime, script.startTime);
+        totalIntersectingScriptsDuration += intersectingScriptDuration;
+        numScripts++;
+        totalForcedStyleAndLayout += script.forcedStyleAndLayoutDuration;
+        const invokerType = script.invokerType;
+        let subpart: INPSubpart = 'processingDuration';
+        if (script.startTime < interactionTime + inputDelay) {
+          subpart = 'inputDelay';
+        } else if (
+          script.startTime >
```
philipwalton, line 302:
```suggestion
          script.startTime >=
```
While testing this, I noticed cases were rAF scripts were showing up in "processingDuration", and I'm pretty sure this is the culprit.

philipwalton: BTW, I think it'd be possible to add a test to ensure the expected scripts end up in their expected buckets. I'm testing now just using the `inp.njk` test page and adding a few extra rAF calls as well as a blocking, classic script and I believe it should be possible to do all of that in a test.

src/attribution/onINP.ts
```diff
@@ -260,6 +262,99 @@ export const onINP = (
...
+    // Gather the loaf summary information into the loafAttribution object
+    const loafSummary: LongAnimationFrameSummary = {
+      numLongAnimationFrames: attribution.longAnimationFrameEntries.length,
+      numIntersectingScripts: numScripts,
+      slowestScript: {
+        entry: slowestScriptEntry,
+        subpart: slowestScriptSubpart,
+        intersectingDuration: slowestScriptDuration,
+        totalDuration: slowestScriptEntry?.duration,
+        compileDuration:
```
philipwalton, line 333: Using the `?.` optional chaining operator here seems risky because if `slowestScriptEntry` is undefined then you'll end up doing math operations on `undefined`, which will result in `NaN`.

I know there's been some discussion of removing these properties anyway, but if we don't then it's probably better to just omit these (or set them to 0) if there's no slowest script.

818Nawaf (approved):

PR #581, opened 2024-12-16
Change INP target map to selector to reduce memory usage

### LINKED ISSUE #580: Keeping elements around for attribution can leak memory
When we save elements in the node map, they are retained in memory, even after they are removed from the DOM. A DOM element can hold a lot of memory, and also it gets confusing since those retained elements appear in DevTools memory panel.

Perhaps we can retain them as a `WeakRef`, and save the attribution data instead of the node itself when it's removed?

https://issues.chromium.org/issues/376777343 is an example that I believe originates from webvitals inside devtools retaining nodes.

/cc @brendankenny @tunetheweb

tunetheweb:
Fixes #580 

This has a couple of advantages:
- Selector more likely to be useful if taken when element is still in DOM
- Saves memory

And one disadvantage:
- Node may not be available if developers don't want to depend on our selector but instead roll their own and need attributes from the Node.

I think that's an acceptable tradeoff.

### DIFF
--- a/src/attribution/onINP.ts
+++ b/src/attribution/onINP.ts
@@ -73,8 +73,8 @@ const entryToEntriesGroupMap: WeakMap<
   pendingEntriesGroup
 > = new WeakMap();
 
-// A mapping of interactionIds to the target Node.
-export const interactionTargetMap: Map<number, Node> = new Map();
+// A mapping of interactionIds to the target selector.
+export const interactionTargetMap: Map<number, string> = new Map();
 
 // A boolean flag indicating whether or not a cleanup task has been queued.
 let cleanupPending = false;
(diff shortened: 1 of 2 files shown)


tunetheweb:
FYI @noamr

philipwalton:
I wanted to mention that these changes revert the feature added in https://github.com/GoogleChrome/web-vitals/pull/479, which was done at the request of users for whom the exposed selector was not sufficient.

I understand the desire to reduce memory usage, but the entire point of adding the `interactionTargetElement` property was to handle cases where the element was removed from the DOM and thus not easily accessible without re-implementing the INP attribution logic (with similar requests for LCP and CLS).

Instead of reverting this feature, I think it would be good to explore options that allow users who want access to the DOM element to get it, without imposing the memory burden on all users. I discussed such a solution in https://github.com/GoogleChrome/web-vitals/pull/562#issuecomment-2548608218.

tunetheweb:
Closing in favour of #585

Issue #448, opened 2024-03-27
Allow to reset the INP calculation


Mattstir:
Hi,

it would be awesome if the package would offer a `resetINP` function or something similar.

**Use case**
We would need this feature as we take a look at the "subpages" of SPAs as individual pages and would be highly interested in the individual INP values of them. Because of the way the entries are collected in `longestInteractionList` this of course is not really possible.

**Possible Solution**
I guess clearing `prevInteractionCount, prevInteractionCount, longestInteractionMap` in this new reset function should be enough? Or do I miss something here?

**Value**
This clear function would allow for more different use cases (in combination with the already present `reportAllChanges: true`) while increasing the complexity and package size only slightly. I think this is a valuable addition!
Thanks!

tunetheweb:
Hi @Mattstir , we're not looking to add this functionality to this library at this time for the reasons explained here: https://github.com/GoogleChrome/web-vitals/pull/433#issuecomment-1961430051

> However, while we appreciate the potential usefulness of this change to allow SPAs to manually measure CLS and INP per route, we do have some concerns with accepting this into the web-vitals library:
> 
> First up, the intent of the library (as given in the [Overview](https://github.com/GoogleChrome/web-vitals?tab=readme-ov-file#overview)) is for:
> 
> measuring all the [Web Vitals](https://web.dev/articles/vitals) metrics on real users, in a way that accurately matches how they're measured by Chrome and reported to other Google tools (e.g. [Chrome User Experience Report](https://developers.google.com/web/tools/chrome-user-experience-report), [Page Speed Insights](https://developers.google.com/speed/pagespeed/insights/), [Search Console's Speed Report](https://webmasters.googleblog.com/2019/11/search-console-speed-report.html)).
> 
> By reseting CLS and INP you will affect the overall "hard-nav" CLS and INP that can be reported.
> 
> This is also part of why the soft branch support has not been merged either.
> 
> Secondly, the "measuring all the Web Vitals metrics" is a key point. By only making CLS and INP available (which I agree is all that CAN be made available given Chrome itself will not reset LCP or FCP), we think this will lead to confusion, or a certain expectation that they will be available.
> 

But let's leave this issue open so others can comment.

FYI: @spetroll

Mattstir:
Hi @tunetheweb 
Thanks for the quick answer. 
And despite that's not what I was hopping for I fully understand that reasoning.

PS: Thanks for providing this useful package! :)

szymekz:
[...]

PR #587, opened 2025-01-29
[soft-navigation] expose reset for cls and inp on navigation

mattkubej:
## Description

This PR exposes `reset` functions for the return values of `onCLS` and `onINP`. These functions allow for resetting the metric measurements, particularly useful in single-page applications (SPAs) with soft navigation.

## Motivation

Surface web-vitals performance signals on soft-navigation within SPAs. 

[Soft-navigation](https://developer.chrome.com/docs/web-platform/soft-navigations-experiment) is an experimental feature within Chrome, but we can instrument SPAs to observe signals during soft-navigation. Resetting the observability of CLS and INP on soft-navigation, as signaled by the client, will provide us additional context on the performance of soft-navigations.

## How to use

 ```javascript
 const { reset: resetCLS } = onCLS(handleCLS, { reportAllChanges: true });
 const { reset: resetINP } = onINP(handleINP, { reportAllChanges: true });

function onSoftNavigation() {
  resetCLS();
  resetINP();
}
```

Similar to the [batching example](https://github.com/GoogleChrome/web-vitals?tab=readme-ov-file#batch-multiple-reports-together), clients could collect groups of events for each soft-navigation and report on them.

### DIFF
--- a/src/attribution/onCLS.ts
+++ b/src/attribution/onCLS.ts
@@ -86,7 +86,7 @@ export const onCLS = (
   onReport: (metric: CLSMetricWithAttribution) => void,
   opts?: ReportOpts,
 ) => {
-  unattributedOnCLS((metric: CLSMetric) => {
+  return unattributedOnCLS((metric: CLSMetric) => {
     const metricWithAttribution = attributeCLS(metric);
     onReport(metricWithAttribution);
   }, opts);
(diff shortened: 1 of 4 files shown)


818Nawaf (approved):

tunetheweb:
Hi @mattkubej , thanks for the PR. This has been suggested a few times before, but we are not currently looking to implement this into the library, for the reasons given in https://github.com/GoogleChrome/web-vitals/issues/448#issuecomment-2026242050 and https://github.com/GoogleChrome/web-vitals/issues/448#issuecomment-2407833874.

We're hoping to make more progress on the Chrome side this year, and so the [soft navs experimental support in this library](https://github.com/GoogleChrome/web-vitals/pull/308)

PR #588, opened 2025-01-31
Remove visibilitychange listener after use

Nifdee:
Similar to https://github.com/GoogleChrome/web-vitals/pull/554, we want to remove those event listeners after they've been used.

### DIFF
--- a/src/lib/whenIdle.ts
+++ b/src/lib/whenIdle.ts
@@ -32,7 +32,7 @@ export const whenIdle = (cb: () => void): number => {
     cb();
   } else {
     handle = rIC(cb);
-    onHidden(cb);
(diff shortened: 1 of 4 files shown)


src/lib/whenIdle.ts
```diff
@@ -32,7 +32,7 @@ export const whenIdle = (cb: () => void): number => {
     cb();
   } else {
     handle = rIC(cb);
-    onHidden(cb);
+    onHidden(cb, { once: true });
```
williazz, line 35: Issue: This may not work because the `once` parameter does not take into account if the visibility state was hidden, preventing the callback from being invoked. 

```
if (document.visibilityState === 'hidden') {
      cb();
}
```

PR #592, opened 2025-02-27
Extend INP attribution with extra LoAF information: longest script and buckets

### LINKED ISSUE #559: Add more LoAF diagnostics to the INP attribution object

tunetheweb:
Closes #559 

This is an alternative to #574 which provides some extra attribution information:

- `totalScriptDuration` - the total duration of all intersecting scripts
- `totalStyleAndLayoutDuration` - The total style and layout duration including any end-of-frame style and layout duration plus any forced style and layout duration.
- `totalPaintDuration` -  the off-main-thread presentation delays (end of LoAF -> end of INP).
- `totalUnattributedDuration` - The total unattributed time not included in any of the previous totals including scripts < 5 milliseconds and other timings not attributed by LoAF (including when a frame is < 50ms and so has no LoAF).
- `longestScript` - the longest intersection script
   - `entry` - the entry
   - `subpart` - the subpart the script ocured in
   - `intersectingDuration` - the intersecting duration (as it may be not all of the script intersected the INP time).

The first four give more information on whether this is a script, or style and layout, or frame presentation delay.

The last item pulls out the important script (by intersecting duration, which is not easily obtainable form the current raw LoAF entry).

Example output:

```json
{
    "interactionTarget": "div.devsite-top-logo-row-wrapper>div.devsite-top-logo-row>devsite-appearance-selector",
    "interactionType": "pointer",
    "interactionTime": 25221.30000000447,
    "nextPaintTime": 25477.30000000447,
    "processedEventEntries": ...,
    "longAnimationFrameEntries": ...,
    "inputDelay": 32.599999994039536,
    "processingDuration": 164.20000000298023,
    "presentationDelay": 59.20000000298023,
    "loadState": "complete",
    "longestScript": {
        "entry": {
            "name": "script",
            "entryType": "script",
            "startTime": 25264.89999999851,
            "duration": 141,
            "invoker": "DEVSITE-HEADER.onclick",
            "invokerType": "event-listener",
            "windowAttribution": "self",
            "executionStart": 25264.89999999851,
            "forcedStyleAndLayoutDuration": 141,
            "pauseDuration": 0,
            "sourceURL": "https://www.gstatic.com/devrel-devsite/prod/v6bfb74446ce17cd0d3af9b93bf26e056161cb79c5a6475bd6a9c25286fcb7861/js/devsite_app_module.js",
            "sourceFunctionName": "",
            "sourceCharPosition": -1
        },
        "subpart": "processing-duration",
        "intersectingDuration": 141
    },
    "totalScriptDuration": 15,
    "totalStyleAndLayoutDuration": 156.29999999701977,
    "totalPaintDuration": 43.900000005960464,
    "totalUnattributedDuration": 40.79999999701977
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1005,6 +1005,9 @@ interface INPAttribution {
    * This includes scripts < 5 milliseconds and other timings not attributed
    * by Long Animation Frame (including when a frame is < 50ms and so has no
    * Long Animation Frame).
+   * When no Long Animation Frames are present this will be undefined, rather
+   * than everything being unattributed to make it clearer when it's expected
+   * to be small.
    */
   totalUnattributedDuration?: number;
 }
@@ -1071,7 +1074,7 @@ interface LCPAttribution {
 #### `TTFBAttribution`
 
 ```ts
-export interface TTFBAttribution {
+interface TTFBAttribution {
   /**
    * The total time from when the user initiates loading the page to when the
    * page starts to handle the request. Large values here are typically due
--- a/README.md
+++ b/README.md
@@ -881,29 +881,6 @@ interface FCPAttribution {
 #### `INPAttribution`
 
 ```ts
-/**
- * Summary information about the longest script intersecting the INP duration
- * as provided by the Long Animation Frame API.
- *
- * NOTE: Only scripts above 5 milliseconds are included in long animation
- * frames.
- */
-interface LongestScriptSummary {
-  /**
-   * The longest Long Animation Frame script that intersects the INP
-   * interaction.
-   */
-  entry: PerformanceScriptTiming;
-  /**
-   * The INP sub-part where the longest script ran.
-   */
-  subpart: INPSubpart; //'input-delay' | 'processing-duration' | 'presentation-delay';
-  /**
-   * The amount of time the longest script intersected the INP duration.
-   */
-  intersectingDuration: number;
-}
-
 interface INPAttribution {
   /**
    * By default, a selector identifying the element that the user first
@@ -975,14 +952,15 @@ interface INPAttribution {
    * candidate interaction's `startTime` and the `processingEnd` time of the
    * last event processed within that animation frame. If the browser does not
    * support the Long Animation Frame API or no `long-animation-frame` entries
-   * are detect, this array will be empty.
+   * are detected, this array will be empty.
    */
   longAnimationFrameEntries: PerformanceLongAnimationFrameTiming[];
   /**
-   * The longest Long Animation Frame script that intersects the INP
-   * interaction.
+   * Summary information about the longest script entry intersecting the INP
+   * duration. Note, only script entries above 5 milliseconds are reported by
+   * the Long Animation Frame API.
    */
-  longestScript?: longestScriptSummary;
+  longestScript?: INPLongestScriptSummary;
   /**
    * The total duration of Long Animation Frame scripts that intersect the INP
    * duration excluding any forced style and layout (that is included in
@@ -991,7 +969,7 @@ interface INPAttribution {
   totalScriptDuration?: number;
   /**
    * The total style and layout duration from any Long Animation Frames
-   * intersecting INP interaction. This includes any end-of-frame style and
+   * intersecting the INP interaction. This includes any end-of-frame style and
    * layout duration + any forced style and layout duration.
    */
   totalStyleAndLayoutDuration?: number;
@@ -1010,6 +988,26 @@ interface INPAttribution {
    * to be small.
    */
   totalUnattributedDuration?: number;
+}
+```
+
+#### `INPLongestScriptSummary`
+
+```ts
+interface INPLongestScriptSummary {
+  /**
+   * The longest Long Animation Frame script entry that intersects the INP
+   * interaction.
+   */
+  entry: PerformanceScriptTiming;
+  /**
+   * The INP sub-part where the longest script ran.
+   */
+  subpart: 'input-delay' | 'processing-duration' | 'presentation-delay';
+  /**
+   * The amount of time the longest script intersected the INP duration.
+   */
+  intersectingDuration: number;
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -1001,7 +1001,7 @@ interface INPLongestScriptSummary {
    */
   entry: PerformanceScriptTiming;
   /**
-   * The INP sub-part where the longest script ran.
+   * The INP subpart where the longest script ran.
    */
   subpart: 'input-delay' | 'processing-duration' | 'presentation-delay';
   /**
--- a/README.md
+++ b/README.md
@@ -882,15 +882,15 @@ interface FCPAttribution {
 
 ```ts
 /**
- * Summary information about the slowest script intersecting the INP duration
+ * Summary information about the longest script intersecting the INP duration
  * as provided by the Long Animation Frame API.
  *
  * NOTE: Only scripts above 5 milliseconds are included in long animation
  * frames.
  */
-export interface longestScriptSummary {
+interface LongestScriptSummary {
   /**
-   * The slowest Long Animation Frame script that intersects the INP
+   * The longest Long Animation Frame script that intersects the INP
    * interaction.
    */
   entry: PerformanceScriptTiming;
@@ -899,7 +899,7 @@ export interface longestScriptSummary {
    */
   subpart: INPSubpart; //'input-delay' | 'processing-duration' | 'presentation-delay';
   /**
-   * The amount of time the slowest script intersected the INP duration.
+   * The amount of time the longest script intersected the INP duration.
    */
   intersectingDuration: number;
 }
@@ -979,7 +979,7 @@ interface INPAttribution {
    */
   longAnimationFrameEntries: PerformanceLongAnimationFrameTiming[];
   /**
-   * The slowest Long Animation Frame script that intersects the INP
+   * The longest Long Animation Frame script that intersects the INP
    * interaction.
    */
   longestScript?: longestScriptSummary;
--- a/README.md
+++ b/README.md
@@ -881,6 +881,29 @@ interface FCPAttribution {
 #### `INPAttribution`
 
 ```ts
+/**
+ * Summary information about the slowest script intersecting the INP duration
+ * as provided by the Long Animation Frame API.
+ *
+ * NOTE: Only scripts above 5 milliseconds are included in long animation
+ * frames.
+ */
+export interface longestScriptSummary {
+  /**
+   * The slowest Long Animation Frame script that intersects the INP
+   * interaction.
+   */
+  entry: PerformanceScriptTiming;
+  /**
+   * The INP sub-part where the longest script ran.
+   */
+  subpart: INPSubpart; //'input-delay' | 'processing-duration' | 'presentation-delay';
+  /**
+   * The amount of time the slowest script intersected the INP duration.
+   */
+  intersectingDuration: number;
+}
+
 interface INPAttribution {
   /**
    * By default, a selector identifying the element that the user first
@@ -898,14 +921,6 @@ interface INPAttribution {
    */
   interactionTime: DOMHighResTimeStamp;
   /**
-   * The best-guess timestamp of the next paint after the interaction.
-   * In general, this timestamp is the same as the `startTime + duration` of
-   * the event timing entry. However, since duration values are rounded to the
-   * nearest 8ms (and can be rounded down), this value is clamped to always be
-   * reported after the processing times.
-   */
-  nextPaintTime: DOMHighResTimeStamp;
-  /**
    * The type of interaction, based on the event type of the `event` entry
    * that corresponds to the interaction (i.e. the first `event` entry
    * containing an `interactionId` dispatched in a given animation frame).
@@ -914,19 +929,18 @@ interface INPAttribution {
    */
   interactionType: 'pointer' | 'keyboard';
   /**
+   * The best-guess timestamp of the next paint after the interaction.
+   * In general, this timestamp is the same as the `startTime + duration` of
+   * the event timing entry. However, since duration values are rounded to the
+   * nearest 8ms (and can be rounded down), this value is clamped to always be
+   * reported after the processing times.
+   */
+  nextPaintTime: DOMHighResTimeStamp;
+  /**
    * An array of Event Timing entries that were processed within the same
    * animation frame as the INP candidate interaction.
    */
   processedEventEntries: PerformanceEventTiming[];
-  /**
-   * If the browser supports the Long Animation Frame API, this array will
-   * include any `long-animation-frame` entries that intersect with the INP
-   * candidate interaction's `startTime` and the `processingEnd` time of the
-   * last event processed within that animation frame. If the browser does not
-   * support the Long Animation Frame API or no `long-animation-frame` entries
-   * are detect, this array will be empty.
-   */
-  longAnimationFrameEntries: PerformanceLongAnimationFrameTiming[];
   /**
    * The time from when the user interacted with the page until when the
    * browser was first able to start processing event listeners for that
@@ -955,6 +969,44 @@ interface INPAttribution {
    * (e.g. usually in the `dom-interactive` phase) it can result in long delays.
    */
   loadState: LoadState;
+  /**
+   * If the browser supports the Long Animation Frame API, this array will
+   * include any `long-animation-frame` entries that intersect with the INP
+   * candidate interaction's `startTime` and the `processingEnd` time of the
+   * last event processed within that animation frame. If the browser does not
+   * support the Long Animation Frame API or no `long-animation-frame` entries
+   * are detect, this array will be empty.
+   */
+  longAnimationFrameEntries: PerformanceLongAnimationFrameTiming[];
+  /**
+   * The slowest Long Animation Frame script that intersects the INP
+   * interaction.
+   */
+  longestScript?: longestScriptSummary;
+  /**
+   * The total duration of Long Animation Frame scripts that intersect the INP
+   * duration excluding any forced style and layout (that is included in
+   * totalStyleAndLayout). Note, this is limited to scripts > 5 milliseconds.
+   */
+  totalScriptDuration?: number;
+  /**
+   * The total style and layout duration from any Long Animation Frames
+   * intersecting INP interaction. This includes any end-of-frame style and
+   * layout duration + any forced style and layout duration.
+   */
+  totalStyleAndLayoutDuration?: number;
+  /**
+   * The off main-thread presentation delay from the end of the last Long
+   * Animation Frame (where available) until the INP end point.
+   */
+  totalPaintDuration?: number;
+  /**
+   * The total unattributed time not included in any of the previous totals.
+   * This includes scripts < 5 milliseconds and other timings not attributed
+   * by Long Animation Frame (including when a frame is < 50ms and so has no
+   * Long Animation Frame).
+   */
+  totalUnattributedDuration?: number;
 }
 ```
 
--- a/src/attribution/onINP.ts
+++ b/src/attribution/onINP.ts
@@ -261,7 +261,7 @@ export const onINP = (
     return intersectingLoAFs;
   };
 
-  const attributeLoAFDetails = (attribution: INPAttribution, value: number) => {
+  const attributeLoAFDetails = (attribution: INPAttribution) => {
     // If there is no LoAF data then nothing further to attribute
     if (!attribution.longAnimationFrameEntries?.length) {
       return;
@@ -330,8 +330,7 @@ export const onINP = (
       ? lastLoAF.startTime + lastLoAF.duration
       : 0;
     if (lastLoAFEndTime >= interactionTime + inputDelay + processingDuration) {
-      totalPaintDuration =
-        attribution.interactionTime + value - lastLoAFEndTime;
+      totalPaintDuration = attribution.nextPaintTime - lastLoAFEndTime;
     }
 
     if (longestScriptEntry && longestScriptSubpart) {
@@ -345,7 +344,8 @@ export const onINP = (
     attribution.totalStyleAndLayoutDuration = totalStyleAndLayoutDuration;
     attribution.totalPaintDuration = totalPaintDuration;
     attribution.totalUnattributedDuration =
-      value -
+      interactionTime +
+      attribution.nextPaintTime -
       totalScriptDuration -
       totalStyleAndLayoutDuration -
       totalPaintDuration;
@@ -407,7 +407,7 @@ export const onINP = (
       totalUnattributedDuration: undefined,
     };
 
-    attributeLoAFDetails(attribution, metric.value);
+    attributeLoAFDetails(attribution);
 
     // Use `Object.assign()` to ensure the original metric object is returned.
     const metricWithAttribution: INPMetricWithAttribution = Object.assign(
--- a/src/attribution/onINP.ts
+++ b/src/attribution/onINP.ts
@@ -23,10 +23,10 @@ import {whenIdleOrHidden} from '../lib/whenIdleOrHidden.js';
 import {onINP as unattributedOnINP} from '../onINP.js';
 import {
   INPAttribution,
+  INPAttributionReportOpts,
   INPMetric,
   INPMetricWithAttribution,
-  INPSubpart,
-  INPAttributionReportOpts,
+  INPLongestScriptSummary,
 } from '../types.js';
 
 interface pendingEntriesGroup {
@@ -277,7 +277,7 @@ export const onINP = (
     let totalPaintDuration = 0;
     let longestScriptDuration = 0;
     let longestScriptEntry: PerformanceScriptTiming | undefined;
-    let longestScriptSubpart: INPSubpart | undefined;
+    let longestScriptSubpart: INPLongestScriptSummary['subpart'] | undefined;
 
     for (const loafEntry of attribution.longAnimationFrameEntries) {
       totalStyleAndLayoutDuration =
@@ -295,7 +295,7 @@ export const onINP = (
           scriptEndTime - Math.max(interactionTime, script.startTime);
         // Since forcedStyleAndLayoutDuration doesn't provide timestamps, we
         // apportion the total based on the intersectingScriptDuration. Not
-        // correct depending on when it occured, but the best we can do.
+        // correct depending on when it occurred, but the best we can do.
         const intersectingForceStyleAndLayoutDuration = script.duration
           ? (intersectingScriptDuration / script.duration) *
             script.forcedStyleAndLayoutDuration
@@ -308,7 +308,7 @@ export const onINP = (
         totalStyleAndLayoutDuration += intersectingForceStyleAndLayoutDuration;
 
         if (intersectingScriptDuration > longestScriptDuration) {
-          // Set the subpart this occured in.
+          // Set the subpart this occurred in.
           longestScriptSubpart =
             script.startTime < interactionTime + inputDelay
               ? 'input-delay'
--- a/src/attribution/onINP.ts
+++ b/src/attribution/onINP.ts
@@ -267,21 +267,18 @@ export const onINP = (
     const inputDelay = attribution.inputDelay;
     const processingDuration = attribution.processingDuration;
     let totalScriptDuration = 0;
-    let totalForcedStyleAndLayoutDuration = 0;
+    let totalStyleAndLayoutDuration = 0;
+    let totalPaintDuration = 0;
     let longestScriptDuration = 0;
     let longestScriptEntry!: PerformanceScriptTiming;
     let longestScriptSubpart!: INPSubpart;
 
     for (const loafEntry of attribution.longAnimationFrameEntries) {
-      attribution.styleAndLayoutDuration =
+      totalStyleAndLayoutDuration =
+        totalStyleAndLayoutDuration +
         loafEntry.startTime +
         loafEntry.duration -
         loafEntry.styleAndLayoutStart;
-
-      attribution.framePresentationDelay =
-        attribution.interactionTime +
-        value -
-        (loafEntry.startTime + loafEntry.duration);
 
       for (const script of loafEntry.scripts) {
         const scriptEndTime = script.startTime + script.duration;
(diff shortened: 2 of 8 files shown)


mmocny:
Before review, some input on naming (since you asked!)

---

"thrashing" is only thrashing if there is actually duplication of effort.  Forced style and layout can happen during script for several reasons and often it just eagerly does what would otherwise happened lazily, not actually more work.  So its hard to know if its thrashing without looking.  (Also `scriptThrashing` without style/layout is a weird name).

WDYT:
- `styleAndLayoutInScript`
- `styleAndLayoutInRendering`

Or, prepare for a sort of "sub-parts":

- `scriptsDuration` for the whole, and then subparts:
  - `scriptsStyleAndLayout`
- `renderingDuration` for the whole, and then subparts:
  - `renderingStyleAndLayout`
  - `renderingRafAndObservers` (Not sure about this)


---

RE: framePresentationDelay` -- I like that fine.  It matches the [`PaintTimingMixin`](https://www.w3.org/TR/paint-timing/#sec-PaintTimingMixin) direction that we have `paintTime` and `presentationTime` and so this is the paintTime -> presentationTime gap, and it should be a "delay" not "duration" because it isn't occupying the main thread (much like input delay)

---

RE: `longestScript` and especially `longestScript.subpart` -- Hmmm.  I guess the main value would be to differentiate input delay vs not, though we already get that via sub-parts?

I'm sort of worried that you could have multiple long scripts (such as a long event and input delay).

Given that we have actual INP sub-parts, and access to the loaf entry with the full list of scripts, I'm just not sure about the value of highlighting just a single longest script-- but no specific concerns either.

mmocny:
Oh I think I see now the value of `intersectingDuration`.  I think this could only ever apply to a single script which only partially overlaps first event timeStamp (and so partially affects input delay).

Maybe instead of longestScript with an overlappingDuration we have a specific value for this:
- `scriptWhichAddedSignificantInputDelayBikeshedName`

...which could be null or not and would have extra metadata like "overlappingDuration" or "blockingDurationAdded" or something.  All the other scripts in the LoAF entries after this one would affect the main processing time of the frame?

tunetheweb:
Dammit! I thought I'd found a workaround to avoid confusing with the "total style and layout" not including forced 😔

> WDYT:
>
> - `styleAndLayoutInScript`
> - `styleAndLayoutInRendering`

I'm not loving the inconsistency away from the LoAF `styleAndLayoutStart` to be honest.

I'm still not 100% convinced knowing forcedStyleAndLayout is that useful (and therefore is worth the confusion with end of frame style and layout). Once you know the script is long running, it should be easy to repeat that in devtools and then see it forced style and layout (whether due to thrashing, or just an early style and layout calc). So I think labelling this as just "script" time is maybe OK. So maybe this should just be dropped to avoid that confusion.

As to the by subpart breakdown, I actually experimented with fully subparts (and have this on another branch):

image

In the end I decided that was very verbose. People will really care about all the scripts (in which case look at the LoAF entries directly), or the longest script (in which case you need to consider intersecting durations only so library should help make that easier). Having script details for each subpart seems a bit verbose for the library to expose. So I took an opinionated choice of only showing the longest one, but reporting extra details (like which subpart it was in and the intersecting duration).

Part of my issue with #574 is I found the `totalDurationsPerSubpart` object similarly overly complex.

> Oh I think I see now the value of intersectingDuration. I think this could only ever apply to a single script which only partially overlaps first event timeStamp (and so partially affects input delay).

Yes that's correct. It's to explain when a script duration exceeds the total INP. Or when script is just a small part of INP. Maybe this is niche enough that we can just accept it will be over the duration sometimes and skip it?

> Maybe instead of longestScript with an overlappingDuration we have a specific value for this:
>
> `scriptWhichAddedSignificantInputDelayBikeshedName`
> ...which could be null or not and would have extra metadata like "overlappingDuration" or "blockingDurationAdded" or > something. All the other scripts in the LoAF entries after this one would affect the main processing time of the frame?

Well the way I see it, this new data in this PR allow you to see if your problem is:
- script (including forced style and layout in that script) - and if so the worst one.
- style and layout
- off main thread work (i.e. browser work).

So the longest script is still valuable even if it's not in input delay. Granted you don't need `intersectingDuration` then and can just look at script duration but it will be the same as `intersectingDuration` then anyway so no harm to always look at that.

tunetheweb:
Update `scriptThrashing` to `scriptsForcedStyleAndLayoutDuration`.

philipwalton:
One minor concern I have with adding new "duration" properties to the top-level attribution object is that then it's much harder to know which properties are part of the official INP subparts and which properties are just other duration values not related to the official subparts.

I'm not sure if there's much we can do about that, and anything we do would likely be inconsistent with the subparts in LCP attribution, but I wanted to make the comment anyway and have the discussion.

src/types/inp.ts
```diff
@@ -116,6 +150,26 @@ export interface INPAttribution {
...
+  /**
+   * The total duration of Long Animation Frame scripts that intersect the INP
+   * duration. Note, this is limited to scripts > 5 milliseconds.
+   */
+  scriptsDuration?: number;
+  /**
+   * The total forced style and layout duration from any
+   * Long Animation Frames intersecting INP interaction.
+   */
+  scriptsForcedStyleAndLayoutDuration?: number;
```
philipwalton, line 162: Similar to the above, I wonder if this should be called `totalForcedStyleAndLayoutDuration`.

src/types/inp.ts
```diff
@@ -116,6 +150,26 @@ export interface INPAttribution {
    * (e.g. usually in the `dom-interactive` phase) it can result in long delays.
    */
   loadState: LoadState;
+  /**
+   * The total duration of Long Animation Frame scripts that intersect the INP
+   * duration. Note, this is limited to scripts > 5 milliseconds.
+   */
+  scriptsDuration?: number;
```
philipwalton, line 157: I wonder if this should be called `totalScriptDuration` to make it clear that it could be a sum from more than one script entry.

src/types/inp.ts
```diff
@@ -37,6 +37,34 @@ export interface INPMetric extends Metric {
   entries: PerformanceEventTiming[];
 }
 
+export type INPSubpart =
```
philipwalton, line 40: I still think we should name these `input-delay`, `processing-duration`, and `presentation-delay` to follow the [TAG casing rules](https://w3ctag.github.io/design-principles/#casing-rules).

I get that it would be inconsistent with the property names, but maybe that's less of an issue in this PR compared to the previous PR?

src/types/inp.ts
tunetheweb: Done.

src/types/inp.ts
tunetheweb: Done.

src/types/inp.ts
tunetheweb: Yeah I think it's less of an issue here. Done.

tunetheweb:
> One minor concern I have with adding new "duration" properties to the top-level attribution object is that then it's much harder to know which properties are part of the official INP subparts and which properties are just other duration values not related to the official subparts.
>
> I'm not sure if there's much we can do about that, and anything we do would likely be inconsistent with the subparts in LCP attribution, but I wanted to make the comment anyway and have the discussion.

It's a fair concern. I can think of two options to solve this:

1) We have talked about whether this all should fit under a "LoAF" object (which would also make it clear why it's missing on non-LoAF supporting browsers). I've kind of moved away from that here as I thought it made the attribution object more complex in documentation and the like, but we could bring it back. We always have `longestScriptDuraton`, but that could be come buried more with this. Think my preference is to leave it flat, but not a super strong preference so could be persuaded.

2) We could lean into the "total" more. At the moment `styleAndLayoutDuration` and `framePresentationDelay` are basically the two new properties for your concern. They are currently the last frame entries, as per

> `styleAndLayoutDuration` - the style and layout duration of the final presentation frame (note we do not include other frames if there are multiple LoAFs to avoid double counting work, as that should be counted in "input delay").

They could also be totals for when INP spans more that one frame (so `totalStyleAndLayoutDuration` and `totalFramePresentationDelay`) which would differentiate them from the 3 core durations more. In most cases they would probably be for one entry, but for multi-frame INPs I was trying to decide which to show anyway.

I think there are positives and negatives to both options even with that that:

- Totals across all frames - allows you to split the whole INP time in an alternative way (totalScript/totalStyleAndLayout/totalFramePresentationDelay). Though that could also be confusing as a competitor to the core subparts. Then again, since we've done it for scripts already, maybe it's weird not to do it for the last two? The other issue is you could get a `totalFramePresentationDelay` larger than `presentationDelay` so that would be weird.
- Only final frame - allows you to split `presentationDelay` phase more. In which case they are more `presentationDelay` subparts (and we could name them more like this). But then maybe it's weird that we have total for scripts, but not for non-script durations?

At the moment I'm thinking 2) makes more sense. The third option is to just leave it as is. WDYT?

philipwalton:
> At the moment I'm thinking 2) makes more sense. The third option is to just leave it as is. WDYT?

Yeah, option (2) seems reasonable to me. My first reaction was that any style/layout/presentation work not in the last frame should be part of "input delay", but the more I thought about it the more I think there is value in an "alternate" breakdown, e.g. phases vs. buckets—where the buckets are not necessarily sequential and help you determine whether poor INP is primarily caused by script, style/layout, or off-main-thread presentation.

That said, 

> * The other issue is you could get a `totalFramePresentationDelay` larger than `presentationDelay` so that would be weird.

I'm not super bothered by this—however, I do think it's a bit weird that "presentation delay" in the subparts is not the same thing as "presentation delay" in `totalFramePresentationDelay` (the former includes main thread work and the latter doesn't).

What if we didn't limit this to just presentation and included all off-main-thread work (including off-main thread input delay)? Then we could call it something like `totalOffMainThreadDuration` or `totalOffMainThreadDelay`?

tunetheweb:
But how can we measure off main thread work? The only reason we have the framePresentationDelay is because it’s the time between LoAF end and INP end.

Which btw means we also don’t have this for prior frames. But then we also don’t need it for them either when looking at buckets, as we’d double count as that would overlap the next frame starting.

So framePresentationDelay should stay as it is, without the total I think. Which also means it’s back to being a subpart of presentationDelay so maybe that’s OK?

Alternatively we just exclude framePresentationDelay and leave it as unattributed time—and maybe we should add totalUnattributedTime for that matter?

philipwalton:
> But how can we measure off main thread work? The only reason we have the framePresentationDelay is because it’s the time between LoAF end and INP end.

I think you can use LoAF to determine what was happening when the interaction took place, and if it's not happening during script execution or style/layout, then you can assume it's off-main delay. I'm not sure if that's always 100% true, but with some quick testing it seems to be.

> Which btw means we also don’t have this for prior frames. But then we also don’t need it for them either when looking at buckets, as we’d double count as that would overlap the next frame starting.

For bucketing, I don't think you would double count it—at least not how I was imagining things working. E.g. with bucketing you wouldn't have input delay, so any style/layout or presentation time in the previous frame would contribute to those respective buckets.

> Alternatively we just exclude framePresentationDelay and leave it as unattributed time—and maybe we should add totalUnattributedTime for that matter?

I think this would be a reasonable option as well. And we could document the various things that could end up being "unattributed", e.g. off-main presentation, input delay with no LoAF script attribution, <5ms, etc.

tunetheweb:
> > But how can we measure off main thread work? The only reason we have the framePresentationDelay is because it’s the time between LoAF end and INP end.
> 
> I think you can use LoAF to determine what was happening when the interaction took place, and if it's not happening during script execution or style/layout, then you can assume it's off-main delay. I'm not sure if that's always 100% true, but with some quick testing it seems to be.

Here's an example from just yesterday showing there is on-main-thread unattributed time:

https://webperformance.slack.com/archives/C04BK7K1X/p1741616198612019
https://issues.chromium.org/issues/402028633

> > Which btw means we also don’t have this for prior frames. But then we also don’t need it for them either when looking at buckets, as we’d double count as that would overlap the next frame starting.
> 
> For bucketing, I don't think you would double count it—at least not how I was imagining things working. E.g. with bucketing you wouldn't have input delay, so any style/layout or presentation time in the previous frame would contribute to those respective buckets.

I meant if we could measure off-main thread frame presentation delay for the first LoAF, it would overlap the on-main thread second LoAF. But we can't do measure that anyway so doesn't really matter. So agree it couldn't overlap.

> > Alternatively we just exclude framePresentationDelay and leave it as unattributed time—and maybe we should add totalUnattributedTime for that matter?
> 
> I think this would be a reasonable option as well. And we could document the various things that could end up being "unattributed", e.g. off-main presentation, input delay with no LoAF script attribution, <5ms, etc.

The other option is to bucket end of frame styleAndLayout and frame presentation delay as totalRenderTime so you'd have:
- totalScriptDuration (with a subpart of totalForcedStyleAndLayoutDuration)
- totalRenderDuration (with a subpart of totalStyleAndLayoutDuration and totalFramePresentationDuration if we want)
- totalUnattributedTime (anything else not in above buckets).

philipwalton:
> Here's an example from just yesterday showing there is on-main-thread unattributed time:
> 
> https://webperformance.slack.com/archives/C04BK7K1X/p1741616198612019 https://issues.chromium.org/issues/402028633

Interesting. I'm having trouble reproducing a LoAF for that frame. Are you able to repro it? Is the style block that shows up in the trace just a gap in the LoAF?

> The other option is to bucket end of frame styleAndLayout and frame presentation delay as totalRenderTime so you'd have:

What if instead of trying to solve the presentation delay naming issue, we just didn't have a property for that at all. So we only have the following (all of which include intersecting durations from all reported LoAFs):

- `totalScriptDuration`
- `totalStyleAndLayoutDuration`
- `totalForcedStyleAndLayoutDuration`

TBH, if we did want to have another bucket within the "presentation" subpart, I'd rather have something like `totalRenderingScriptDuration`, which would only include script entries that occur after `renderStart` within each LoAF. I think this would address some of Michal's original concerns that large presentation delays can be caused by main-thread scripts.

tunetheweb:
As discussed off-line, moved this to a bucketing model which matches total INP:

image

mmocny:
I like where this conversation ended up, and I think the final diagram Barry put together is starting to look really good.  (I haven't started to review the patch for details, but will do so after a few questions).

1. I see that the buckets are: Script, S&L, Paint, Unattributed--- but I think paint is also unattributed in LoAF?  Are there just 3 total buckets?  (looking at some of my old diagrams, I think thats where I settled: Yellow, Gray, Purple, with Green being for "devtools only")

2. I think the simplified buckets, without splitting by phases, might be "just right" in the aggregate.  But would we also still report "input delay" (up to first event processing) for historical consistency and just to differentiate actual-interaction-slow from page-TBT-type-issues?  And also perhaps expose an (now-off-main-only) "presentation delay"?

tunetheweb:
To answer your questions:

1. "Paint and composite" in the diagram is from end of LoAF until end of INP. So yes it doesn't include any on-main-thread Paint timing (as that's not in LoAF). Open to ideas on better naming here?
2. So we're keeping the traditional 3-way subparts, but now have an alternative view of buckets as well. This does worry me a little that there's basically two breakdowns now, which is not ideal but I think the alternative of breaking down the subparts into buckets (so potentially 3 x 4 = 12 breakdowns) also has it's downsides.

To me the subparts are our primary recommendation (is it your interaction that's the problem, or general business?), and the buckets give an alternative view as to what's the overall problem (is it primarily scripting, s&l, paint, something we don't know, or a mixture of all three?). It's not quite the full invoker/invoker-type combo as was proposed in #574 but I found that overly complex to grok so hopefully this is, as you say, "just right".

So with this we now kind of have the view you see in DevTools - the subparts is like the Interactions lane with the 3 phases, and the buckets are like the Summary breakdown if you highlight that section of the flame chart. We of course don't have the full flame chart details, but the Longest Script at least gives the entry point of the worst part of that (which is probably where you should concentrate if that's large). And those wanting a fuller picture, to be even closer to the flame chart can use the LoAF entries to get all the scripts if they so desire, but that's overkill for many IMHO. With subparts, buckets, and longest script we think we're striking a balance between enough detail, but not too much complexity.

mmocny:
> Open to ideas on better naming here?

Ah, thanks.  In that case it ~mostly what I wanted: off-main-presentation-delay.  I think that given the history, probably "presentation delay" is not a good name to use, so:

- Do as you suggest and name it for ALL the actual stages of rendering:
  - Looking at RenderingNG docs, simplifying a bit, I think the stages are: Paint, Commit, Raster (+Image Decode), Draw.
  - This can get wordy and in theory can go out of sync with what happens.
  - Interop concerns, too.

- Find a single name and make it stick.  I think the main-thread stages are generally known as "rendering" to folks (though there is some inconsistency), so maybe all the off-thread stages shoul dhave a single common name...
  - From the [HTML spec](https://html.spec.whatwg.org/#update-the-rendering), step 22, nothing jumps out:
    > update the rendering or user interface of doc and its [node navigable](https://html.spec.whatwg.org/#node-navigable) to reflect the current state
  - Perhaps "Rasterization" or "GPU Rendering" or something

> So we're keeping the traditional 3-way subparts,

I think that the four timestamps:
- first-event-timeStamp
- first-event-processing-start
- last-event-processing-end
- presentationTime

Are all useful because it lets you "draw" the interaction.

All the 3 phases can easily be computed from these values, but maybe only the "input delay" should be done by default, as the other buckets just do a better job?  Not sure!

> So with this we now kind of have the view you see in DevTools [...]

+1.

tunetheweb:
> I think that the four timestamps:
> 
> * first-event-timeStamp
> * first-event-processing-start
> * last-event-processing-end
> * presentationTime
> 
> Are all useful because it lets you "draw" the interaction.
> 
> All the 3 phases can easily be computed from these values, but maybe only the "input delay" should be done by default, as the other buckets just do a better job? Not sure!

Alternatively, having the 3 subpart durations avoids breaking backwards compatibility and allows those timings to be easily computed from them anyway! So not seeing the need to change this.

Also not understanding what you mean by "but maybe only the "input delay" should be done by default, as the other buckets just do a better job?"? `totalScriptDuration` will span both input delay, and processing duration and even presentation delay so doesn't tell you if it was a "your event handlers" problem (i.e. "processing duration"), or a before or after problem (i.e. "input delay" or "presentation delay"—which can often be unrelated to the specific interaction and just a sign of general busyness.

mmocny:
I meant, maybe exposed as a unique value by default-- but I missed that we already expose these values so not worth removing.

README.md
```diff
@@ -955,6 +969,44 @@ interface INPAttribution {
...
+   * The total style and layout duration from any Long Animation Frames
+   * intersecting INP interaction. This includes any end-of-frame style and
+   * layout duration + any forced style and layout duration.
+   */
+  totalStyleAndLayoutDuration?: number;
+  /**
+   * The off main-thread presentation delay from the end of the last Long
+   * Animation Frame (where available) until the INP end point.
+   */
+  totalPaintDuration?: number;
```
mmocny, line 1002: Whenever naming a time range is hard, an alternative that Chrome sometimes uses is to just name via Timestamp1ToTimestamp2.  It's wordy, but it works well once you know it.

What about `totalPaintToPresentationDuration`?  That's literally what this is, and will help reduce questions :P

src/attribution/onINP.ts
```diff
@@ -260,6 +261,91 @@ export const onINP = (
     return intersectingLoAFs;
   };
 
+  const attributeLoAFDetails = (attribution: INPAttribution, value: number) => {
```
mmocny, line 264: (Sending first pass comments before reviewing this fyi)

README.md
```diff
@@ -913,20 +928,19 @@ interface INPAttribution {
...
    */
   interactionType: 'pointer' | 'keyboard';
+  /**
+   * The best-guess timestamp of the next paint after the interaction.
+   * In general, this timestamp is the same as the `startTime + duration` of
+   * the event timing entry. However, since duration values are rounded to the
+   * nearest 8ms (and can be rounded down), this value is clamped to always be
+   * reported after the processing times.
+   */
+  nextPaintTime: DOMHighResTimeStamp;
```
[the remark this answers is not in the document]
tunetheweb: So this isn't new and is already included in v4. As we're doing a major release, now is the time to consider changing this. But it's a separate issue to this PR. Can you open a new issue?

README.md
```diff
@@ -955,6 +969,44 @@ interface INPAttribution {
...
+   * Animation Frame (where available) until the INP end point.
+   */
+  totalPaintDuration?: number;
+  /**
+   * The total unattributed time not included in any of the previous totals.
+   * This includes scripts < 5 milliseconds and other timings not attributed
+   * by Long Animation Frame (including when a frame is < 50ms and so has no
+   * Long Animation Frame).
+   */
+  totalUnattributedDuration?: number;
```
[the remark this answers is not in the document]
tunetheweb: Nope. That's exactly what it is and how it's calculated.

README.md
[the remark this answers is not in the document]
philipwalton: FWIW, this name was chosen for consistency with the metric name—Interaction to **Next Paint**. So I wouldn't change this name unless we're also thinking about changing the metric name (which I assume we're not?).

README.md
```diff
@@ -881,6 +881,29 @@ interface FCPAttribution {
 #### `INPAttribution`
 
 ```ts
+/**
+ * Summary information about the slowest script intersecting the INP duration
```
philipwalton, line 885:
```suggestion
 * Summary information about the longest script intersecting the INP duration
```

If we're going to call it "longest" script in the API then I think it should match in the comment.

README.md
```diff
@@ -881,6 +881,29 @@ interface FCPAttribution {
...
+ * as provided by the Long Animation Frame API.
+ *
+ * NOTE: Only scripts above 5 milliseconds are included in long animation
+ * frames.
+ */
+export interface longestScriptSummary {
```
philipwalton, line 891:
```suggestion
interface LongestScriptSummary {
```

Interface names should be capitalized. Also, for the README I've been omitting the `export` keywords, to keep things as simple as possible in the docs.

README.md
```diff
@@ -881,6 +881,29 @@ interface FCPAttribution {
...
+  /**
+   * The slowest Long Animation Frame script that intersects the INP
```
philipwalton, line 893:
```suggestion
   * The longest Long Animation Frame script that intersects the INP
```

README.md
```diff
@@ -881,6 +881,29 @@ interface FCPAttribution {
...
+   * interaction.
+   */
+  entry: PerformanceScriptTiming;
+  /**
+   * The INP sub-part where the longest script ran.
+   */
+  subpart: INPSubpart; //'input-delay' | 'processing-duration' | 'presentation-delay';
+  /**
+   * The amount of time the slowest script intersected the INP duration.
```
philipwalton, line 902:
```suggestion
   * The amount of time the longest script intersected the INP duration.
```

README.md
```diff
@@ -955,6 +969,44 @@ interface INPAttribution {
...
+   * If the browser supports the Long Animation Frame API, this array will
+   * include any `long-animation-frame` entries that intersect with the INP
+   * candidate interaction's `startTime` and the `processingEnd` time of the
+   * last event processed within that animation frame. If the browser does not
+   * support the Long Animation Frame API or no `long-animation-frame` entries
+   * are detect, this array will be empty.
+   */
+  longAnimationFrameEntries: PerformanceLongAnimationFrameTiming[];
+  /**
+   * The slowest Long Animation Frame script that intersects the INP
```
philipwalton, line 982:
```suggestion
   * The longest Long Animation Frame script that intersects the INP
```

README.md
philipwalton: Given that the metric name is "Interaction to Next Paint", I'm concerned that `totalPaintToPresentationDuration` could be reasonably interpreted as time spent after the "paint" of INP.

I get that paint and presentation are different things, but given the metric name, I just wanted to point out that we're already not being consistent (arguably `presentationDelay` should have been called `paintDelay`, but I think that ship has sailed).

README.md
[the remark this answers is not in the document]
philipwalton: Is it worth clarifying that if the browser doesn't support LoAF then this will not be set (as opposed to all of the time being considered "unattributed").

philipwalton (changes requested):
I think it might be worth adding another test that creates a situation where there are two LoAFs, just to make sure that case is handled. (I'm happy to offer some ideas for how to do that if it's unclear.)

README.md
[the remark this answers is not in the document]
tunetheweb: Yeah will add a comment. I think that’s preferable to it all being unattributed as that would skew the unattributed portion on aggregate.

README.md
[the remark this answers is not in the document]
mmocny: Fair points.

`nextPaintPresentationTime` would perhaps have been ideal in hindsight, but its not worth the churn.

README.md
mmocny: Hmmm, valid.

The name of the INP metric was itself meant to be consistent with all the other paint metric names, which also have the paint-on-main-vs-presentation-off-main timing confusion. Especially so now with LoAF+PaintTimingMixin, and with the interop work to make these differences explicit.

It's even more wordy, but what about `totalRenderingEndToPresentationDuration`?

Given the status of other things, `totalPaintDuration` is probably fine-- but I worry that would be confused as being the whole "to next paint" time range, which we're already reporting as "presentation delay".

README.md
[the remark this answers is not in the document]
tunetheweb: Fixed in https://github.com/GoogleChrome/web-vitals/pull/592/commits/1ecdcc01e73d20a41a0b6b12283f524f1f68370f

tunetheweb:
> I think it might be worth adding another test that creates a situation where there are two LoAFs, just to make sure that case is handled. (I'm happy to offer some ideas for how to do that if it's unclear.)

Given the flakiness of LoAFs in CI already at the moment (it's one of the tests that fails most often), and your recent discovery that they aren't always emitted in time without experimental web platform features enabled, I'd prefer to wait until one or both of those are fixes before expanding on this in the test suite

WDYT @philipwalton ?

philipwalton:
> Given the flakiness of LoAFs in CI already at the moment (it's one of the tests that fails most often), and your recent discovery that they aren't always emitted in time without experimental web platform features enabled, I'd prefer to wait until one or both of those are fixes before expanding on this in the test suite
> 
> WDYT @philipwalton ?

Sure, happy to address as a separate PR, but we should manually verify that it works in these cases and then come up with a plan to ensure it's properly tested in the future.

src/types/inp.ts
```diff
@@ -37,6 +37,34 @@ export interface INPMetric extends Metric {
...
+  | 'input-delay'
+  | 'processing-duration'
+  | 'presentation-delay';
+
+/**
+ * Summary information about the slowest script intersecting the INP duration
```
philipwalton, line 46: I'm still seeing "slowest" script used here rather than longest. Given that I think we're happy with what's in the README, can you update this now as well?

tunetheweb: Done!

src/attribution/onINP.ts
```diff
@@ -260,6 +261,96 @@ export const onINP = (
...
+        loafEntry.styleAndLayoutStart;
+
+      for (const script of loafEntry.scripts) {
+        const scriptEndTime = script.startTime + script.duration;
+        if (scriptEndTime < interactionTime) {
+          continue;
+        }
+        const intersectingScriptDuration =
+          scriptEndTime - Math.max(interactionTime, script.startTime);
+        // Since forcedStyleAndLayoutDuration doesn't provide timestamps, we
```
philipwalton, line 296: Hmmm, I hadn't considered this case before. I wonder if this is more or less accurate that just capping the `forcedStyleAndLayoutDuration` at `intersectingScriptDuration`?

src/attribution/onINP.ts
```diff
@@ -260,6 +261,96 @@ export const onINP = (
...
+
+    // Calculate the totalPaintDuration from the last LoAF after
+    // presentationDelay starts (where available)
+    const lastLoAF = attribution.longAnimationFrameEntries.at(-1);
+    const lastLoAFEndTime = lastLoAF
+      ? lastLoAF.startTime + lastLoAF.duration
+      : 0;
+    if (lastLoAFEndTime >= interactionTime + inputDelay + processingDuration) {
+      totalPaintDuration =
+        attribution.interactionTime + value - lastLoAFEndTime;
```
philipwalton, line 334: Should we use `value` here is should we use `nextPaintTime` - `startTime`, which would more accurately match the duration for `presentationDelay`?

src/attribution/onINP.ts
```diff
@@ -260,6 +261,96 @@ export const onINP = (
...
+        entry: longestScriptEntry,
+        subpart: longestScriptSubpart,
+        intersectingDuration: longestScriptDuration,
+      };
+    }
+    attribution.totalScriptDuration = totalScriptDuration;
+    attribution.totalStyleAndLayoutDuration = totalStyleAndLayoutDuration;
+    attribution.totalPaintDuration = totalPaintDuration;
+    attribution.totalUnattributedDuration =
+      value -
```
philipwalton, line 348: Same question here about whether or not this should be based on `nextPaintTime`.

src/attribution/onINP.ts
tunetheweb: But if you have 50ms of script, 100ms of forced style and layout, and 50ms of script
And then you have 80ms of intersection

Are options are:
- Proportion it (40ms script and 40ms forced style and layout) - that's what the PR currently does and for this and is closest to the actual intersecting amount (30ms forced style and layout and 50ms script).
- Capping at script - so 80ms script.
- Capping at force style and layout - so 80ms forced style and layout.

I don't think any of these are "right", but I don't see any reason to favour style and layout over script so went with first one.

Additionally, the aim of these buckets is to show whether you have a "script" or a "style and layout" problem in your app. So I think proportioning it is most in keeping with that?

Oh and FYI @mmocny since you were asking if we needed you to review this or not, this is one piece that maybe you might have thoughts on?

philipwalton: I agree that one isn't clearly "right", and I don't know the answer here (it was a genuine question). I don't have a good sense for whether or not forced style and layout tends to be more-or-less evenly dispersed throughout the script or whether it tends to all happen at the beginning or the end.

tunetheweb: Even it if is predominantly at the beginning or middle or end (and I don't know that either!), to me it's still a bit of a fluke as to how much it overlaps (sometimes user might interaction at the beginning of the previous frame, sometimes near the end). So I think knowing how much of the whole script is actually script versus force s&l is most useful. Hence why I did that, and then basically compressed it down to the overlapping time.

philipwalton: Ok, makes sense. And it's not significantly more code then capping it, so SGTM.

src/attribution/onINP.ts
tunetheweb: Yes. Fixed. Though in this case we want the time not the duration so it's even easier.
Also saves me passing `value` to this function.

src/attribution/onINP.ts
tunetheweb: Done.

README.md
```diff
@@ -881,6 +881,29 @@ interface FCPAttribution {
...
+ * frames.
+ */
+interface LongestScriptSummary {
+  /**
+   * The longest Long Animation Frame script that intersects the INP
+   * interaction.
+   */
+  entry: PerformanceScriptTiming;
+  /**
+   * The INP sub-part where the longest script ran.
```
philipwalton, line 898: Non-blocking: we should change these to subpart throughout now that we've standardized on that in our docs.

philipwalton:
@tunetheweb I was going to make some suggestions, but then I thought it would be faster to just make the changes. Let me know if you have concerns about anything I updated.

README.md

PR #596, opened 2025-03-19
Cleanup onHidden listener

williazz:
Hi, I was reviewing how you record INP and saw that the onHidden listener could be cleaned up. Hopefully this is useful to your team

### DIFF
--- a/src/lib/onHidden.ts
+++ b/src/lib/onHidden.ts
@@ -15,9 +15,11 @@
  */
 
 export const onHidden = (cb: () => void) => {
-  document.addEventListener('visibilitychange', () => {
+  const onHidden = () => {
     if (document.visibilityState === 'hidden') {
       cb();
+      document.removeEventListener('visibilitychange', onHidden);
     }
-  });
+  };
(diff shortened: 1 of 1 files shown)


williazz:
We cannot use the `once` parameter because we cannot teardown the listener until the visibility state is hidden, which may require multiple invocations.

williazz:
Seems that maybe this was intended behavior. 

> INP should be continually monitored for changes throughout the entire lifespan of a page—including if the user returns to the page after it's been hidden/backgrounded. However, since browsers often [will not fire additional callbacks once the user has backgrounded a page](https://developer.chrome.com/blog/page-lifecycle-api/#advice-hidden), `callback` is always called when the page's visibility state changes to hidden. As a result, the `callback` function might be called multiple times during the same page load (see [Reporting only the delta of changes](#report-only-the-delta-of-changes) for how to manage this).

tunetheweb:
Thanks for the PR.

However chunks of this have been rewritten in the v5 branch and I don't think this applies anymore. In particular thanks to #541.

PR #598, opened 2025-03-20
Update README to make it more clear that regardless of reportAllChanges the final reported metric is the same

### LINKED ISSUE #597: [Question] Why is reportAllChanges false by default during onINP()?
Hello all

When `reportAllChanges` is false for `onINP()`, then it seems that only the first interaction is recorded, which does not meet your recommendation that "INP should be continually monitored for changes throughout the entire lifespan of a page". In doing so, this is basically just FID except it reports `PerformanceEventTiming.duration` instead of the processing delay. 

Did we ever consider flipping this to true by default? Most monitoring solutions allow performing calculations on metrics such as maximum or p values, so users should be able to get a decent statistic on all INP candidates. This is even without normalizing the candidates to calculate the maximum INP latency per user visit, ignoring outliers. 

Thanks!

tunetheweb:
Fixes #597


tunetheweb:
Thanks for the PR.

I need to spend some time looking at this and the similar #588 PR. I think there may be an issue to resolve here but not 100% sure as quite a bit has changed in the v5 branch that’s the basis for our next release and not had a chance to fully dig in yet. I have a couple of other PRs open so want to get them sorted first. Then will return to this.

README.md
```diff
@@ -688,7 +688,7 @@ function onCLS(callback: (metric: CLSMetric) => void, opts?: ReportOpts): void;
...
 
 > [!IMPORTANT]
 > CLS should be continually monitored for changes throughout the entire lifespan of a page—including if the user returns to the page after it's been hidden/backgrounded. However, since browsers often [will not fire additional callbacks once the user has backgrounded a page](https://developer.chrome.com/blog/page-lifecycle-api/#advice-hidden), `callback` is always called when the page's visibility state changes to hidden. As a result, the `callback` function might be called multiple times during the same page load (see [Reporting only the delta of changes](#report-only-the-delta-of-changes) for how to manage this).
```
williazz, line 694: Same feedback here

```suggestion
> The `callback` function might be called multiple times during the same page load (see [Reporting only the delta of changes](#report-only-the-delta-of-changes) for how to manage this). CLS should be continually monitored for changes throughout the entire lifespan of a page—including if the user returns to the page after it's been hidden/backgrounded. However, since browsers often [will not fire additional callbacks once the user has backgrounded a page](https://developer.chrome.com/blog/page-lifecycle-api/#advice-hidden), `callback` is always called when the page's visibility state changes to hidden.
```

williazz:
Thank you very much for your collaboration

williazz:
> I need to spend some time looking at this and the similar https://github.com/GoogleChrome/web-vitals/pull/588 PR. I think there may be an issue to resolve here but not 100% sure as quite a bit has changed in the v5 branch that’s the basis for our next release and not had a chance to fully dig in yet. I have a couple of other PRs open so want to get them sorted first. Then will return to this.

Thanks for the update. I also realize now that it may have been intended behavior to call the onHidden callback multiple times without cleaning it up, since that is what your documentation states.

williazz:
https://github.com/GoogleChrome/web-vitals/issues/597#issuecomment-2740226713
> onINP already monitors all interactions across the page lifecycle and but then only reports it at the end of the page (or on tab switching...)

How is INP reported at "the end of the page"? From my limited of your codebase, I see that the maximum INP (ignoring outliers) is reported on visibility state changes to hidden. But I could not see where INP is reported at the end of the page. 

The use case that I am concerned about is:
1. User lands on a website
2. User interacts with the website for 30 minutes without ever triggering a visibility state change to hidden (by hiding the window, switching tabs, etc.)
3. User closes the website.

tunetheweb:
The `visibilitychange` event fires for both tab hides, and also navigations away. Check out the [Page Lifecycle API docs](https://developer.chrome.com/docs/web-platform/page-lifecycle-api/image/page-lifecycle-api-state.svg).

README.md
```diff
@@ -725,7 +725,7 @@ Calculates the [INP](https://web.dev/articles/inp) value for the current page an
...
 
 > [!IMPORTANT]
 > INP should be continually monitored for changes throughout the entire lifespan of a page—including if the user returns to the page after it's been hidden/backgrounded. However, since browsers often [will not fire additional callbacks once the user has backgrounded a page](https://developer.chrome.com/blog/page-lifecycle-api/#advice-hidden), `callback` is always called when the page's visibility state changes to hidden. As a result, the `callback` function might be called multiple times during the same page load (see [Reporting only the delta of changes](#report-only-the-delta-of-changes) for how to manage this).
```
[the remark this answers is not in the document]
tunetheweb: In https://github.com/GoogleChrome/web-vitals/pull/598/commits/daf4a19f9065a23acf01f1988a0fc672c9f937f3I've changes this instead to remove the warning earlier, away from the `reportAllChanges` explanation. Hopefully this makes it clearer.

williazz:
https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState
> The Document.visibilityState read-only property returns the visibility of the document. It can be used to check whether the document is in the background or in a minimized window, or is otherwise not visible to the user.

Thanks I see it here as well now.

philipwalton:
I know the current README says the `reportAllChanges` option is not recommended, but if/when the `fetchLater()` API becomes more widely available, I think there may be cases where it does make more sense to use `reportAllChanges: true`, to get the increased reliability of that API and not depend on the reliability of `visibilitychange` events firing.

I think rather than discouraging the use of `reportAllChanges` to address this confusion, we could just clarify that, regardless of whether `reportAllChanges` is true or false, the final reported value will always be the same.

tunetheweb:
> I know the current README says the `reportAllChanges` option is not recommended, but if/when the `fetchLater()` API becomes more widely available, I think there may be cases where it does make more sense to use `reportAllChanges: true`, to get the increased reliability of that API and not depend on the reliability of `visibilitychange` events firing.

I think there's still a risk of `reportAllChanges` resulting in too much work and causing perf issues depending on what developers do in their `onXXX` callbacks. Of course web-vitals.js does its own processing regardless of `reportAllChanges` but we try to minimise that to the minimum needed to accurately calculate the metrics and take take to yield where necessary. But anything devs do beyond that in the callbacks could be quite intensive so I do think we want to discourage that. This is especially the case for INP which can result in a lot of events (especially when typing for example), or when the default 40ms event timing threshold is reduced—cases I've seen examples of recently.

The `fetchLater()` API use case is interesting. I guess that could be used to queue up the beacon on each change, but only send it once (as in [this example](https://developer.chrome.com/blog/fetch-later-api-origin-trial#example_use), but I do worry that, depending how much processing you need to do to prepare that beacon, you could fall into my concern above. Maybe we need some sort of debounce here to reduce any impact of this.

And I thought `visibilitychange` events were reliable? But it's just there's a short period from that to do any processing and get the beacon out. And so I could see `fetchLater()` being perhaps more reliable for the latter part, but if it's the former that's the issue then again doing that more frequently on every change, may still be a concern.

> I think rather than discouraging the use of `reportAllChanges` to address this confusion, we could just clarify that, regardless of whether `reportAllChanges` is true or false, the final reported value will always be the same.

Happy to do that for now in this PR, and we can discuss all of the above more to decide on that.

philipwalton:
> And I thought `visibilitychange` events were reliable? But it's just there's a short period from that to do any processing and get the beacon out. And so I could see `fetchLater()` being perhaps more reliable for the latter part, but if it's the former that's the issue then again doing that more frequently on every change, may still be a concern.

The `visibilitychange` event is the most reliable lifecycle event we have, but it's still not 100% reliable and I suspect that's one of the reasons developers don't see 100% (or near 100%) beacon success rates.

> Happy to do that for now in this PR, and we can discuss all of the above more to decide on that.

Yeah, that SGTM. Let's use this PR to clarify the confusion around `reportAllChanges` and then discuss whether we should recommend `reportAllChanges` separately.

tunetheweb:
> > Happy to do that for now in this PR, and we can discuss all of the above more to decide on that.
> 
> Yeah, that SGTM. Let's use this PR to clarify the confusion around `reportAllChanges` and then discuss whether we should recommend `reportAllChanges` separately.

Done.

README.md
```diff
@@ -688,11 +688,11 @@ function onCLS(callback: (metric: CLSMetric) => void, opts?: ReportOpts): void;
...
 
+If the `reportAllChanges` [configuration option](#reportopts) is set to `true`, the `callback` function will be called as soon as the value is initially determined as well as any time the value changes throughout the page lifespan (Note [not necessarily for every layout shift](#report-the-value-on-every-change)). Regardless of whether `reportAllChanges` is used, the final reported value will be the same as the `onCLS()` function will monitor all CLS changes throughout the entire lifespan of a page, but without `reportAllChanges` it only invoke the callback when needed, rather than on every larger CLS.
```
philipwalton, line 694:
```suggestion
If the `reportAllChanges` [configuration option](#reportopts) is set to `true`, the `callback` function will be called as soon as the value is initially determined as well as any time the value changes throughout the page lifespan (though [not necessarily for every layout shift](#report-the-value-on-every-change)). Note that regardless of whether `reportAllChanges` is used, the final reported value will be the same.
```

What you have is a bit repetitive. How about simplifying to just this? (Here and in the other places as well.)

philipwalton (approved):

PR #599, opened 2025-03-21
Add retry logic for GitHub Actions

### LINKED ISSUE #601: Flakey tests
These seem to fail most often (the first one the most). Not sure if we can add any more waits or the like to make them less flakey? I've opened a PR #599 to automate the reruns in GitHub Actions.

### LCP 1

[0-3] Error in "onLCP().stops reporting after the document changes to hidden (reportAllChanges === false)"
Error: waitUntil condition timed out after 10000ms
    at async beaconCountIs (file:///Users/runner/work/web-vitals/web-vitals/test/utils/beacons.js:26:3)
    at async Context.<anonymous> (file:///Users/runner/work/web-vitals/web-vitals/test/e2e/onLCP-test.js:283:5)

### INP 1

[0-2] AssertionError in "onINP().attribution.includes LoAF entries if the browser supports it"
AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:
  assert(inp1.attribution.longAnimationFrameEntries.length > 0)
    at Context.<anonymous> (file:///Users/runner/work/web-vitals/web-vitals/test/e2e/onINP-test.js:843:7)

### INP 2

[chrome 134.0.6998.118 mac #0-2] 1) onINP() reports a new interaction after bfcache restore
[chrome 134.0.6998.118 mac #0-2] Expected values to be strictly equal:
+ actual - expected

+ 'good'
- 'needs-improvement'      
      actual expected
      
      'goodneeds-improvement'

tunetheweb:
Fixes #601 

We already test suite twice, but frequently this fails due to flakiness.

This PR adds:
- change to log beacons to /tmp which uses the in-memory tmpfs so should be less subject to delays.
- additional logic to the rerun the whole GitHub Action step, which should relaunch the browser as well. Note there is no native GitHub Action logic for this, so have to use an a third-party action ([such as this one](https://github.com/marketplace/actions/retry-step)), of just add a loop. I went for the latter option here.

It doesn't solve the flakiness, but at least automates the reruns I'm doing manually anyway...

### DIFF
--- a/test/e2e/onLCP-test.js
+++ b/test/e2e/onLCP-test.js
@@ -202,7 +202,7 @@ describe('onLCP()', async function () {
   it('does not report if the document was hidden at page load time', async function () {
     if (!browserSupportsLCP) this.skip();
 
-    await navigateTo('/test/lcp?hidden=1', {readyState: 'interactive'});
+    await navigateTo('/test/lcp?hidden=1', {readyState: 'complete'});
 
     await stubVisibilityChange('visible');
 
@@ -370,7 +370,7 @@ describe('onLCP()', async function () {
   it('reports if the page is restored from bfcache even when the document was hidden at page load time', async function () {
     if (!browserSupportsLCP) this.skip();
 
-    await navigateTo('/test/lcp?hidden=1', {readyState: 'interactive'});
+    await navigateTo('/test/lcp?hidden=1', {readyState: 'complete'});
 
     await stubVisibilityChange('visible');
 
--- a/test/e2e/onLCP-test.js
+++ b/test/e2e/onLCP-test.js
@@ -260,7 +260,7 @@ describe('onLCP()', async function () {
     if (!browserSupportsLCP) this.skip();
 
     await navigateTo('/test/lcp?imgDelay=0&imgHidden=1', {
-      readyState: 'interactive',
+      readyState: 'complete',
     });
 
     // Wait for a frame to be painted.
--- a/test/e2e/onLCP-test.js
+++ b/test/e2e/onLCP-test.js
@@ -266,8 +266,8 @@ describe('onLCP()', async function () {
       readyState: 'complete',
     });
 
-    // Wait for a frame to be painted.
-    await browser.executeAsync((done) => requestAnimationFrame(done));
+    // Wait a bit to ensurethe library it loaded as this test is flakey.
+    await browser.pause(1000);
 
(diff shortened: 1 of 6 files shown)


test/e2e/onLCP-test.js
```diff
@@ -266,8 +266,8 @@ describe('onLCP()', async function () {
       readyState: 'complete',
     });
 
-    // Wait for a frame to be painted.
-    await browser.executeAsync((done) => requestAnimationFrame(done));
+    // Wait a bit to ensurethe library it loaded as this test is flakey.
```
philipwalton, line 269:
```suggestion
    // Wait a bit to ensure the library is loaded, as this test is flakey.
```

philipwalton:
Could you explain why you changed so many of the tests to wait for `complete` instead of `interactive`?

> * change to log beacons to /tmp which uses the in-memory tmpfs so should be less subject to delays.

We could potentially just change the location of that file to be in `/tmp` if that makes things easier. There's no reason it *has* to be under `/test`.

philipwalton (changes requested):

tunetheweb:
> Could you explain why you changed so many of the tests to wait for `complete` instead of `interactive`?

Oh it appears I never hit send on a comment I wrote up for this!!! :-(

So my thinking was `interactive` says "the document has been parsed but sub-resources such as scripts, images, stylesheets and frames are still loading". So there is a risk that the library has not loaded, or that it's not run anyway at least.

Now I'm not 100% sure about how this works for `<script type="module">` (I presume same as `defer`ed scripts so potentially later?) and the imports it makes, and I do see you have logic to add `__testImport` resource loads to `__readyPromises` but that depends on the script having executed (which as I say I'm not convinced is guaranteed to have happened before `interactive`).

Additionally, for most of these tests, basically the only thing waiting to `complete` should be the library load (which we need!) and that should be quick anyway, so didn't see the harm in changing this.

Saying this, I still saw _some_ flakiness with this change, which I can't explain so all of the above could be for naught, but it definitely seemed a bit more stable.

> > * change to log beacons to /tmp which uses the in-memory tmpfs so should be less subject to delays.
> 
> We could potentially just change the location of that file to be in `/tmp` if that makes things easier. There's no reason it _has_ to be under `/test`.

I did think about that, but wondered if there were some Windows people running this, that that would break.

tunetheweb:
Closing this while we see how #603 fares....

PR #600, opened 2025-03-21
Eliminate use of `?module` query string for unpkg since no longer working

westonruter:
I discovered today that our [`benchmark-web-vitals`](https://github.com/GoogleChromeLabs/wpp-research/blob/0184694a93dfa197ac29e4fe428e424dbd7a19b5/cli/commands/benchmark-web-vitals.mjs) command just stopped working. After disabling headless mode in Puppeteer and enabling DevTools, I discovered this in the console:

unnamed

So for some reason, unpkg is no longer redirecting https://unpkg.com/web-vitals@3?module to https://unpkg.com/web-vitals@3.5.2/dist/web-vitals.js but rather to https://unpkg.com/web-vitals@3.5.2/dist/web-vitals.iife.js and this breaks our integration which is using the script module version.

This PR updates the code examples (which we based our integration off of) to remove the use of `?module`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,11 @@ _**Important!** The [unpkg.com](https://unpkg.com), [jsDelivr](https://www.jsdel
 
 ```html
 <script type="module">
-  import {onCLS, onINP, onLCP} from 'https://unpkg.com/web-vitals@4/dist/web-vitals.js';
+  import {
+    onCLS,
+    onINP,
+    onLCP,
+  } from 'https://unpkg.com/web-vitals@4/dist/web-vitals.js';
 
   onCLS(console.log);
   onINP(console.log);
--- a/README.md
+++ b/README.md
@@ -122,9 +122,8 @@ _**Important!** The [unpkg.com](https://unpkg.com), [jsDelivr](https://www.jsdel
 **Load the "standard" build** _(using a module script)_
 
 ```html
-<!-- Append the `?module` param to load the module version of `web-vitals` -->
 <script type="module">
-  import {onCLS, onINP, onLCP} from 'https://unpkg.com/web-vitals@4?module';
+  import {onCLS, onINP, onLCP} from 'https://unpkg.com/web-vitals@4/dist/web-vitals.js';
 
   onCLS(console.log);
   onINP(console.log);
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -154,13 +153,12 @@ _**Important!** The [unpkg.com](https://unpkg.com), [jsDelivr](https://www.jsdel
 **Load the "attribution" build** _(using a module script)_
 
 ```html
-<!-- Append the `?module` param to load the module version of `web-vitals` -->
 <script type="module">
   import {
     onCLS,
     onINP,
     onLCP,
-  } from 'https://unpkg.com/web-vitals@4/dist/web-vitals.attribution.js?module';
```
westonruter, line 163: The use of `?module` here doesn't seem to have been useful anyway since `web-vitals.attribution.js` is already a module.

tunetheweb:
Started following this issue, which I presume is related:
https://github.com/unpkg/unpkg/issues/420

Would like to see an answer from unpkg saying this is intentional and what should be done going forward, before committing any change here.

westonruter:
In any case, I don't see any mention of `?module` on https://unpkg.com/ for how this is supposed to resolve to the module version.

tunetheweb:
There was 2 mins ago!

This is what it used to say: https://web.archive.org/web/20250307145326/https://unpkg.com/ (admittedly it does say it's experimental).

westonruter:
Right you are!

> image

tunetheweb:
Got confirmation this was an unintended change and will be fixed: https://github.com/unpkg/unpkg/issues/420#issuecomment-2745246513

Will close this PR once I hear it's fixed as shouldn't need this.

tunetheweb:
And it's fixed!

westonruter:
Still, why use `?module` if it is very experimental?

README.md
westonruter: And the use here or `?module` doesn't do anything, right?

tunetheweb:
@philipwalton set this up so maybe he can answer?

philipwalton:
I'd be open to changing it to point directly to the module version of the file. The reason I originally went with `?module` is because it was shorter and easier to remember. It also gives us future flexibility to change the build output directory without potentially breaking things.

That said, once we added the explicit major version number to the README it's also less simple, so maybe there isn't a strong reason to keep using `?module`, especially since there are no sub-imports that need resolving. (But again, if we ever *did* want to add sub-imports, we'd have to switch back.)

PR #602, opened 2025-03-24
Avoid throwing error if interaction does not exist

tunetheweb:
Fixes a bug introduced in #585.

After 10 interactions, it's possible, the next one interaction has an `interactionId` but is not longer than the smallest one.

In this case we should not run `_onAfterProcessingInteraction` (and specifically `saveInteractionTarget` which assumes an interaction will be passed).

Also includes a rename of `_onAfterProcessingInteraction` to the more accurate `_onAfterProcessingINPCandidate`

### DIFF
--- a/src/lib/InteractionManager.ts
+++ b/src/lib/InteractionManager.ts
@@ -138,9 +138,8 @@ export class InteractionManager {
           this._longestInteractionMap.delete(interaction.id);
         }
       }
-    }
 
-    if (interaction) {
+      // Call any post-processing on the interaction
       this._onAfterProcessingINPCandidate?.(interaction!);
     }
   }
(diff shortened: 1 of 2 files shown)


src/lib/InteractionManager.ts
```diff
@@ -138,8 +138,9 @@ export class InteractionManager {
           this._longestInteractionMap.delete(interaction.id);
         }
       }
-    }
 
-    this._onAfterProcessingInteraction?.(interaction!);
+      // Call any post-processing on the interaction
+      this._onAfterProcessingINPCandidate?.(interaction!);
```
philipwalton, line 143:
```suggestion
      this._onAfterProcessingINPCandidate?.(interaction);
```

Turns out I shouldn't have ignored TypeScripts warnings on this!

PR #603, opened 2025-03-25
Fix flakey LCP test

philipwalton:
This PR fixes the first issue in #601 by waiting until the library is loaded and a frame has painted (which means an LCP entry should be dispatched) before changing the visibility state in the test to hidden.


philipwalton:
> LGTM and that's the most common failing case, but we should look where else we need this.

Agreed. Just wanted to get something to help the other PRs pass.

PR #611, opened 2025-04-17
Fixing flakey tests

### LINKED ISSUE #601: Flakey tests

### LINKED ISSUE #608: Firefox INP tests disabled
Reminder to myself that in #592 I disabled Firefox INP tests (INP has started to become available in Nightly):

https://github.com/GoogleChrome/web-vitals/blob/3b9acdb1702bf1dbd76f2091cb585619d9d65481/wdio.conf.cjs#L95-L98

This can be re-enabled if https://bugzilla.mozilla.org/show_bug.cgi?id=1960645 is fixed.

tunetheweb:
Fixing flakey tests (especially now Firefox Nightly supports INP so they are even more flakey!)

Fixes #601 
Fixes #608

### DIFF
--- a/test/e2e/onINP-test.js
+++ b/test/e2e/onINP-test.js
@@ -17,10 +17,12 @@
 import assert from 'assert';
 import {beaconCountIs, clearBeacons, getBeacons} from '../utils/beacons.js';
 import {browserSupportsEntry} from '../utils/browserSupportsEntry.js';
+import {firstContentfulPaint} from '../utils/firstContentfulPaint.js';
 import {navigateTo} from '../utils/navigateTo.js';
 import {nextFrame} from '../utils/nextFrame.js';
 import {stubForwardBack} from '../utils/stubForwardBack.js';
 import {stubVisibilityChange} from '../utils/stubVisibilityChange.js';
+import {webVitalsLoaded} from '../utils/webVitalsLoaded.js';
 
 const ROUNDING_ERROR = 8;
 
@@ -93,8 +95,11 @@ describe('onINP()', async function () {
     // after user input.
     await navigateTo('/test/inp?click=100&loadAfterInput=1');
 
-    // Wait until
-    await nextFrame();
+    // Wait until the library is loaded and the first paint occurs to ensure
+    // that an LCP entry can be dispatched prior to the document changing to
+    // hidden.
+    await webVitalsLoaded();
+    await firstContentfulPaint();
 
     const h1 = await $('h1');
     await simulateUserLikeClick(h1);
--- a/test/e2e/onINP-test.js
+++ b/test/e2e/onINP-test.js
@@ -100,6 +100,9 @@ describe('onINP()', async function () {
     const h1 = await $('h1');
     await simulateUserLikeClick(h1);
 
+    // Wait until the library is loaded
+    await webVitalsLoaded();
+
     await stubVisibilityChange('hidden');
 
     await beaconCountIs(1);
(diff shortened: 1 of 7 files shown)


canova:
Hmm it looks like these lines are failing now on Firefox:
https://github.com/GoogleChrome/web-vitals/blob/00f78c5c785aaed744784b58e97d2b8b5f5776e1/test/e2e/onINP-test.js#L370-L373

Did you manage to figure out why? It looks weird, considering that we have a 300ms `pointerdown` blocking time above it, here:
https://github.com/GoogleChrome/web-vitals/blob/00f78c5c785aaed744784b58e97d2b8b5f5776e1/test/e2e/onINP-test.js#L353

tunetheweb:
Woo hoo - got them all to pass now!!

Basically it was a combo of either not waiting for the library to load (and reduce the default threshold to 40ms), or having an INP beneath that 104ms.

Solution is it basically wait for the library to load to begin the test.
Or
Do an interaction that takes longer than 104ms so it's in the bugger.

This has been an occasional problem for Chrome, but it seemed to be more consistently wrong for Firefox. Either way making tests less flakey will benefit testing both!

test/e2e/onINP-test.js
```diff
@@ -696,17 +697,18 @@ describe('onINP()', async function () {
...
+      // Wait until the library is loaded and the first paint occurs to ensure
+      // The 40ms event duration is set
+      await webVitalsLoaded();
+      await firstContentfulPaint();
+
       const h1 = await $('h1');
       await simulateUserLikeClick(h1);
 
-      await stubVisibilityChange('hidden');
-
```
tunetheweb, line 709: @philipwalton this was weird and this change shouldn't be necessary.

Firefox would often fail to report only 1 (and sometimes none!) with this method. But changing it to `&reportAllChanges=1&reportAllChanges2=1` fixed it.

Since the point of this test wasn't to test the visibilitychange thing I was OK changing it, but it might warrant further investigation to see if there's some other flaw in multiple `onINP` calls in Firefox not playing nice with each other (e.g. the reporting of the first `onINP` prevents the second from getting the callback).

philipwalton: SGTM

test/views/inp.njk
```diff
@@ -118,7 +118,7 @@
       inp.instance = 1;
 
       // Log for easier manual testing.
-      console.log(inp);
+      console.log('INP1', inp);
```
philipwalton, line 121: Did you want to do this everywhere, or just for INP?

philipwalton (approved):

test/views/inp.njk
tunetheweb: Good point. Fixed em all.

PR #612, opened 2025-04-22
Use `visibility-state` performance entries

tunetheweb:
Certain metrics (LCP, FCP) should not be emitted if the page was initially loaded in a hidden state since the performance metrics are based on navigation start and paints won't happen until the page is shown.

In the past there was no API to check for the initial hidden value, so the best we could do was observe `visibilitychange` changes. However, this did not handle cases when a page was hidden initially, before the library was loaded and started monitoring for those events.

From Chrome 115 the [`VisibilityStateEntry`](https://chromestatus.com/feature/5683502144028672) was added to the Performance Timeline so we should also check that when supported to better handle this case.

Other browsers can still only depend on `visibilitychange` events and so may still report larger FCP and LCP times when the page is initially opened in a hidden state.

### DIFF
--- a/src/lib/getVisibilityWatcher.ts
+++ b/src/lib/getVisibilityWatcher.ts
@@ -77,6 +77,9 @@ export const getVisibilityWatcher = () => {
     // a perfect heuristic, but it's the best we can do until the
     // `visibility-state` performance entry becomes available in all browsers.
     firstHiddenTime = firstVisibilityStateHiddenTime ?? initHiddenTime();
+    // We're still going to listen to for changes so we can handle things like
+    // bfcache restores and/or prerender without having to examine individual
+    // timestamps in detail.
     addChangeListeners();
 
     // Reset the time on bfcache restores.
--- a/src/lib/getVisibilityWatcher.ts
+++ b/src/lib/getVisibilityWatcher.ts
@@ -64,11 +64,12 @@ const removeChangeListeners = () => {
 export const getVisibilityWatcher = () => {
   if (firstHiddenTime < 0) {
     // Check if we have a previous hidden `visibility-state` performance entry.
-    const firstVisibilityStateHiddenTime =
-      !document.prerendering &&
-      performance
+    // prettier-ignore - Prettier and eslint disagree on this line!
+    const firstVisibilityStateHiddenTime = !document.prerendering
+      ? globalThis.performance
         .getEntriesByType('visibility-state')
-        .filter((e) => e.name === 'hidden')[0]?.startTime;
+        .filter((e) => e.name === 'hidden')[0]?.startTime
+      : undefined;
     // Prefer that, but if it's not available and the document is hidden when
     // this code runs, assume it was hidden since navigation start. This isn't
     // a perfect heuristic, but it's the best we can do until the
--- a/src/lib/getVisibilityWatcher.ts
+++ b/src/lib/getVisibilityWatcher.ts
@@ -65,6 +65,7 @@ export const getVisibilityWatcher = () => {
   if (firstHiddenTime < 0) {
     // Check if we have a previous hidden `visibility-state` performance entry.
     const firstVisibilityStateHiddenTime =
+      !document.prerendering &&
       PerformanceObserver.supportedEntryTypes.includes('visibility-state')
         ? performance
           .getEntriesByType('visibility-state')
--- a/src/lib/getVisibilityWatcher.ts
+++ b/src/lib/getVisibilityWatcher.ts
@@ -64,12 +64,14 @@ const removeChangeListeners = () => {
 export const getVisibilityWatcher = () => {
   if (firstHiddenTime < 0) {
     // Check if we have a previous hidden `visibility-state` performance entry.
-    // prettier-ignore - Prettier and eslint disagree on this line!
+    /* eslint-disable indent */
     const firstVisibilityStateHiddenTime = !document.prerendering
       ? globalThis.performance
-        .getEntriesByType('visibility-state')
-        .filter((e) => e.name === 'hidden')[0]?.startTime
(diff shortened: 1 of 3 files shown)


test/e2e/onLCP-test.js
```diff
@@ -219,6 +225,38 @@ describe('onLCP()', async function () {
     assert.strictEqual(beacons.length, 0);
   });
 
+  it('does not report if the document was hidden before library loaded', async function () {
```
philipwalton, line 228: Good test case!

src/lib/getVisibilityWatcher.ts
```diff
@@ -63,11 +63,20 @@ const removeChangeListeners = () => {
...
-    // since navigation start. This isn't a perfect heuristic, but it's the
-    // best we can do until an API is available to support querying past
-    // visibilityState.
-    firstHiddenTime = initHiddenTime();
+    // Check if we have a previous hidden `visibility-state` performance entry.
+    /* eslint-disable indent */
+    const firstVisibilityStateHiddenTime = !document.prerendering
+      ? globalThis.performance
+          .getEntriesByType('visibility-state')
+          .filter((e) => e.name === 'hidden')[0]?.startTime
```
philipwalton, line 71: This works.

Arguably, though, if the browser supports `visibility-state` entries then you don't need the `visibilitychange` listeners, and you can just always get the first hidden time from the first entry in the `firstHiddenTime()` getter, rather than there.

philipwalton (approved):

src/lib/getVisibilityWatcher.ts
tunetheweb: That's not quite true in the case of bfcache where you wanna ignore times prior to the restore.

It also allows you support partial prerenders that are activated and moved away from before FCP/LCP.

So good to leave it as is? I could also added a comment if you wanted?

sgomes:
Hey @tunetheweb! 👋

I'm reworking our RUM library based on v5.0, and came across the code in this change. As far as I can tell, there has been a behaviour change for prerendered pages here, and I wanted to confirm it is intended.

As far as I can tell, following this change metrics that depend on `visibilityWatcher.firstHiddenTime` (e.g. FCP or LCP) will not report on prerendered pages in Chrome anymore. This is because in `lib/getVisibilityWatcher.ts`, `firstVisibilityStateHiddenTime` will evaluate to `0`, which will initialise the `firstHiddenTime` to `0`. Ordinarily, there would be a `!document.prerendering` check in `initHiddenTime()`, but that call gets skipped altogether.

Running the e2e tests locally (with an updated Chrome) results in several seemingly related failures, which leads me to think this may be a bug, but I wanted to double-check before filing anything.

Thanks!

sgomes:
Perhaps the intention was something along the lines of:

```js
if (firstHiddenTime !== undefined && firstVisibilityStateHiddenTime >= getActivationStart()) {
  firstHiddenTime = firstVisibilityStateHiddenTime;
} else {
  firstHiddenTime = initHiddenTime();
}
```
?

sgomes:
Looking further at this, I think the issue is that the testing framework you have in place doesn't test this correctly, since there's no actual prerendering taking place; the prerendering info seems to be getting stubbed instead.

As far as I can tell, in this scenario, `firstVisibilityStateHiddenTime` will not show the same behaviour as in a real prerendered page.

I'll try to come up with a minimum reproducible test case and file a bug. I'll link to it here once I have it.

tunetheweb:
Yup. Just looking at it now. The issue is the FCP and LCP metrics are gated on whenActivated and by that point document.prerendering is already false :-(

sgomes:
> Yup. Just looking at it now. The issue is the FCP and LCP metrics are gated on whenActivated and by that point document.prerendering is already false :-(

Right, what we'd need to check would be the value of `document.prerendering` at the time of `firstVisibilityStateHiddenTime`. Unless there's some fancy property we can look at that I don't know about, I don't think we can do that directly.

As far as I can tell, the best option is to ignore any `firstVisibilityStateHiddenTime` that comes before `activationStart`, as long as `activationStart` exists. A `firstVisibilityStateHiddenTime` before `activationStart` means that the page was hidden during prerender (which we know it will be), so it's not useful information. If `firstVisibilityStateHiddenTime` comes *after* `activationStart`, then it *is* relevant, and it means the page was hidden in a user-meaningful way.

What do you think?

tunetheweb:
Yeah the easiest thing is to add an explicit check to the filter:

```js
const activationStart = getActivationStart();
.filter((e) => e.name === 'hidden' && e.startTime >= activationStart)[0]?.startTime;
```

That works and will make that change. I'm also going to switch the order for readability:

```js
    const activationStart = getActivationStart();
    const firstVisibilityStateHiddenTime = document.prerendering
      ? Infinity
      : globalThis.performance
          .getEntriesByType('visibility-state')
          .filter((e) => e.name === 'hidden' && e.startTime >= activationStart)[0]?.startTime;
    ```
    
    Now just trying to figure out how to fix the test stubbing to catch this properly...

sgomes:
> Now just trying to figure out how to fix the test stubbing to catch this properly...

An alternative approach is to do real prerendering instead, if at all possible. I haven't used Webdriver.IO before, but it seems like it should be possible to create a setup page with speculation rules and a link to the actual test page, and then programatically click that link to navigate to the prerendered page.

That's the approach we take in our testing, and has been working well for us. We had to build a specific Puppeteer-based test for this, though, as I simply couldn't get prerendering to work with Playwright, for whatever reason.

tunetheweb:
Ah the problem was that `visibility-state` entry on the performance timeline is set to 0 for prerendered pages. But was not set in our stubbing function.

Fixed in https://github.com/GoogleChrome/web-vitals/pull/621

PR #613, opened 2025-04-22
5.0.0-rc.1 Changelog

tunetheweb:
Wait for #612 to be merged first as included that in the CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-### v5.0.0-rc.2 (2024-04-22)
+### v5.0.0-rc.2 (2024-04-23)
 
 - Add upgrading to v5 doc ([#558](https://github.com/GoogleChrome/web-vitals/pull/558))
 - Use visibility-state performance entries ([#612](https://github.com/GoogleChrome/web-vitals/pull/612))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+### v5.0.0-rc.2 (2024-04-22)
+
+- Use visibility-state performance entries ([#612](https://github.com/GoogleChrome/web-vitals/pull/612))
+- Add upgrading to v5 doc ([#558](https://github.com/GoogleChrome/web-vitals/pull/558))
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+### v5.0.0-rc.2 (2024-04-22)
```
tunetheweb, line 3: Update this to release date.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "web-vitals",
-  "version": "5.0.0-rc.0",
+  "version": "5.0.0-rc.1",
```
philipwalton, line 3: This will get automatically updated by the release script.

CHANGELOG.md
philipwalton: I think it's fine to do a PR for this, but what I always do is just add it manually right before doing a release and then push to main 😎 

That way you don't have to guess the date ahead of time.

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+
+- Use visibility-state performance entries ([#612](https://github.com/GoogleChrome/web-vitals/pull/612))
+- Add upgrading to v5 doc ([#558](https://github.com/GoogleChrome/web-vitals/pull/558))
```
philipwalton, line 6: Nit: I think it would be fine to omit this from the CHANGELOG (or not, either way)

package.json
tunetheweb: Oh I knew that worked for major/minor/patch but wasn't sure that also worked for RC patches.

CHANGELOG.md
tunetheweb: Yeah but then you couldn't critique my English :-)

package.json
tunetheweb: Reverted

CHANGELOG.md
tunetheweb: Figured no harm to give this extra visibility. Especially for RC where we won't be announcing it as much.

In fact for that reason, I've moved it to be the first PR noted here in https://github.com/GoogleChrome/web-vitals/pull/613/commits/e6c8b48c99f249ef0b4e741352dfa59ac4c24879

PR #615, opened 2025-04-24
Fix alpha/beta/rc publish tags

tunetheweb:
Just noticed I accidently pushing the RC to the `latest` tag :-(

Corrected now as I presume we want that point to the latest stable release?

Another reason we should have this in the `npm run` scripts (though with the correct parameters!) so don't have to remember this step in future.


philipwalton (approved):
Adding the explicit tag LGTM, but I don't think the build step is needed because calling `npm version` should do that, see: https://github.com/GoogleChrome/web-vitals/blob/b47742de666be1a60faf2311517af4075de7a374/package.json#L98

Issue #502, opened 2024-07-03
Add clarification regarding event listener in event batching section


mikep-dev:
I believe the section regarding event batching requires some clarification.
https://github.com/GoogleChrome/web-vitals?tab=readme-ov-file#batch-multiple-reports-together

The provided code works fine, but I feel it's worth emphasizing the importance of attaching listener for the `visibilitychange` event to the correct target. Check out this demo:
https://stackblitz.com/edit/stackblitz-starters-faxsov?file=index.html

In the demo I've attached event listener to `document` and `window` and configured `onCLS` to add metric to queue and log message to console.

![image](https://github.com/GoogleChrome/web-vitals/assets/15386741/e9c4be5a-3d4d-4fa8-b9e1-a16134166414)

As you can see, if you were to use `document.addEventListener` instead of global `addEventListener` or `window.addEventListener` you would lose CLS data, since CLS metric is reported after that. 

This happens because `onCLS` reports data on `visibilitychange` on `document` but it attaches its listener inside `onFCP` callback. That means our listener that sends analytics data will be attached (and executed) earlier if its attached to the same target. The solution is simple: take advantage of event bubbling and attach the listener to `window` so that it executes after `onCLS` internal listener.

I've made a mistake and targeted `document` and wasted time scratching my head wondering why all other metrics are collected on time, but CLS was always late. I believe adding a proper note to `README` might save some developers time in the future, for example:

> ***Note**: to make sure your event listener is executed after all metrics have been reported, attach it to `window` object with `addEventListener` or `window.addEventListener` instead of `document.addEventListener`, which is internally used by some reporters.*

philipwalton:
Yes, we should update this library to ensure that if someone calls any of the various `onXXX()` functions, the moment that happens all necessary `visibilitychange` listeners get added. That way as long as folks call the `onXXX()` functions before registering their own listeners, our listeners will run first.

We already have the `getVisibilityWatcher.js` module that is initialized early by most of these metric functions. Why don't we add an `onHidden()` method to the [returned watcher object](https://github.com/GoogleChrome/web-vitals/blob/15a377360f3ee7303ba68edc579b1c725111bd98/src/lib/getVisibilityWatcher.ts#L96-L100) that accepts a callback, and then all metric functions can plug into that rather than adding their own `visibilitychange` listeners later in the lifecycle. WDYT?

tunetheweb:
[...]

PR #616, opened 2025-05-01
Misc README updates

tunetheweb:
- Removes note about `getXXX()`
- Fixes bad header markup.
- Simplifies reporting - fixes #81
- Also expands v5 upgrading doc to add references to code splitting docs for popular bundlers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -403,9 +403,6 @@ onINP(addToQueue);
 onLCP(addToQueue);
 
 // Report all available metrics whenever the page is backgrounded or unloaded.
-// NOTE: do not use `document.addEventListener('visibilitychange'..)` since
-// web-vitals itself will report on that event, and you want to include those
-// entires so wait until this bubbles up to window.
 addEventListener('visibilitychange', () => {
   if (document.visibilityState === 'hidden') {
     flushQueue();
--- a/README.md
+++ b/README.md
@@ -274,9 +274,9 @@ The `sendToAnalytics()` function uses the [`navigator.sendBeacon()`](https://dev
 import {onCLS, onINP, onLCP} from 'web-vitals';
 
 function sendToAnalytics(metric) {
-  // Replace with whatever serialization method you prefer.
-  // Note: JSON.stringify will likely include more data than you need.
-  const body = JSON.stringify(metric);
+  // Replace with whatever serialization method you prefer including
+  // whatever attributes you want to include.
+  const body = JSON.stringify({[metric.name]: metric.value});
 
   // Use `navigator.sendBeacon()` to send the data, which supports
   // sending while the page is unloading.
--- a/README.md
+++ b/README.md
@@ -74,20 +74,8 @@ For details on the difference between the builds, see <a href="#which-build-is-r
 To load the "standard" build, import modules from the `web-vitals` package in your application code (as you would with any npm package and node-based build tool):
 
 ```js
-import('web-vitals')
-  .then(({onCLS, onINP, onLCP}) => {
-    onCLS(console.log);
-    onINP(console.log);
-    onLCP(console.log);
-  })
-  .catch((error) => {
-    console.error('Failed to load web-vitals library:', error);
-  });
+import {onLCP, onINP, onCLS} from 'web-vitals';
 ```
-
-<a name="dynamic-imports"></a>
-
-_Note that this code example uses a [dynamic import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import). This is recommended for importing third-party libraries which may have different browser support requirements that your application so that it can fail gracefully without affecting the rest of your code. The alternative is to transpile the code to your required browser support level, but be aware that [many bundlers exclude the `node_modules` for transpiling by default](https://philipwalton.com/articles/the-state-of-es5-on-the-web/). For brevity, the remainder of this README shows static imports._
 
 <a name="attribution-build"></a>
 
@@ -102,8 +90,8 @@ The "attribution" build is slightly larger than the "standard" build (by about 1
 To load the "attribution" build, change any `import` statements that reference `web-vitals` to `web-vitals/attribution`:
 
 ```diff
-- import('web-vitals').then({onCLS, onINP, onLCP}) => {
-+ import('web-vitals/attribution').then({onCLS, onINP, onLCP}) => {
+import {onLCP, onINP, onCLS} from 'web-vitals';
+import {onLCP, onINP, onCLS} from 'web-vitals/attribution';
 ```
 
 Usage for each of the imported function is identical to the standard build, but when importing from the attribution build, the [metric](#metric) objects will contain an additional [`attribution`](#attribution) property.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -74,17 +74,22 @@ For details on the difference between the builds, see <a href="#which-build-is-r
 To load the "standard" build, import modules from the `web-vitals` package in your application code (as you would with any npm package and node-based build tool):
 
 ```js
-import {onLCP, onINP, onCLS} from 'web-vitals';
-
-onCLS(console.log);
-onINP(console.log);
-onLCP(console.log);
+import('web-vitals')
```
philipwalton, line 77: TBH I prefer the original examples, and I think they are going to be much more familiar and approachable to the average developer.

Also (as we discussed offline), I don't think I'd ever actually recommend writing code like this, which makes me very hesitant to include it as the first example people see for loading the library.

If the concern is to ensure we have guidance on how to load the library in a separate file, then I'd rather add a new section to the usage instructions showing how to do that.

Something like this:

**`./performance-monitoring.js`**:
```js
import {onCLS, onINP, onLCP} from 'web-vitals';

export function initPerformanceMonitoring() {
  onCLS(console.log);
  onINP(console.log);
  onLCP(console.log);
}
```

**`./app.js`**:
```js
async function main() {
  // Use dynamic import so all your performance monitoring code (including
  // the `web-vitals` library) gets bundled in its own file, so it won't impact
  // the rest of you application code.
  const initPerformanceMonitoring = await import('./performance-monitoring.js');

  // Initialize monitoring of all web-vitals metrics.
  initPerformanceMonitoring();
}
```

README.md
```diff
@@ -74,17 +74,22 @@ For details on the difference between the builds, see <a href="#which-build-is-r
...
+    onLCP(console.log);
+  })
+  .catch((error) => {
+    console.error('Failed to load web-vitals library:', error);
+  });
 ```
 
-> [!NOTE]
-> In version 2, these functions were named `getXXX()` rather than `onXXX()`. They've [been renamed](https://github.com/GoogleChrome/web-vitals/pull/222) in version 3 to reduce confusion (see [#217](https://github.com/GoogleChrome/web-vitals/pull/217) for details) and will continue to be available using the `getXXX()` until at least version 4. Users are encouraged to switch to the new names, though, for future compatibility.
+<a name="dynamic-imports"></a>
```
philipwalton, line 88: As mention in my other comment, I don't think we need this note, unless we want to move it to its own dedicated section on code splitting.

README.md
```diff
@@ -268,9 +274,9 @@ The `sendToAnalytics()` function uses the [`navigator.sendBeacon()`](https://dev
 import {onCLS, onINP, onLCP} from 'web-vitals';
 
 function sendToAnalytics(metric) {
-  // Replace with whatever serialization method you prefer.
-  // Note: JSON.stringify will likely include more data than you need.
-  const body = JSON.stringify(metric);
+  // Replace with whatever serialization method you prefer including
```
philipwalton, line 277: I think changing the code to list only the specific data being sent is good. But maybe include some more bits of data in the example so it's more clear how someone else might customize this for their needs?

```js
const body = JSON.stringify({
  name: metric.name,
  value: metric.value,
  id: metric.id,
  
  // Include additional data as needed...
});
```


I'd also rephrase the comment to something like: "Replace with whatever specific data you want to capture."

"Replace with whatever metric data you need

README.md
```diff
@@ -405,6 +411,9 @@ onINP(addToQueue);
 onLCP(addToQueue);
 
 // Report all available metrics whenever the page is backgrounded or unloaded.
+// NOTE: do not use `document.addEventListener('visibilitychange'..)` since
```
philipwalton, line 414: I think we need to investigate this more, and if it is a problem, fix the problem rather than recommending people not add this to `document`.

philipwalton (changes requested):

README.md
tunetheweb: OK let's revert all this. I agree that's certainly not how I would write code either but was basing it on your comment here: https://github.com/GoogleChrome/web-vitals/pull/558#discussion_r2051133767

I'm still a tiny bit uncomfortable we've moved to Baseline Newly available but don't promote best practices of that in the README, but think my suggestion is not how people write code, and your way is too much for the README. So I'll just live with that.

If I do do a blog post for this, then maybe that's the place for it.

README.md
tunetheweb: Ditto.

README.md
tunetheweb: I think we can drop the comment and just stick with the "Include additional data as needed..." comment.

Did that.

README.md
philipwalton: Maybe we add my code example to the upgrading guide? Or in the section where we say "via your bundlers code splitting feature", we can link to references that show how to do code splitting in bundlers? E.g.

- https://rollupjs.org/tutorial/#code-splitting
- https://esbuild.github.io/api/#splitting
- https://webpack.js.org/guides/code-splitting/

README.md
tunetheweb: In think I know what this is. 2 seconds...

tunetheweb: Removed this. Will discuss further in the issue: https://github.com/GoogleChrome/web-vitals/issues/502#issuecomment-2848090203

README.md
tunetheweb: > Or in the section where we say "via your bundlers code splitting feature", we can link to references that show how to do code splitting in bundlers? E.g.
> 
> * https://rollupjs.org/tutorial/#code-splitting
> * https://esbuild.github.io/api/#splitting
> * https://webpack.js.org/guides/code-splitting/

I added that to this PR in https://github.com/GoogleChrome/web-vitals/pull/616/commits/485887739bfa7e01621c647e1e3a07607ba715e4. Rereading this, I think that's enough.

docs/upgrading-to-v5.md
```diff
@@ -17,7 +17,7 @@ All of the [builds](README#build-options) in `web-vitals` v5 use only [Baseline
 
 If your site needs to support legacy browsers, you can still use the `web-vitals` library without causing errors in those browsers by adhering to the following recommendations:
 
-We recommend loading the `web-vitals` library in a separate script file from your site's main application bundle(s), either via `<script type="module">` and import statements or via your bundlers code splitting feature. This ensures that any errors encountered while loading the library do not impact other code on your site.
+We recommend loading the `web-vitals` library in a separate script file from your site's main application bundle(s), either via `<script type="module">` and `import` statements or via your bundler's code splitting feature (for example, [rollup](https://rollupjs.org/tutorial/#code-splitting), [escbuild](https://esbuild.github.io/api/#splitting), and [webpack](https://webpack.js.org/guides/code-splitting/)). This ensures that any errors encountered while loading the library do not impact other code on your site.
```
philipwalton, line 20:
```suggestion
We recommend loading the `web-vitals` library in a separate script file from your site's main application bundle(s), either via `<script type="module">` and `import` statements or via your bundler's code splitting feature (for example, [rollup](https://rollupjs.org/tutorial/#code-splitting), [esbuild](https://esbuild.github.io/api/#splitting), and [webpack](https://webpack.js.org/guides/code-splitting/)). This ensures that any errors encountered while loading the library do not impact other code on your site.
```

philipwalton (approved):

PR #617, opened 2025-05-02
Fix Firefox CI flakiness

tunetheweb:


test/utils/isCloseTo.js
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+import assert from 'assert';
+
+/**
+ * Returns if two numbers are within a maxDelta of each other
+ * @return Bool
+ */
+export function isCloseTo(actual, expected, maxDelta) {
```
philipwalton, line 23: Nit: I would rename this to `assertIsCloseTo()` to match similar things we're already doing in the [LCP tests](https://github.com/GoogleChrome/web-vitals/blob/03388e81a28dafbef462399cdaa2a34a80fb2369/test/e2e/onLCP-test.js#L802).

test/utils/waitUntilIdle.js
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+/**
+ * Returns a promise that resolves once the browser has run a rIC
+ * (or a setTimeout for non-supporting browsers).
+ * @return {Promise<void>}
+ */
+export function waitUntilIdle() {
+  return browser.executeAsync((done) => {
+    if ('requestIdleCallback' in window) {
```
philipwalton, line 24:
```suggestion
    if ('requestIdleCallback' in self) {
```

Nit: other tests use [`self`](https://github.com/GoogleChrome/web-vitals/blob/03388e81a28dafbef462399cdaa2a34a80fb2369/test/utils/browserSupportsEntry.js#L27), so just to be consistent.

philipwalton (changes requested):

test/utils/waitUntilIdle.js
tunetheweb: Done.

philipwalton (approved):

PR #619, opened 2025-05-08
Update README with example for generateTarget

tunetheweb:
- Correct one of the definitions in README.
- Add an example
 - Swap round the two differences that attribution has to 1) put the more important one first, and 2) to the more long-winded one second so it flows.


tunetheweb:
Note the failing Firefox test is due to this: https://bugzilla.mozilla.org/show_bug.cgi?id=1950098#c2

README.md
```diff
@@ -790,15 +790,29 @@ console.log(LCPThresholds); // [ 2500, 4000 ]
...
+1. Their callback is invoked with a `MetricWithAttribution` objects instead of a `Metric` object. Each `MetricWithAttribution` extends the `Metric` object and adds an additional `attribution` object, which contains potentially-helpful debugging information that can be sent along with the metric values for the current page visit in order to help identify issues happening to real-users in the field.
 
-```ts
-interface AttributionReportOpts extends ReportOpts {
-  generateTarget?: (el: Node | null) => string;
-}
-```
+2. They accept an `AttributionReportOpts` objects instead of a `ReportOpts` object. The `AttributionReportOpts` object supports an additional, optional, `generateTarget()` function that lets developers customize how DOM elements are stringified for reporting purposes. When passed, the return value `generateTarget()` function will be used for any "target" properties in the following attribution objects: [`CLSAttribution`](#CLSAttribution), [`INPAttribution`](#INPAttribution), and [`LCPAttribution`](#LCPAttribution).
```
philipwalton, line 795:
```suggestion
2. They accept an `AttributionReportOpts` objects instead of a `ReportOpts` object. The `AttributionReportOpts` object supports an additional, optional, `generateTarget()` function that lets developers customize how DOM elements are stringified for reporting purposes. When passed, the return value of the `generateTarget()` function will be used for any "target" properties in the following attribution objects: [`CLSAttribution`](#CLSAttribution), [`INPAttribution`](#INPAttribution), and [`LCPAttribution`](#LCPAttribution).
```

README.md
```diff
@@ -790,15 +790,29 @@ console.log(LCPThresholds); // [ 2500, 4000 ]
...
+
+   ```ts
+   interface AttributionReportOpts extends ReportOpts {
+     generateTarget?: (el: Node | null) => string;
+   }
+   ```
+
+   For example, if a web page has unique `data-name` attibute on many elements you may prefer to use those over the built-in selector-style strings that are generated by the library by default.
```
philipwalton, line 803:
```suggestion
   For example, if a web page has unique `data-name` attibute on many elements, you may prefer to use those over the built-in selector-style strings that are generated by default.
```

README.md
```diff
@@ -790,15 +790,29 @@ console.log(LCPThresholds); // [ 2500, 4000 ]
...
+
+   ```js
+   function customGenerateTarget(el) {
+     if (el.dataset.name) {
+       return el.dataset.name;
+     }
+
+     // Do some other fallback when data-name doesn't exist (not-shown)
```
philipwalton, line 811:
```suggestion
     // Do some other fallback when data-name doesn't exist (not-shown).
```

README.md
```diff
@@ -790,15 +790,29 @@ console.log(LCPThresholds); // [ 2500, 4000 ]
...
+   }
 
-2. Their callback is invoked with a `MetricWithAttribution` objects instead of a `Metric` object. Each `MetricWithAttribution` extends the `Metric` object and adds an additional `attribution` object, which contains potentially-helpful debugging information that can be sent along with the metric values for the current page visit in order to help identify issues happening to real-users in the field.
+   onLCP(generateTarget: customGenerateTarget)
```
philipwalton, line 814:
```suggestion
   onLCP(sendToAnalytics, {generateTarget: customGenerateTarget});
```

This is missing this first argument.

philipwalton (changes requested):

PR #620, opened 2025-05-09
Fix link markdown in `docs/upgrading-to-v4.md`

remarkablemark:
The link was not rendering correctly in https://github.com/GoogleChrome/web-vitals/blob/cae70b1/docs/upgrading-to-v4.md:

Screenshot 2025-05-08 at 11 32 30 PM


tunetheweb (approved):
Thanks!

PR #623, opened 2025-05-28
Handle CLS with no sources

tunetheweb:
In #585 we changed this code:

```js
    if (largestEntry?.sources?.length) {
      const largestSource = getLargestLayoutShiftSource(largestEntry.sources);
```

to this:

```ts
    if (entry.sources.length) {
      const largestSource = getLargestLayoutShiftSource(entry.sources);
```

Unfortunately that removed a check for entry sources which is causing a bug (b/420907153 is the internal issue id)

I've not been able to reliable reproduce how this happens, so can't add a test for it but guessing it happens when the browser garbage collects old entries between emitting the Performance Observer entries and web-vitals processing them?

Either way this is just restoring a check that was there previously (in two places now since we split the logic).

### DIFF
--- a/src/attribution/onCLS.ts
+++ b/src/attribution/onCLS.ts
@@ -72,7 +72,7 @@ export const onCLS = (
   layoutShiftManager._onAfterProcessingUnexpectedShift = (
     entry: LayoutShift,
   ) => {
-    if (entry.sources.length) {
+    if (entry?.sources?.length) {
       const largestSource = getLargestLayoutShiftSource(entry.sources);
       if (largestSource) {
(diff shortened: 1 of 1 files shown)


tunetheweb:
> The only situation where I'd expect this to happen is if the user is on a really old version of Chrome that existed before the layout shift attribution was added. I can't see from the Third Eye view in the linked bug report what UAs this is impacting. Do you know?

To close this out, it looks like it was likely due to a Chrome 79 used on a Smart TV device. The layout instability API was [released in Chrome 77](https://chromestatus.com/feature/5110682739539968), but sources where [not added until Chrome 84](https://chromestatus.com/feature/5712483207610368) so that would make sense.

PR #627, opened 2025-05-31
Remove visibiliytchange event listeners when no longer required

### LINKED ISSUE #622: Large number of visibilitychange handlers
Currently the library adds a large number of visibilitychange handlers and doesn't seem to clear them. https://github.com/GoogleChrome/web-vitals/blob/0263378477936a576d950f400d8616cd20a0b2e6/src/lib/whenIdleOrHidden.ts#L33

Image

## How to reproduce

1. Go to web.dev
2. Paste the classic script from here https://github.com/GoogleChrome/web-vitals?tab=readme-ov-file#from-a-cdn
3. Interact with the page, scroll around, click stuff
4. Run `getEventListeners(document).visibilitychange` in the console

## Thoughts

Seems a bit icky / memory-leaky? Maybe worth removing the handler when the onIdle code runs?

Also, I'm surprised the library runs code on mouseout/pointerleave etc at all, which it does because they are `event` entries from the performance observer.

Image

tunetheweb:
Fixes #622

### DIFF
--- a/src/lib/whenIdleOrHidden.ts
+++ b/src/lib/whenIdleOrHidden.ts
@@ -33,8 +33,7 @@ export const whenIdleOrHidden = (cb: () => void) => {
     rIC(() => {
       cb();
       // Remove the above EventListener since no longer required
-      // to prevent multiple ones building up wasting needless
-      // memory and processing time
+      // See: https://github.com/GoogleChrome/web-vitals/issues/622
       document.removeEventListener('visibilitychange', cb);
     });
   }
--- a/src/lib/whenIdleOrHidden.ts
+++ b/src/lib/whenIdleOrHidden.ts
@@ -29,7 +29,13 @@ export const whenIdleOrHidden = (cb: () => void) => {
     cb();
   } else {
     cb = runOnce(cb);
-    rIC(cb);
     document.addEventListener('visibilitychange', cb, {once: true});
+    rIC(() => {
+      cb();
+      // Remove the above EventListener since no longer required
+      // to prevent multiple ones building up wasting needless
+      // memory and processing time
+      document.removeEventListener('visibilitychange', cb);
+    });
   }
 };


leoswing:
Could anyone help to approved this PR and publish a new version? Same issue occured and needs to fix.

tunetheweb:
@leoswing can you explain the urgency? Our understand and external analysis in https://github.com/GoogleChrome/web-vitals/issues/622#issuecomment-2922994092 suggests the impact is minimal. But yes I'd still like to fix this and release a new version soon.

leoswing:
> @leoswing can you explain the urgency? Our understand and external analysis in [#622 (comment)](https://github.com/GoogleChrome/web-vitals/issues/622#issuecomment-2922994092) suggests the impact is minimal. But yes I'd still like to fix this and release a new version soon.

@tunetheweb hi, in our scenario，the application of micro-frontend seems encountered the serious memory leak problems. So much appreciated if a fix version could be released soon in these days. Thanks a lot ~

M-ZubairAhmed:
@tunetheweb happens in our Mattermost app as well

tunetheweb:
I agree it's happening. What is not so clear is if this is causing problems as per previous comments.

So you're saying you upgraded to v5 and started experiencing serious memory leak problems you did not see in v4.2.4.?

src/lib/whenIdleOrHidden.ts
```diff
@@ -29,7 +29,12 @@ export const whenIdleOrHidden = (cb: () => void) => {
     cb();
   } else {
     cb = runOnce(cb);
-    rIC(cb);
     document.addEventListener('visibilitychange', cb, {once: true});
+    rIC(() => {
+      cb();
+      // Remove the above event listener since no longer required.
+      // See: https://github.com/GoogleChrome/web-vitals/issues/622
+      document.removeEventListener('visibilitychange', cb);
```
philipwalton, line 37:
```suggestion
      document.removeEventListener('visibilitychange', cb, {once: true});
```
I think you might need to make the signature exactly the same in order for the listener ti get removed. Can you confirm?

philipwalton (approved):

src/lib/whenIdleOrHidden.ts
philipwalton: I just tried and it seems to work both ways, so LGTM.

PR #635, opened 2025-07-14
Only finalize LCP on user events (`isTrusted=true`)

### LINKED ISSUE #632: onLCP() stops after synthetic click(), while PerformanceObservers don't
As per [this Chromium Issue](https://g-issues.chromium.org/issues/424077994#comment9), the [current onLCP() implementation](https://github.com/GoogleChrome/web-vitals/blob/main/src/onLCP.ts) will try to [listen to events](https://github.com/GoogleChrome/web-vitals/blob/main/src/onLCP.ts#L103) in order to [stopListening to LCP observer and report final candidate](https://github.com/GoogleChrome/web-vitals/blob/main/src/onLCP.ts#L93-L97).

However, the events are not filtered down to `isTrusted` events, such that artificial `click` events seem to trigger the event and will stop the algorithm.

---

It's not clear to me how to prioritize this-- are there any legitimate patterns where synchronous events might be triggered during early load as part of the default loading experience?

The use case of the issue linked might have been lab automation, where input maybe should be stopping LCP?

tunetheweb:
Fixes #632 

Note this meant the stubVisibilityChange no longer finalized LCP in the test suite so I switched most of those to a new `switchToNewTabAndBack` function.

### DIFF
--- a/test/e2e/onLCP-test.js
+++ b/test/e2e/onLCP-test.js
@@ -83,7 +83,7 @@ describe('onLCP()', async function () {
     // Wait until all images are loaded and fully rendered.
     await imagesPainted();
 
-    // Click on the h1.
+    // Click on the h1 to finalize LCP.
     const h1 = await $('h1');
     await h1.click();
 
@@ -99,10 +99,6 @@ describe('onLCP()', async function () {
     // Wait until all images are loaded and fully rendered.
     await imagesPainted();
 
-    // Click on the h1.
-    const h1 = await $('h1');
-    await h1.click();
-
     await beaconCountIs(2);
     assertFullReportsAreCorrect(await getBeacons());
   });
@@ -115,7 +111,7 @@ describe('onLCP()', async function () {
     // Wait until all images are loaded and fully rendered.
     await imagesPainted();
 
-    // Click on the h1.
+    // Click on the h1 to finalize LCP.
     const h1 = await $('h1');
     await h1.click();
 
@@ -169,7 +165,7 @@ describe('onLCP()', async function () {
     // Wait until web-vitals is loaded
     await webVitalsLoaded();
 
-    // Click on the h1.
+    // Click on the h1 to finalize LCP.
     const h1 = await $('h1');
     await h1.click();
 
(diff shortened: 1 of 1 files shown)


test/e2e/onLCP-test.js
```diff
@@ -115,6 +115,9 @@ describe('onLCP()', async function () {
     // Wait until all images are loaded and fully rendered.
     await imagesPainted();
 
+    // Pause to ensure web-vitals has a chance to collect performance metrics
```
philipwalton, line 118: Line 124 already has built-in waiting for collection, so that makes me wonder if this is needed for some other reason? Is there another signal we could wait for that brings the same reliability without the arbitrariness of 100ms?

test/e2e/onLCP-test.js
```diff
@@ -921,3 +904,12 @@ const assertFullReportsAreCorrect = (beacons) => {
   assert(lcp2.entries[0].startTime > lcp1.entries[0].startTime);
   assert.match(lcp2.navigationType, /navigate|reload/);
 };
+
+const switchToNewTabAndBack = async () => {
```
philipwalton, line 908: 👍 we probably should have been doing something like this the whole time.

src/onLCP.ts
```diff
@@ -96,17 +96,25 @@ export const onLCP = (
         report(true);
       });
 
+      // Need a separate wrapper to ensure the `runOnce` function above is
+      // common for all three functions
+      const stopListeningWrapper = (e: Event) => {
```
philipwalton, line 101: Nit: we've used `event` (rather than `e`) in other places in this library so I'd rename to be consistent.

philipwalton (approved):
Overall, LGTM. I just had a few minor comments.

test/e2e/onLCP-test.js
tunetheweb: I _think_ the racy part is DCL waits for the library to be loaded, but not for the PO to have fired before we navigate away (and then also the callbacks for run once and idle until urgent).

it seems to mostly work on Chrome but is flakey on Firefox.

And we can’t await the beacon before the navigation since it’s the navigation that triggers the beacon. I can change this to a H1 click instead of a nav to trigger the beacon and that does work. I was in two minds whether to do that anyway, but wanted to confirm the reason it was racy (and why mostly in this test).

test/e2e/onLCP-test.js
tunetheweb: Yeah I only recently added a test to do an actual tab change in #612 (and you also liked it then!) so before that we didn’t have anything like this.

PR #636, opened 2025-07-15
Guard INP attribution no entries

### LINKED ISSUE #639: Cannot read properties of undefined (reading 'processingStart')
Hi,

Trying out the new soft-navs branch, have installed and enabled ` chrome://flags/#soft-navigation-heuristics` in canary chrome (ver. 140)

Seems to work pretty well, but after clicking around some soft navigations, then minimising the browser, I'm getting this in the console:

```
web-vitals.attribution.js?module:1 Uncaught TypeError: Cannot read properties of undefined (reading 'processingStart')
    at v (web-vitals.attribution.js?module:1:9919)
    at web-vitals.attribution.js?module:1:11472
    at web-vitals.attribution.js?module:1:1330
    at HTMLDocument.<anonymous> (web-vitals.attribution.js?module:1:8335)
```

Not sure if that's a bug, or an implementation issue, do we need to add something I've blanked out on the docs?

tunetheweb:
Fixes #639 

INP attribution doesn't currently check if there are any `event` entries, whereas all the other metrics do.

This can result in an error when attempting to access the first entry.

This does require making each of the `attribution` fields optional (similar to CLS), but given this can only happen when it would error anyway I think that's OK without a major change.

### DIFF
--- a/src/attribution/onINP.ts
+++ b/src/attribution/onINP.ts
@@ -269,9 +269,12 @@ export const onINP = (
       return;
     }
 
-    const interactionTime = attribution.interactionTime;
-    const inputDelay = attribution.inputDelay;
-    const processingDuration = attribution.processingDuration;
+    // We can assume these will be set as we only call attributeLoAFDetails
+    // when we have an entry, and so these details.
+    const interactionTime = attribution.interactionTime!;
+    const inputDelay = attribution.inputDelay!;
+    const processingDuration = attribution.processingDuration!;
+    const nextPaintTime = attribution.nextPaintTime!;
 
     // Stats across all LoAF entries and scripts.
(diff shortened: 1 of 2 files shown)


philipwalton:
> This can result in an error when attempting to access the first entry.

Can you clarify how it's possible for there to be an INP with no entries?

> INP attribution doesn't currently check if there are any `event` entries, whereas all the other metrics do.

With all of the other metrics, it's possible for there to be an metric event with no entries:

- CLS: when there are no layout shifts, so CLS is 0
- LCP: after a bfcache restore, since there aren't LCP entries for that (yet)
- FCP: after a bfcache restore
- TTFB: after a bfcache restore

But these cases do not apply to INP. Is there another case I'm not aware of?

PR #637, opened 2025-07-15
Register visibility-change early

### LINKED ISSUE #502: Add clarification regarding event listener in event batching section

tunetheweb:
Fixes #502

@philipwalton I did what you suggested in https://github.com/GoogleChrome/web-vitals/issues/502#issuecomment-2848132053 .

### DIFF
--- a/src/lib/getVisibilityWatcher.ts
+++ b/src/lib/getVisibilityWatcher.ts
@@ -18,7 +18,7 @@ import {onBFCacheRestore} from './bfcache.js';
 import {getActivationStart} from './getActivationStart.js';
 
 let firstHiddenTime = -1;
-const onHiddenFunctions: (() => void)[] = [];
+const onHiddenFunctions: Set<() => void> = new Set();
 
 const initHiddenTime = () => {
   // If the document is hidden when this code runs, assume it was always
@@ -34,8 +34,10 @@ const initHiddenTime = () => {
 const onVisibilityUpdate = (event: Event) => {
   // Handle changes to hidden state
   if (document.visibilityState === 'hidden') {
-    for (const onHiddenFunction of onHiddenFunctions) {
-      onHiddenFunction();
+    if (event.type === 'visibilitychange') {
+      for (const onHiddenFunction of onHiddenFunctions) {
+        onHiddenFunction();
+      }
     }
 
     // If the document is 'hidden' and no previous hidden timestamp has been
@@ -53,7 +55,7 @@ const onVisibilityUpdate = (event: Event) => {
       // We no longer need the `prerenderingchange` event listener now we've
       // set an initial init time so remove that
       // (we'll keep the visibilitychange one for onHiddenFunction above)
-      addEventListener('prerenderingchange', onVisibilityUpdate, true);
+      removeEventListener('prerenderingchange', onVisibilityUpdate, true);
     }
   }
 };
@@ -80,7 +82,7 @@ export const getVisibilityWatcher = () => {
 
     // Listen for visibility changes so we can handle things like bfcache
     // restores and/or prerender without having to examine individual
-    // timestamps in detail and also for onHiddenFunction calls.
+    // timestamps in detail and also for onHidden function calls.
     addEventListener('visibilitychange', onVisibilityUpdate, true);
     // IMPORTANT: when a page is prerendering, its `visibilityState` is
     // 'hidden', so in order to account for cases where this module checks for
@@ -102,8 +104,8 @@ export const getVisibilityWatcher = () => {
     get firstHiddenTime() {
       return firstHiddenTime;
     },
-    set onHidden(cb: () => void) {
-      onHiddenFunctions.push(cb);
+    onHidden(cb: () => void) {
+      onHiddenFunctions.add(cb);
     },
   };
 };
--- a/src/lib/getVisibilityWatcher.ts
+++ b/src/lib/getVisibilityWatcher.ts
@@ -66,7 +66,7 @@ const removeChangeListeners = () => {
 // loads (e.g. for batch processing) run early.
 // See https://github.com/GoogleChrome/web-vitals/issues/502
 const onHiddenFunctions: (() => void)[] = [];
-document.addEventListener('visibilitychange', () => {
+addEventListener('visibilitychange', () => {
   if (document.visibilityState === 'hidden') {
     for (const onHiddenFunction of onHiddenFunctions) {
       onHiddenFunction();
--- a/src/lib/getVisibilityWatcher.ts
+++ b/src/lib/getVisibilityWatcher.ts
@@ -62,6 +62,18 @@ const removeChangeListeners = () => {
   removeEventListener('prerenderingchange', onVisibilityUpdate, true);
 };
 
+// Register visibility change early so those registering after library
+// loads (e.g. for batch processing) run early.
+// See https://github.com/GoogleChrome/web-vitals/issues/502
+const onHiddenFunctions: (() => void)[] = [];
+document.addEventListener('visibilitychange', () => {
+  if (document.visibilityState === 'hidden') {
+    for (const onHiddenFunction of onHiddenFunctions) {
+      onHiddenFunction();
+    }
+  }
+});
+
 export const getVisibilityWatcher = () => {
   if (firstHiddenTime < 0) {
     // Check if we have a previous hidden `visibility-state` performance entry.
@@ -100,6 +112,9 @@ export const getVisibilityWatcher = () => {
   return {
     get firstHiddenTime() {
       return firstHiddenTime;
+    },
+    set onHidden(cb: () => void) {
+      onHiddenFunctions.push(cb);
     },
   };
 };
--- a/src/lib/getVisibilityWatcher.ts
+++ b/src/lib/getVisibilityWatcher.ts
@@ -66,7 +66,7 @@ const removeChangeListeners = () => {
 // loads (e.g. for batch processing) run early.
 // See https://github.com/GoogleChrome/web-vitals/issues/502
 const onHiddenFunctions: (() => void)[] = [];
-addEventListener('visibilitychange', () => {
+globalThis.addEventListener?.('visibilitychange', () => {
   if (document.visibilityState === 'hidden') {
     for (const onHiddenFunction of onHiddenFunctions) {
       onHiddenFunction();
--- a/src/lib/getVisibilityWatcher.ts
+++ b/src/lib/getVisibilityWatcher.ts
@@ -18,6 +18,7 @@ import {onBFCacheRestore} from './bfcache.js';
 import {getActivationStart} from './getActivationStart.js';
 
 let firstHiddenTime = -1;
+const onHiddenFunctions: (() => void)[] = [];
 
 const initHiddenTime = () => {
   // If the document is hidden when this code runs, assume it was always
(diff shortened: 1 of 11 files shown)


src/lib/whenIdleOrHidden.ts
```diff
@@ -29,12 +29,12 @@ export const whenIdleOrHidden = (cb: () => void) => {
     cb();
   } else {
     cb = runOnce(cb);
-    document.addEventListener('visibilitychange', cb, {once: true});
+    addEventListener('visibilitychange', cb, {once: true});
```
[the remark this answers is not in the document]
tunetheweb: Not needed IIUC, more consistent, and saves 8 bytes. All the rest were removed in this PR anyway.

test/views/layout.njk
```diff
@@ -43,7 +43,7 @@
        */
       self.__stubVisibilityChange = (visibilityState) => {
         self.__stubVisibilityState(visibilityState);
-        document.dispatchEvent(new Event('visibilitychange'));
+        dispatchEvent(new Event('visibilitychange'));
```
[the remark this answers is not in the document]
tunetheweb: Yes as I haven't changed FCP, INP, and CLS yet, since they don't _need_ real tab switches—unlike LCP which does now thanks to #635 . Maybe should change most of those for consistency. Can do that in a separate PR if you want.

Even then, two LCP tests still depend on this (we want to start as hidden, and then do stuff), which can't be easily be tested with real tab switches since webdriver.io doesn't have a "open in hidden tab" option last time I checked.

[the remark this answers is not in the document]
philipwalton: Gotcha. In that case keeping this makes sense to me.

src/lib/getVisibilityWatcher.ts
```diff
@@ -62,6 +62,22 @@ const removeChangeListeners = () => {
   removeEventListener('prerenderingchange', onVisibilityUpdate, true);
 };
 
+// Register visibility change early so those registering after library
+// loads (e.g. for batch processing) run early.
+// See https://github.com/GoogleChrome/web-vitals/issues/502
+const onHiddenFunctions: (() => void)[] = [];
+globalThis.addEventListener?.(
```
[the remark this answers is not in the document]
philipwalton: > Are you suggesting no longer calling `removeChangeListeners` then since we want this one for `onHidden` to be to fire on each visibility change rather than just the first one like the current ones do.

Yes, or alternatively you could just remove the `prerenderingchange` one, but since that doesn't fire often it might not be that big of a deal.

> That would be quite a bit of logic change as far as I can tell and not sure if it's worth it.

I don't think it would be that much of a logic change? And if it ends up being complex to have it all in one listener then you could have multiple, but I still think it would be better to add them all at the same time rather than one on import and one on initialize.

src/lib/whenIdleOrHidden.ts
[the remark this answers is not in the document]
tunetheweb: Or we could just set the [`capture` option](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#capture)?

src/lib/getVisibilityWatcher.ts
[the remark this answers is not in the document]
tunetheweb: Updated. Yeah was smaller than I thought.

src/lib/getVisibilityWatcher.ts
```diff
@@ -31,35 +32,30 @@ const initHiddenTime = () => {
...
-    // If the event is a 'visibilitychange' event, it means the page was
-    // visible prior to this change, so the event timestamp is the first
-    // hidden time.
-    // However, if the event is not a 'visibilitychange' event, then it must
-    // be a 'prerenderingchange' event, and the fact that the document is
-    // still 'hidden' from the above check means the tab was activated
-    // in a background state and so has always been hidden.
-    firstHiddenTime = event.type === 'visibilitychange' ? event.timeStamp : 0;
+  // Handle changes to hidden state
+  if (document.visibilityState === 'hidden') {
```
philipwalton, line 36: Should we also check whether the event is a `visibilitychange` event?

I'm not sure if this library will add `onHidden()` functions before `prerenderingchange`, but maybe it's better to be explicit just in case?

src/lib/getVisibilityWatcher.ts
```diff
@@ -18,6 +18,7 @@ import {onBFCacheRestore} from './bfcache.js';
 import {getActivationStart} from './getActivationStart.js';
 
 let firstHiddenTime = -1;
+const onHiddenFunctions: (() => void)[] = [];
```
philipwalton, line 21: Optional: if you made this a `Set` instead of an array, then it would prevent you from adding the same function twice (which matches the native `addEventListener()` behavior). I don't think that will matter in this library, but in general I like to use Sets for these types of things to to keep the behavior the same.

src/lib/getVisibilityWatcher.ts
```diff
@@ -31,35 +32,30 @@ const initHiddenTime = () => {
...
+      // in a background state and so has always been hidden.
+      firstHiddenTime = event.type === 'visibilitychange' ? event.timeStamp : 0;
 
-const removeChangeListeners = () => {
-  removeEventListener('visibilitychange', onVisibilityUpdate, true);
-  removeEventListener('prerenderingchange', onVisibilityUpdate, true);
+      // We no longer need the `prerenderingchange` event listener now we've
+      // set an initial init time so remove that
+      // (we'll keep the visibilitychange one for onHiddenFunction above)
+      addEventListener('prerenderingchange', onVisibilityUpdate, true);
```
philipwalton, line 56:
```suggestion
      removeEventListener('prerenderingchange', onVisibilityUpdate, true);
```

src/lib/getVisibilityWatcher.ts
```diff
@@ -93,13 +95,15 @@ export const getVisibilityWatcher = () => {
...
         firstHiddenTime = initHiddenTime();
-        addChangeListeners();
       });
     });
   }
   return {
     get firstHiddenTime() {
       return firstHiddenTime;
     },
+    set onHidden(cb: () => void) {
```
philipwalton, line 105:
```suggestion
    onHidden(cb: () => void) {
```
I think you want this to be a normal function and not a setter, right?

src/onCLS.ts
```diff
@@ -85,12 +87,10 @@ export const onCLS = (
           opts!.reportAllChanges,
         );
 
-        document.addEventListener('visibilitychange', () => {
-          if (document.visibilityState === 'hidden') {
-            handleEntries(po.takeRecords() as CLSMetric['entries']);
-            report(true);
-          }
-        });
+        visibilityWatcher.onHidden = () => {
```
philipwalton, line 90: Ahh, I see you're using it as a setter here as well.

I'd make this a regular function that takes a callback, otherwise it looks like setting a function overrides the previously set value (similar to the native event handler properties like `.onload = `).

src/onINP.ts
```diff
@@ -129,12 +132,10 @@ export const onINP = (
       // where the first interaction is less than the `durationThreshold`.
       po.observe({type: 'first-input', buffered: true});
 
-      document.addEventListener('visibilitychange', () => {
-        if (document.visibilityState === 'hidden') {
-          handleEntries(po.takeRecords() as INPMetric['entries']);
-          report(true);
-        }
-      });
+      visibilityWatcher.onHidden = () => {
```
philipwalton, line 135: Same here as above.

philipwalton (changes requested):

src/lib/getVisibilityWatcher.ts
tunetheweb: Added.

src/lib/getVisibilityWatcher.ts
tunetheweb: 🤦 Thanks!

tunetheweb:
I got bored of manually testing this each time so added a test case. Despite my worries in the original comment it wasn't too bad to add.

test/views/cls.njk
```diff
@@ -34,19 +34,50 @@
   <script type="module">
     const {onCLS} = await __testImport('{{ modulePath }}');
 
+    const queue = new Set();
```
philipwalton, line 37: I think it's good to add a test for an example we show in the README, but I wonder if it's overkill to test it for every single metric.

What if instead of testing it for every metric we added a single new test page that tests how all of the metrics work together and includes tests for the examples shown in the README. WDYT?

test/views/cls.njk
```diff
@@ -34,19 +34,50 @@
...
+        navigator.sendBeacon(`/collect`, JSON.stringify(__toSafeObject(cls)));
+      }
     }, {
       reportAllChanges: self.__reportAllChanges,
       generateTarget: self.__generateTarget && ((el) => el.dataset.target),
     });
 
+    if (self.__visibilityChangeEventHandlers) {
+      // Equal to `window.addEventListener`
+      addEventListener('visibilitychange', () => {
```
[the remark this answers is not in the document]
tunetheweb: True. Removed from all tests.

test/e2e/onCLS-test.js
```diff
@@ -832,6 +832,28 @@ describe('onCLS()', async function () {
     assert.strictEqual(cls1_1.navigationType, cls2_3.navigationType);
   });
 
+  it('reports on document visibility change handling', async function () {
```
philipwalton, line 835: I think the description in this test is misleading. The library always reported on `visibilitychange`, and there are already numerous tests for that behavior.

What's different here is your testing to ensure that if a developer chooses to implement their own batching (following the example in the README), that all of the relevant metric callbacks have run before their code does. So I'd rename this to clarify that.

Maybe something like: "it reports before developer-added visibilitychange listeners run"

test/views/cls.njk
```diff
@@ -34,19 +34,41 @@
...
+
     onCLS((cls) => {
       cls.instance = 1;
 
       // Log for easier manual testing.
       console.log('CLS:', cls);
 
-      // Test sending the metric to an analytics endpoint.
-      navigator.sendBeacon(`/collect`, JSON.stringify(__toSafeObject(cls)));
+      if (self.__documentVisibilityChangeHandling) {
```
philipwalton, line 48: I would call this something different (perhaps `self.__batchReporting`) because right now it's not very clear what this does.

test/views/cls.njk
philipwalton: It's up to you. I agree they're metric specific, but personally I'd be totally comfortable with a single test that reports all of the metrics, to test them all at the same time.

That type of test would also ensure that calling all of the metric functions together doesn't cause any of them to break (we don't currently have a test for that).

test/views/cls.njk
tunetheweb: Fixed

test/e2e/onCLS-test.js
tunetheweb: I changed it to "reports on batch reporting using document.visibilitychange" cause the key thing was the `document` bit.

test/views/cls.njk
```diff
@@ -34,19 +34,41 @@
...
+      } else {
+        // Test sending the metric to an analytics endpoint.
+        navigator.sendBeacon(`/collect`, JSON.stringify(__toSafeObject(cls)));
+      }
     }, {
       reportAllChanges: self.__reportAllChanges,
       generateTarget: self.__generateTarget && ((el) => el.dataset.target),
     });
 
+    if (self.__batchReporting) {
```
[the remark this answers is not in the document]
tunetheweb: It’s true the Set should never be > 1 with the test cases as they’re written now.

The metrics are re-inited in bfcache but that would necessitate a visibilitychange after which the queue should be cleared (but that’s an assumption/dependency I don’t think we should reply on). And in the soft navs branch they could be re-inited within a page view. But for now, particularly with the test cases as they exist, that shouldn’t happen as you say. I’d still rather have a loop to code this properly rather than making these assumptions.

Are you suggesting removing these test cases and not testing this until we get to that other PR? I think they’re still useful now.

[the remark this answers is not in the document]
philipwalton: Ultimately I don't think it's that big of a deal, and I don't mean to block merging this or releasing the other pending features.

I think my main concerns are that it's still not super clear what's being tested. E.g. batching is really about sending multiple metrics together in the same HTTP request, but in this test the use of batching is actually only testing manual sending in the `visibilitychange` (rather than relying on the library to send things at the right time)—it's not actually testing any sort of batch sending.

So my concern is that if for some reason someone other than you or me had to update these tests in the future, it might not be clear why these tests are needed or what they're actually testing.

PR #660, opened 2025-11-29
Use queueMicrotask() for PerformanceObserver callbacks

Mimori256:
This PR replaces `Promise.resolve().then(){ with `queueMicrotask()` when scheduling the `PerformanceObserver` callback.

This change makes the intention of microtask queuing clear and documents the Safari workaround, and also reduces the overhead of creating unnecessary promises.

`queueMicrotask()` is part of the Baseline widely available set, so this change should not introduce any new browser compatibility issues.


tunetheweb (approved):
Sorry missed this. I think this is OK and better practice, even if functionally equivalent.

@philipwalton any opinions?

PR #662, opened 2025-12-19
Simplify the event and LoAF entry clean up logic

philipwalton:
This PR updates logic and comments in the `queueCleanup()` function for `attribution/onINP.ts` to make it easier to follow what's happening.

### DIFF
--- a/test/e2e/onFCP-test.js
+++ b/test/e2e/onFCP-test.js
@@ -348,11 +348,7 @@ describe('onFCP()', async function () {
         /^(loading|dom-(interactive|content-loaded)|complete)$/,
       );
 
-      // TODO Firefox has a bug causing this to fail. Remove once fixed
-      // https://bugzilla.mozilla.org/show_bug.cgi?id=1965441
-      if (browser.capabilities.browserName !== 'firefox') {
-        assert.deepEqual(fcp.attribution.fcpEntry, fcpEntry);
-      }
+      assert.deepEqual(fcp.attribution.fcpEntry, fcpEntry);
 
       // When FCP is reported, not all values on the NavigationTiming entry
       // are finalized, so just check some keys that should be set before FCP.


src/attribution/onINP.ts
```diff
@@ -202,39 +202,45 @@ export const onINP = (
...
-    // Keep all render times that are part of a pending INP candidate or
-    // that occurred within the 50 most recently-dispatched groups of events.
-    const longestInteractionGroups =
+    // Create a set of entries groups that are part of the longest
+    // interactions (for faster lookup below).
+    const longestInteractionGroups = new Set(
       interactionManager._longestInteractionList.map((i) => {
         return entryToEntriesGroupMap.get(i.entries[0]);
-      });
-    const minIndex = pendingEntriesGroups.length - MAX_PREVIOUS_FRAMES;
```
philipwalton: I removed this variable and just put the logic directly into the filter conditional. To me that makes it easier to understand what's happening, because having a variable with the term "index" in it—that can be negative—is more confusing to me than just looking at the logic directly.

src/attribution/onINP.ts
```diff
@@ -202,39 +202,45 @@ export const onINP = (
...
+      return (
+        longestInteractionGroups.has(group) ||
+        i >= pendingEntriesGroups.length - MAX_PENDING_FRAMES
+      );
     });
 
-    // Keep all pending LoAF entries that either:
-    // 1) intersect with entries in the newly cleaned up `pendingEntriesGroups`
-    // 2) occur after the most recently-processed event entry (for up to MAX_PREVIOUS_FRAMES)
-    const loafsToKeep: Set<PerformanceLongAnimationFrameTiming> = new Set();
```
philipwalton: Renaming `loafsToKeep` to `intersectionLoAFs` because these aren't the only LoAFs that are being kept.

src/attribution/onINP.ts
```diff
@@ -202,39 +202,45 @@ export const onINP = (
...
+    const intersectingLoAFs: Set<PerformanceLongAnimationFrameTiming> =
+      new Set();
     for (const group of pendingEntriesGroups) {
       const loafs = getIntersectingLoAFs(group.startTime, group.processingEnd);
       for (const loaf of loafs) {
-        loafsToKeep.add(loaf);
+        intersectingLoAFs.add(loaf);
       }
     }
-    const prevFrameIndexCutoff = pendingLoAFs.length - 1 - MAX_PREVIOUS_FRAMES;
```
philipwalton: @tunetheweb I removed the use of this `prevFrameIndexCutoff` because I don't think it's needed.

Since, for INP, LoAF entries are only relevant if they intersect with event entries, the check for intersections with the ~50 `pendingEntriesGroups` should be sufficient. Also, since there  can be event timing entries in cases where there are no LoAF entries, keeping ~50 LoAF entries ends up applying to much more than the previous 50 frames, so I believe that's unnecessary.

philipwalton: But please double check this, since AFAICT this is the only change in behavior from the previous code logic.

tunetheweb: I think it's OK logically. But I wonder if the reason you added it in #512 was to allow it to avoid looking up the Set when this matches? Checking if it's in the Set is a constant time thanks to it being a Set, but it's not as cheap as a quick `>=` index check. So maybe keeping this to allow that early exit is still better?

tunetheweb (changes requested):
I'm not 100% convinced this PR is actually that much of an improvement. See comments above, but especially for these next two comments:

src/attribution/onINP.ts
tunetheweb: Actually ignore me. It looks like it would only keep a max of 50 pending LoAFs and discard others. Which might not not be the right 50 to keep. So I agree with removing this.

PR #670, opened 2026-01-07
deps: update browserslist database (caniuse-lite)

Mimori256:
This PR updates the browserslist database (`caniuse-lite`) to the latest version to resolve the "caniuse-lite is outdated" warning during build

The build process was showing the following warning as the current caniuse-lite version is 1.0.30001571, while the latest version is 1.0.30001762 (191 versions behind):
```
Browserslist: caniuse-lite is outdated. Please run:
  npx update-browserslist-db@latest
  Why you should do it regularly: https://github.com/browserslist/update-db#readme
```

This update ensures latest browser support information for Babel transpilation


Mimori256:
Thanks for the review! 

The build warning was bugging me a bit, so I figured it was worth updating. You're right that we'll probably need to do this periodically. FYI there are actions like https://github.com/c2corg/browserslist-update-action that can help automate this, if that's ever useful (Of course, it's also possible to create a custom action if that fits better)

PR #735, opened 2026-04-30
fix: correct INP navigationURL after bfcache restore on soft-nav pages

### LINKED ISSUE #734: [soft-navs]: onINP bfcache restore reports original hard-nav URL instead of soft-nav URL as `navigationURL`
### Description

When an SPA performs a soft navigation and the user hard-navigates away and returns via bfcache restore, INP `navigationURL` reports the original hard navigation URL instead of the URL that was active when the page was put into bfcache. CLS and LCP report the correct soft navigation URL.

For instance:

1. Hard navigate to `http://localhost` (SPA loads)
2. Perform a soft navigation to `/about` (`metric.navigationId` is now the soft nav ID for all metrics)
3. Hard navigate away (e.g. to `http://example.com`; page enters bfcache)
4. Press back (bfcache restore fires `pageshow` with `persisted: true`)
5. Trigger any interaction (forces INP to report)

**Expected:** All metrics (CLS, LCP, INP) report `http://localhost/about`

**Actual:** Only INP reports `http://localhost` (the original hard-nav URL)

This affects any SPA using soft navigations with bfcache enabled, which covers virtually all production deployments.

### Root cause

In `onINP.ts`, the `onBFCacheRestore` callback creates a metric with the correct `navigationId` via `initNewINPMetric`, but immediately overwrites both `metric` and `report` before `doubleRAF` fires:

```ts
onBFCacheRestore(() => {
  interactionManager._resetInteractions();
  initNewINPMetric('back-forward-cache', metric.navigationId); // metric A: navigationId = softNavId ✓
  doubleRAF(() => report());   // captures `report` by reference, not yet updated

  metric = initMetric('INP'); // metric B: navigationId = hardNavId ✗ (overwrites A)
  report = bindReporter(       // report now points to reporter B
    onReport, metric, INPThresholds, opts.reportAllChanges,
  );
  // doubleRAF fires reporter B, but metric B.value === -1,
  // so bindReporter's `if (metric.value >= 0)` guard suppresses the report entirely.
});
```

By contrast, `onCLS` correctly captures the reporter value at call time rather than via closure:

```ts
onBFCacheRestore(() => {
  initNewCLSMetric('back-forward-cache', metric.navigationId);
  doubleRAF(report); // value captured at call time ✓
});
```

## Proposed fix

Follow the same pattern as `onCLS`: capture the `report` value directly after `initNewINPMetric` updates it, and remove the subsequent re-initialization that overwrites `metric` and `report` with a stale hard-nav context.

```ts
onBFCacheRestore(() => {
  initNewINPMetric('back-forward-cache', metric.navigationId);
  doubleRAF(report); // capture value, not reference
});
```

After this change:
- `metric` and `report` remain bound to metric A (soft nav URL) after the callback returns.
- Subsequent interactions accumulate into metric A and are reported with the correct URL.
- `doubleRAF` fires reporter A, which is suppressed if no interactions have occurred yet; the same behavior as CLS.

Mimori256:
resolves #734 

## Problem

When an SPA performs a soft navigation and the user then navigates away and returns via bfcache restore, INP reports the original hard-navigation URL instead of the URL that was active when the page was cached. CLS and LCP are unaffected.

**Steps to reproduce:**
1. Hard-navigate to an SPA (e.g. `http://localhost`)
2. Perform a soft navigation (e.g. to `/about`)
3. Hard-navigate away (page enters bfcache)
4. Press Back (bfcache restore fires)
5. Trigger any interaction to force INP to report

**Expected:** INP `navigationURL` = `http://localhost/about`
**Actual:** INP `navigationURL` = `http://localhost`

## Root cause

`onBFCacheRestore` in `onINP.ts` called `initNewINPMetric`, which correctly updated `metric` and `report` to the bfcache context, but then immediately overwrote both with a fresh hard-navigation metric:

```ts
onBFCacheRestore(() => {
  initNewINPMetric('back-forward-cache', metric.navigationId); // metric A ✓
  doubleRAF(() => report()); // reference — resolves to reporter B at fire time ✗

  metric = initMetric('INP');       // metric B: hard-nav, wrong URL ✗
  report = bindReporter(...);       // reporter B ✗
});
```

When `doubleRAF` fired it called reporter B (whose `metric.value` is `-1`), so the report was silently suppressed. All subsequent interactions accumulated into metric B and were reported with the original hard-nav URL.

## Fix

Follow the same pattern `onCLS` already uses: pass `report` by value (not by reference) immediately after `initNewINPMetric` updates it, and remove the redundant re-initialization that overwrote the correct context.

```ts
onBFCacheRestore(() => {
  initNewINPMetric('back-forward-cache', metric.navigationId);
  doubleRAF(report); // value captured at call time ✓
});
```

`initNewINPMetric` already calls `interactionManager._resetInteractions()` internally, so the explicit call that preceded it was also redundant and has been removed.

### DIFF
--- a/src/onINP.ts
+++ b/src/onINP.ts
@@ -184,17 +184,8 @@ export const onINP = (
       // Only report after a bfcache restore if the `PerformanceObserver`
       // successfully registered.
       onBFCacheRestore(() => {
-        interactionManager._resetInteractions();
         initNewINPMetric('back-forward-cache', metric.navigationId);
-        doubleRAF(() => report());
-
-        metric = initMetric('INP');
-        report = bindReporter(
-          onReport,
-          metric,
-          INPThresholds,
-          opts.reportAllChanges,
-        );
+        doubleRAF(report);
       });
     }
   });
--- a/src/onINP.ts
+++ b/src/onINP.ts
@@ -185,7 +185,6 @@ export const onINP = (
       // successfully registered.
       onBFCacheRestore(() => {
         initNewINPMetric('back-forward-cache', metric.navigationId);
-        doubleRAF(report);
       });
     }
   });
--- a/src/onINP.ts
+++ b/src/onINP.ts
@@ -16,7 +16,6 @@
 
 import {onBFCacheRestore} from './lib/bfcache.js';
 import {bindReporter} from './lib/bindReporter.js';
-import {doubleRAF} from './lib/doubleRAF.js';
 import {initMetric} from './lib/initMetric.js';
 import {initUnique} from './lib/initUnique.js';
 import {InteractionManager} from './lib/InteractionManager.js';


src/onINP.ts
```diff
@@ -184,17 +184,8 @@ export const onINP = (
...
-        doubleRAF(() => report());
-
-        metric = initMetric('INP');
-        report = bindReporter(
-          onReport,
-          metric,
-          INPThresholds,
-          opts.reportAllChanges,
-        );
+        doubleRAF(report);
```
tunetheweb: Thanks I don't think you need this line either? Do you agree?

Unlike CLS, INP doesn't always report so can wait until an actual interaction. It's also not in the [`main` branch version](https://github.com/GoogleChrome/web-vitals/blob/8e3554212e323a89d7a348c219dc86d0a570edbb/src/onINP.ts#L142).

Technically, CLS doesn't need to report a 0 value either, but in #297 we added an explicit report for that to deal with an FCP/CLS timing issue. It's no harm to report a 0 CLS (since that will be reported at the end of the session anyway if there are no other CLS reports), whereas reporting a 0 INP is unexpected in the case where there are no interactions.

```suggestion
```

tunetheweb (changes requested):

src/onINP.ts
Mimori256: Oh good point. INP initialises to -1 and `bindReporter` supresses reports when `value < 0`, so the `doubleRAF` would always be a no-op here

tunetheweb (approved):

src/onINP.ts
tunetheweb: doubleRAF is no longer needed.

```suggestion
```

PR #739, opened 2026-05-11
soft-navs: Replace filter() with find()/some() for early termination

Mimori256:
### summary
- Replace `filter()[0]` with `find()` in `getSoftNavigationEntry()` (`src/lib/softNavs.ts`)
- Replace `filter().length === 0` with `!some()` in `onINP` entry handler (`src/onINP.ts`)

Both changes are functionally equivalent but improve clarity and efficiency
by short-circuiting on first match instead of iterating the entire array.

### Details

`softNavs.ts`: `filter()` returns an array, so the existing `if (softNavEntry)` guard was always truthy
```js
Boolean([]) // ← returns true
``` 
the code worked only because `array[0]` is `undefined` on empty results. Using `find()` makes the intent explicit and avoids the misleading branch.

`onINP.ts`: `filter(...).length === 0` allocates an intermediate array just to check emptiness. `!some(...)` returns as soon as a matching entry is found.

### DIFF
--- a/src/lib/softNavs.ts
+++ b/src/lib/softNavs.ts
@@ -29,10 +29,7 @@ export const getSoftNavigationEntry = (
 ): SoftNavigationEntry | undefined => {
   if (!navigationId) return;
 
-  const softNavEntry = window.performance
+  return window.performance
     .getEntriesByType('soft-navigation')
-    .filter((entry) => entry.navigationId === navigationId);
-  if (softNavEntry) return softNavEntry[0];
-
-  return;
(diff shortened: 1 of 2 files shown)


tunetheweb (approved):
LGTM

tunetheweb:
BTW removed this one:
 
 > Replace filter()[0] with find() in getSoftNavigationEntry() (src/lib/softNavs.ts)
 
 As `getSoftNavigationEntry()` has been removed in the latest build.

PR #747, opened 2026-06-08
Fix getSelector to not throw

sliftist:
Only elements have class lists, apparently, so you need to actually check for if the node type is of element, not if it's not of type document.

I'm unsure if this is actually the source code that's included in Google Chrome, in which case it's enabled forcefully by default with no option of disabling. And so, if we have any exceptions, even if we catch them, it breaks the break on caught exceptions in DevTools, which breaks that feature for every website on the internet. In which case, Web Vitals needs to never have any exceptions, even if they're caught, ever.

However, this might not even be the code that's included in Google Chrome, in which case it doesn't matter.

### DIFF
--- a/src/lib/getSelector.ts
+++ b/src/lib/getSelector.ts
@@ -27,7 +27,7 @@ export const getSelector = (node: Node | null) => {
   let sel = '';
 
   try {
-    while (node?.nodeType === 1) {
+    while (node?.nodeType !== 9) {
       const el: Element = node as Element;
       const part = el.id
         ? '#' + el.id
--- a/src/lib/getSelector.ts
+++ b/src/lib/getSelector.ts
@@ -27,7 +27,7 @@ export const getSelector = (node: Node | null) => {
   let sel = '';
 
   try {
-    while (node?.nodeType !== 9) {
+    while (node?.nodeType === 1) {
       const el: Element = node as Element;
       const part = el.id
         ? '#' + el.id
--- a/src/lib/getSelector.ts
+++ b/src/lib/getSelector.ts
@@ -31,7 +31,7 @@ export const getSelector = (node: Node | null) => {
       const el: Element = node as Element;
       const part = el.id
         ? '#' + el.id
-        : [getName(el), ...Array.from(el.classList).sort()].join('.');
+        : [getName(el), ...Array.from(el.classList ?? []).sort()].join('.');
       if (sel.length + part.length > MAX_LEN - 1) {
         return sel || part;
       }


tunetheweb:
Only Nodes and Document Nodes will be eligible for anything that calls `getSelector`.

Can you explain what problem you're fixing here? Do you have examples of this throwing? Or is this just "maybe a problem"?

sliftist:
Yes, I'm running into examples of it throwing. It's making debugging my application extremely difficult because it keeps throwing caught exceptions.

The issue is it can receive a TextNode from LayoutShiftAttribute: https://developer.mozilla.org/en-US/docs/Web/API/LayoutShiftAttribution/node .  The specification might have changed from an element to a node. https://developer.mozilla.org/en-US/docs/Web/API/LayoutShiftAttribution says: `Returns the element that has shifted (null if it has been removed).`, however LayoutShiftAttribution correctly describes it as a `node`.


This is roughly the call stack. It's hard to find it exactly because the code is minified and has no source map.

observer.ts:48
    callback(list.getEntries() as PerformanceEntryMap[K]);
LayoutShiftManager.ts:48
    this._onAfterProcessingUnexpectedShift?.(entry);
onCLS.ts:79
    const customTarget = opts.generateTarget?.(node) ?? getSelector(node);

tunetheweb:
OK then I'm not adverse to adding this change but I wonder why the `catch` doesn't catch this?

It would also be good to add a unit test if we can.

sliftist:
The catch catches it fine. The issue is that this code, or some variation of it, is injected by devtools by default, with apparently no way to disable it.

And so any *caught* errors in any of this code effectively breaks `Pause on caught exceptions` debugging in devtools.

It's such a minor issue. It's not even technically a bug. But any noise in this library is now massively amplified by its forceful inclusion by devtools.

tunetheweb:
Ah OK. Yes it is included in DevTools in the Performance Panel for [the "Live Metrics" view](https://developer.chrome.com/blog/devtools-realtime-cwv) to provide basic performance details of Core Web Vitals without having to take a full performance trace (which is left for more detailed performance analysis once the performance issue has successfully been reproduced).

OK I can see how the `Pause on caught exceptions` would be affected by this.

However, I'd like to suggest a different change, to still allow the parent Nodes to be selected. Will comment on the PR.

src/lib/getSelector.ts
```diff
@@ -27,7 +27,7 @@ export const getSelector = (node: Node | null) => {
   let sel = '';
 
   try {
-    while (node?.nodeType !== 9) {
+    while (node?.nodeType === 1) {
```
tunetheweb: Let's leave this in place to allow this code to be entered and iterate up to parents:

```suggestion
    while (node?.nodeType !== 9) {
```

src/lib/getSelector.ts
tunetheweb:
```suggestion
        : [getName(el), ...Array.from(el.classList ?? []).sort()].join('.');
```

PR #755, opened 2026-06-26
Cap RequestIdleCallback to 1 second to ensure metrics report even when busy

### LINKED ISSUE #754: INP not updated in DevTools for very busy apps
We had a report that INP was not updating in Live Metrics in DevTools.

This turned out to be due to INP using idle-until-urgent and the main thread never becoming idle.

This will also potentially affect anyone else using `reportAllChanges: true` (not remember for busy apps anyway!)

A simple solution would be to cap the `requestIdleCallback` to some reasonable limit (1 second?) so it always reports even on busy apps.

tunetheweb:
Fixes #754

### DIFF
--- a/src/lib/whenIdleOrHidden.ts
+++ b/src/lib/whenIdleOrHidden.ts
@@ -29,7 +29,12 @@ export const whenIdleOrHidden = (cb: () => void) => {
     }
     return setTimeout(callback);
   };
-  const cIC = globalThis.cancelIdleCallback || clearTimeout;
+  const cIC = (ref: number | ReturnType<typeof setTimeout>) => {
+    if ('requestIdleCallback' in globalThis && typeof ref === 'number') {
+      return globalThis.cancelIdleCallback(ref);
+    }
+    return clearTimeout(ref);
+  };
 
   // If the document is hidden, run the callback immediately, otherwise
   // race an idle callback with the next `visibilitychange` event.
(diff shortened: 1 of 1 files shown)


mmocny (approved):

src/lib/whenIdleOrHidden.ts
```diff
@@ -21,6 +21,11 @@ import {runOnce} from './runOnce.js';
  * if the browser's visibility state is (or becomes) hidden.
  */
 export const whenIdleOrHidden = (cb: () => void) => {
+  // Cap the requestIdleCallback to 1 sec for very busy apps
+  // https://github.com/GoogleChrome/web-vitals/issues/754
+  // If not using rIC, then the setTimeout timeout should be 0
+  const timeout = 'requestIdleCallback' in globalThis ? 1000 : 0;
+
   const rIC = globalThis.requestIdleCallback || setTimeout;
```
mmocny, line 29: Drive by comment: setTimeout fallback, without rAF+setTimeout and 0ms delay, will potentially get scheduled before next paint whenever next paint is vsync aligned and we have an empty task queue.

(Maybe that doesn't matter here if the continuation is fast)

mmocny (approved):

src/lib/whenIdleOrHidden.ts
tunetheweb: Fair. But there's a question as to how complicated fallbacks should be.

Maybe browsers that don't implement rIC should be encouraged to rather than worked around as best possible so they see the need for it less 😁

So gonna leave as is for simplicity.

PR #759, opened 2026-07-16
Fix attribution and navigationId issues in soft-navigation support

philipwalton:
This PR addresses three distinct issues in the soft-navigation and attribution logic: correcting a double-subtraction of `activationStart` in LCP attribution, preventing stale soft-navigation attribution after back-forward cache (bfcache) restores in FCP and LCP, and preserving the initial `navigationId` for FCP when paint entries omit `navigationId`.

---

### Problem
In `src/attribution/onLCP.ts`, `lcpRequestStart` is calculated as an activation-relative time:

```ts
const lcpRequestStart = Math.max(
  ttfb,
  lcpResourceEntry
    ? (lcpResourceEntry.requestStart || lcpResourceEntry.startTime) - activationStart
    : 0,
);
```

Since `ttfb` is `Math.max(0, responseStart - activationStart)`, `lcpRequestStart` is already relative to `activationStart`. However, `lcpResponseEnd` subtracted `activationStart` from `lcpRequestStart` a second time:

```ts
const lcpResponseEnd = Math.min(
  metric.value,
  Math.max(
    lcpRequestStart - activationStart, // bug: activationStart subtracted twice
    lcpResourceEntry ? lcpResourceEntry.responseEnd - activationStart : 0,
    0,
  ),
);
```

On prerendered pages (where `activationStart > 0`) with element-only LCPs (or resource entries starting early), `lcpRequestStart - activationStart` yielded a negative number, pulling `lcpResponseEnd` down to `0`. This resulted in negative `resourceLoadDuration` (`lcpResponseEnd - lcpRequestStart`) and artificially inflated `elementRenderDelay`.

### Fix
Use `lcpRequestStart` directly in the `lcpResponseEnd` calculation without subtracting `activationStart` again:

```ts
const lcpResponseEnd = Math.min(
  metric.value,
  Math.max(
    lcpRequestStart,
    lcpResourceEntry ? lcpResourceEntry.responseEnd - activationStart : 0,
    0,
  ),
);
```

Additionally, simplified `activationStart` and `responseStart` extractions using optional chaining and nullish coalescing (`navigationEntry?.activationStart ?? 0` and `navigationEntry?.responseStart ?? 0`).

---

### Problem
In `src/attribution/onFCP.ts` and `src/attribution/onLCP.ts`, the decision to take the hard-navigation branch vs. the soft-navigation branch was made by checking whether `metric.navigationId` matched the initial navigation ID (`hardNavId`):

```ts
const hardNavId = getNavigationEntry()?.navigationId || 0;
if (!metric.navigationId || metric.navigationId === hardNavId) {
  // hard-nav branch
} else {
  // soft-nav lookup
}
```

When a bfcache restore occurs following a soft navigation, `onBFCacheRestore` generates a metric with `navigationType = 'back-forward-cache'` while retaining the `navigationId` of the prior soft navigation. Because `metric.navigationId` differed from `hardNavId`, the metric incorrectly took the soft-navigation branch and was attributed against a stale soft-navigation entry.

### Fix
Branch directly on `metric.navigationType` rather than comparing navigation IDs:

```ts
if (metric.navigationType !== 'soft-navigation') {
  // hard-nav / default attribution branch
} else {
  // soft-navigation map lookup
}
```

This change is applied to both `src/attribution/onFCP.ts` and `src/attribution/onLCP.ts`. Metrics resulting from bfcache restores (`navigationType === 'back-forward-cache'`) now enter the hard-navigation branch, returning default attribution objects as intended. The unused `hardNavId` variables have been removed.

---

### Problem
In `src/onFCP.ts`, `handleEntries` previously executed:

```ts
metric.navigationId = entry.navigationId || 0;
```

In browsers without soft-navigations support (or for standard `first-contentful-paint` entries that do not carry a `navigationId`), `entry.navigationId` is `undefined`. This unconditionally reset `metric.navigationId` to `0`, clobbering the valid ID assigned during `initMetric('FCP')`.

### Fix
Preserve the existing `metric.navigationId` if `entry.navigationId` is falsy:

```ts
metric.navigationId = entry.navigationId || metric.navigationId;
```

---

## Summary of Changes

- **`src/attribution/onLCP.ts`**:
  - Removed double-subtraction of `activationStart` in `lcpResponseEnd` calculation.
  - Replaced `!metric.navigationId || metric.navigationId === hardNavId` check with `metric.navigationType !== 'soft-navigation'`.
  - Cleaned up navigation entry property lookups using optional chaining.

- **`src/attribution/onFCP.ts`**:
  - Replaced `!metric.navigationId || metric.navigationId === hardNavId` check with `metric.navigationType !== 'soft-navigation'`.

- **`src/onFCP.ts`**:
  - Updated `metric.navigationId` assignment to fall back to `metric.navigationId` instead of `0`.

### DIFF
--- a/src/attribution/onFCP.ts
+++ b/src/attribution/onFCP.ts
@@ -43,7 +43,6 @@ export const onFCP = (
   }
 
   const attributeFCP = (metric: FCPMetric): FCPMetricWithAttribution => {
-    const hardNavId = getNavigationEntry()?.navigationId || 0;
     // Use a default object if no other attribution has been set.
     let attribution: FCPAttribution = {
       timeToFirstByte: 0,
@@ -51,7 +50,7 @@ export const onFCP = (
       loadState: getLoadState(getBFCacheRestoreTime()),
     };
 
-    if (!metric.navigationId || metric.navigationId === hardNavId) {
+    if (metric.navigationType !== 'soft-navigation') {
       if (metric.entries.length) {
         const navigationEntry = getNavigationEntry();
         const fcpEntry = metric.entries.at(-1);
--- a/src/attribution/onLCP.ts
+++ b/src/attribution/onLCP.ts
@@ -70,7 +70,6 @@ export const onLCP = (
   };
 
   const attributeLCP = (metric: LCPMetric): LCPMetricWithAttribution => {
-    const hardNavId = getNavigationEntry()?.navigationId || 0;
     // Use a default object if no other attribution has been set.
     let attribution: LCPAttribution = {
       timeToFirstByte: 0,
@@ -104,16 +103,10 @@ export const onLCP = (
       let activationStart = 0;
       let responseStart = 0;
 
-      if (!metric.navigationId || metric.navigationId === hardNavId) {
+      if (metric.navigationType !== 'soft-navigation') {
         navigationEntry = getNavigationEntry();
-        activationStart =
-          navigationEntry && navigationEntry.activationStart
-            ? navigationEntry.activationStart
-            : 0;
-        responseStart =
-          navigationEntry && navigationEntry.responseStart
-            ? navigationEntry.responseStart
-            : 0;
+        activationStart = navigationEntry?.activationStart ?? 0;
+        responseStart = navigationEntry?.responseStart ?? 0;
       } else {
         // Set activationStart to the navigation start time
         activationStart = metric.navigationStartTime || 0;
@@ -140,7 +133,7 @@ export const onLCP = (
           // Cap at LCP time (videos continue downloading after LCP for example)
           metric.value,
           Math.max(
-            lcpRequestStart - activationStart,
+            lcpRequestStart,
             lcpResourceEntry
               ? lcpResourceEntry.responseEnd - activationStart
               : 0,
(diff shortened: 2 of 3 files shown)


mmocny:
For (2): makes sense, but interestingly a related Q was raised on web perf slack just recently about treating bf-cache restorations "as soft navs".  While I dont literally think that we should emit soft-nav entries, we do have an experiment where a bfcache events are reported to perf timeline (today, as a dedicated entry, but perhaps it could also be an event timing? alongside other page lifecycle events?) and perhaps these should also update `navigationId` values.

---

For (3): I dont understand what:

> clobbering the valid ID assigned during initMetric('FCP')

means?  if the new entry does not have a `navigationId` doesn't that mean that this browser does not support soft-navs nor `navigationId` and none of the pathways would have a valid ID assigned?  Or do we default to something just for convenience?

tunetheweb:
Yeah 3) is a non-issue as the default navId will have been 0 anyway (which is what will have been set up when initialising the metric anyway).

Then again, it's no harm to leave the initialised one in case this every changes in future (e.g. a bfcache restore initialises a dummy navId, and then an FCP entry happens afterwards for some reason if a browser starts emitting FCP on bfcache, or the logic is refactored to call this `handleEntries` for some reason).

tunetheweb (approved):

PR #761, opened 2026-07-16
Clean up soft navigations implementation

philipwalton:
This PR performs a behavior-preserving cleanup pass across the soft navigations implementation added on the `soft-navs` feature branch, based on code review feedback. It eliminates dead code, extracts redundant map-management logic, resolves variable shadowing, and updates outdated comments.

All changes preserve existing functionality, with one intentional micro-optimization in `src/lib/observe.ts` to skip sorting entries for single-type observers.

## Key Changes

- **`src/lib/observe.ts`**:
  - Replaced redundant `Object.assign({...opts}, opts || {})` in `po.observe()` with a clean object literal `{ type: t, buffered: true, ...opts }`. `opts` already defaults to `{}` in the signature.
  - Gated `entries.sort()` to run only when `supportedTypes.length > 1`. Single-type performance observers already receive entries in chronological order, so skipping sort for single-type observers avoids unnecessary overhead.

- **`src/lib/softNavs.ts`**:
  - Extracted `storeSoftNavEntry(map, entry)` helper to handle inserting soft navigation entries into `_softNavigationEntryMap` and pruning the map size to the 2 most recent entries.

- **`src/onFCP.ts` & `src/onLCP.ts`**:
  - Replaced inline map insertion and pruning blocks with calls to `storeSoftNavEntry()`.
  - Removed dead `opts = opts || {}` reassignments in `src/onFCP.ts` and `src/onTTFB.ts`.

- **`src/attribution/onTTFB.ts` & `src/onTTFB.ts`**:
  - Simplified `metric?.entries?.[0]` to `metric.entries[0]`, as `metric` is guaranteed to be a newly initialized metric object with an `entries` array on this code path.
  - Removed redundant `opts = opts || {}` default check in `src/onTTFB.ts`.

- **`src/onLCP.ts`**:
  - Renamed inner variable `entries` to `metricEntries` inside `handleEntries()` to eliminate shadowing of the outer function parameter `entries`.

- **`src/lib/initMetric.ts`**:
  - Reduced back-to-back duplicate calls of `getNavigationEntry()` down to a single call, storing and reusing `hardNavEntry`.

- **Comments & Test Fixtures (`src/types.ts`, `src/onLCP.ts`, `test/views/layout.njk`)**:
  - Rewrote comments that referenced outdated origin-trial transition state to clearly explain current fallback mechanisms and cleanup prerequisites.
  - Simplified `document.nav = ++document.nav` to `document.nav++` in test template.

### DIFF
--- a/src/onLCP.ts
+++ b/src/onLCP.ts
@@ -172,9 +172,9 @@ export const onLCP = (
             continue;
           }
 
-          // The LCP entry is exposed either nested under
-          // `largestContentfulPaint` or directly via `renderTime`, depending
-          // on the implementation, so handle both.
+          // Older implementations expose the render time directly on the entry
+          // rather than nested under `largestContentfulPaint`.
+          // TODO: remove after the origin trial ends.
           const renderTime =
             ICPEntry.largestContentfulPaint?.renderTime ||
             ICPEntry.renderTime ||
--- a/src/onLCP.ts
+++ b/src/onLCP.ts
@@ -19,7 +19,7 @@ import {onBFCacheRestore} from './lib/bfcache.js';
 import {bindReporter} from './lib/bindReporter.js';
 import {doubleRAF} from './lib/doubleRAF.js';
 import {getActivationStart} from './lib/getActivationStart.js';
-import {checkSoftNavsEnabled} from './lib/softNavs.js';
+import {checkSoftNavsEnabled, storeSoftNavEntry} from './lib/softNavs.js';
 import {getVisibilityWatcher} from './lib/getVisibilityWatcher.js';
 import {initMetric} from './lib/initMetric.js';
 import {initUnique} from './lib/initUnique.js';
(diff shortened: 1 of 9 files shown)


src/onLCP.ts
```diff
@@ -182,8 +172,9 @@ export const onLCP = (
             continue;
           }
 
-          // We're changing the shape to nest LCP entries within
-          // interaction-contentful-paint so handle both for now
+          // The LCP entry is exposed either nested under
+          // `largestContentfulPaint` or directly via `renderTime`, depending
+          // on the implementation, so handle both.
```
tunetheweb:
```suggestion
          // Older implementations expose the render time directly on the entry
          // rather than nested under `largestContentfulPaint`.
          // TODO: remove after the origin trial ends.
```

tunetheweb (approved):
LGTM with a nit

PR #764, opened 2026-07-17
Prep CHANGELOG and unique id prefix for v6.0.0

tunetheweb:
Updates to prep for v6

PR #766, opened 2026-07-17
Update unique id to v6

tunetheweb:


tunetheweb:
Merging this into #764 instead

PR #772, opened 2026-07-27
Prevent JavaScript error if PerformanceObserver not available

steffenweber:
Some Firefox users seem to disable the PerformanceObserver API which causes JavaScript errors: "ReferenceError: PerformanceObserver is not defined"

### DIFF
--- a/src/lib/softNavs.ts
+++ b/src/lib/softNavs.ts
@@ -18,6 +18,7 @@ import type {ReportOpts} from '../types.js';
 
 export const checkSoftNavsEnabled = (opts?: ReportOpts) => {
   return (
+    'PerformanceObserver' in window &&
     PerformanceObserver.supportedEntryTypes.includes('soft-navigation') &&
     // Older implementations expose the value as an attribute rather than the
     // method. We only support the newer method as that was what was launched
--- a/src/lib/softNavs.ts
+++ b/src/lib/softNavs.ts
@@ -19,7 +19,9 @@ import type {ReportOpts} from '../types.js';
 export const checkSoftNavsEnabled = (opts?: ReportOpts) => {
   return (
     // Firefox has a preference to disable this, which some people use so add a guard
-    globalThis.PerformanceObserver?.supportedEntryTypes.includes('soft-navigation') &&
+    globalThis.PerformanceObserver?.supportedEntryTypes.includes(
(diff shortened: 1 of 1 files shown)


src/lib/softNavs.ts
```diff
@@ -18,6 +18,7 @@ import type {ReportOpts} from '../types.js';
 
 export const checkSoftNavsEnabled = (opts?: ReportOpts) => {
   return (
+    'PerformanceObserver' in window &&
     PerformanceObserver.supportedEntryTypes.includes('soft-navigation') &&
```
tunetheweb: I think we can shortcut this:

```suggestion
PerformanceObserver?.supportedEntryTypes.includes('soft-navigation') &&
```

Alternatively `observe.ts` has had a PerformanceObserver call for a few years but is covered by a `try/catch` so guessing that's why it's not caused problems there. Maybe we should do same here for consistency? Or change both to `PerformanceObserver?.` for less code (probably my peference).

Any opinions here?

tunetheweb (changes requested):

src/lib/softNavs.ts
tunetheweb: Actually we can't get rid of the `try/catch` as easily in `observe.ts`. So let's go with the change here, but it this needs a slight tweak:

```suggestion
    // Firefox has a preference to disable this, which some people use so add a guard
    globalThis.PerformanceObserver?.supportedEntryTypes.includes('soft-navigation') &&
```

src/lib/softNavs.ts
```diff
@@ -18,7 +18,8 @@ import type {ReportOpts} from '../types.js';
 
 export const checkSoftNavsEnabled = (opts?: ReportOpts) => {
   return (
-    PerformanceObserver.supportedEntryTypes.includes('soft-navigation') &&
+    // Firefox has a preference to disable this, which some people use so add a guard
+    globalThis.PerformanceObserver?.supportedEntryTypes.includes('soft-navigation') &&
```
tunetheweb:
```suggestion
    globalThis.PerformanceObserver?.supportedEntryTypes.includes(
      'soft-navigation',
    ) &&
```

steffenweber:
Looks good to me, thank you!

PR #776, opened 2026-08-05
Add Resource Timing buffer for LCP attribution

### LINKED ISSUE #775: LCP attribution loses `lcpResourceEntry` on Resource Timing buffer overflow
**Summary**

In the attribution build, the `lcpResourceEntry` is resolved via `performance.getEntriesByType('resource')`. Because the Resource Timing buffer is capped (250 entries by default) and silently drops entries once full, the LCP resource may no longer be present by the time attribution looks it up — so `lcpResourceEntry` comes back `undefined` even though a resource entry did exist. This is much more likely on long-lived SPA pages, where a soft-navigation LCP happens after the buffer has already filled.

(Context: raised this in the web-perf Slack; confirmed it's unintentional and suggested opening an issue so it isn't forgotten.)

**Impact**

We use web-vitals for RUM and rely on `lcpResourceEntry` for LCP attribution (`requestStart` / `responseEnd`, `loadDelay` / `loadDuration`, and inferring the LCP type). When it's `undefined` we lose all of that for the affected page impressions. For hard navigations this is rare (if your LCP resource isn't in the first ~250 resources you likely have bigger problems), but for soft navigations deep in a session it's common.

**Steps to reproduce**

1. Load an SPA page and let it fetch more than 250 resources over its lifetime (so the Resource Timing buffer fills).
2. Perform a soft navigation whose LCP element is an image fetched after the buffer filled.
3. Observe the LCP attribution: `attribution.lcpResourceEntry` is `undefined`, even though the resource was fetched and had a `PerformanceResourceTiming` entry at fetch time.

**Expected:** The LCP resource entry is available for attribution regardless of how many resources the page has already loaded.

**Actual:** `lcpResourceEntry` is `undefined` once the Resource Timing buffer has overflowed, because `getEntriesByType('resource')` only returns what's currently retained in the (capped) buffer.

**Environment:** web-vitals ^6.0.1, Chrome 151+.

tunetheweb:
Fixes #775


src/attribution/onLCP.ts
```diff
@@ -83,9 +96,10 @@ export const onLCP = (
       const lcpEntry = metric.entries.at(-1)!;
       const lcpResourceEntry =
         lcpEntry.url &&
-        performance
+        (performance
           .getEntriesByType('resource')
-          .find((e) => e.name === lcpEntry.url);
+          .find((e) => e.name === lcpEntry.url) ||
+          resourceBuffer.find((e) => e.name === lcpEntry.url));
```
mmocny: What about swapping the order of these two lines, so you don't need to scan the whole buffer if its in the list of recents?

mmocny (approved):

src/attribution/onLCP.ts
tunetheweb: Good idea - done!

PR #778, opened 2026-08-06
Add and export resetCLS to reset CLS metric

srubin:
Mirrors the resetINP change (#1, #2) for CLS: dispatching a `reset-web-vitals-cls` event re-initializes the CLS metric and its session-window state while keeping existing callbacks, so consumers can report CLS in fixed intervals (Descript reports the worst session burst per 5-minute window, like INP).

Exported from both the standard and attribution builds, and bumps the version to `3.5.2-descript.3`.

Includes an e2e test mirroring 'resets INP metric'; I authored it to match the existing suite but have not run the test suite locally. After merge this needs an npm publish of `3.5.2-descript.3` so the monorepo can pick it up (BIX-8613).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ For details on the difference between the builds, see <a href="#which-build-is-r
 To load the "standard" build, import modules from the `web-vitals` package in your application code (as you would with any npm package and node-based build tool):
 
 ```js
-import {onLCP, onFID, onCLS} from 'web-vitals';
+import {onLCP, onFID, onCLS} from '@descript/web-vitals';
 
 onCLS(console.log);
 onFID(console.log);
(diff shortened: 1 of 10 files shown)


srubin:
Apologies — opened against upstream by mistake; this is intended for Descript's fork.

tunetheweb:
Interesting! Have you considered the soft nav support added in v6?

Though that would be Chromium-only whereas INP is available in Safari and Firefox so the manual reset would alow them to be supported for INP (they don't support CLS).

srubin:
Oh, TIL. But no, I don't think soft navs are sufficient for our use case, because the URL isn't changing. In our case, we are tracking these metrics for each 5 minute segment of an hours-long audio/video editing session.

tunetheweb:
Oh gotcha. Yeah that is a different use case.

PR #783, opened 2026-08-14
Scope the navigation interaction count to each InteractionManager

### LINKED ISSUE #782: [soft-navs] Only the first onINP() instance reports when several are registered
`prevInteractionCount` in `lib/InteractionManager.ts` is module state, but `_resetInteractions()` is an instance method that writes to it. Since `initUnique()` gives every `onINP()` call its own `InteractionManager`, multiple instances on one page share that counter.

The visible effect with `reportSoftNavs: true` is that only the **first** instance to process a `soft-navigation` entry reports an INP for that navigation. Every other instance silently reports nothing.

This appeared when I had a Sentry SDK using the web vitals module along another RUM that uses the same web vitals module. So the first consumer will get reports correctly while the subsequent consumers will get under-reported INPs.

### Repro

```js
import {onINP} from 'web-vitals';

onINP(m => console.log('A', m.value, m.navigationType, m.navigationId), {reportSoftNavs: true});
onINP(m => console.log('B', m.value, m.navigationType, m.navigationId), {reportSoftNavs: true});
```

**Expected:** `A` and `B` both log an INP per soft navigation.

**Actual:** only `A` logs. `B` never fires.

I think changing the `_prevInteractionCount` to be an instance member instead should fix this, submitting a PR for it.

logaretm:
Moves `prevInteractionCount` to a per-instance `InteractionManager` field so multiple `onINP()` calls stop resetting each other's interaction count.

I added assertions for this case as well.

closes #782


tunetheweb (approved):
Thanks!

tunetheweb:
Thanks for spotting and fixing that! And for the test case.

This fix has been included in v6.1.1

PR #789, opened 2026-08-25
Fix negative inputDelay in INP attribution

### LINKED ISSUE #788: INP attribution can report a negative `inputDelay`
`attribution.inputDelay` can be reported as a negative number, with `processingDuration` inflated by exactly the same magnitude.

`src/attribution/onINP.ts:417` reads:

```js
const processingStart = group.processingStart;
```

and line 457 derives `inputDelay: processingStart - firstEntry.startTime`.

The problem is the scope of `group`. `groupEntriesByRenderTime()` buckets every `event` entry whose render time is within 8ms into one frame group, taking a frame-wide minimum (lines 179-181):

```js
group.processingStart = Math.min(entry.processingStart, group.processingStart);
```

That group is not limited to the INP interaction. `groupEntriesByRenderTime` is installed as `interactionManager._onBeforeProcessingEntry` (line 275), and `InteractionManager._processEntry` calls that hook at line 115 — *before* the guard on line 118:

```js
this._onBeforeProcessingEntry?.(entry);

// Skip further processing for entries that cannot be INP candidates.
if (!(entry.interactionId || entry.entryType === 'first-input')) return;
```

So entries with `interactionId === 0`, and entries belonging to a *different* interaction that painted in the same frame, both contribute to `group.processingStart`. When any of them began processing before the INP interaction's hardware `startTime`, `inputDelay` goes negative.

Two realistic triggers:

1. A long `pointerover` / `mouseover` handler still blocking when the user clicks — Event Timing exposes these with `interactionId === 0`.
2. A rage-click pair where two interactions paint in one frame and the later, shorter one is the p98 pick.

This came from #748, which changed the line from `firstEntry.processingStart` to `group.processingStart`. That PR's reasoning was sound for its stated purpose — "don't assume the first entry is the processingStart and instead use the group one", for entries emitted out of order *within* one interaction — but `group` also spans events that aren't part of the interaction at all.

**Impact**

Negative `inputDelay` flows into RUM dashboards, and `processingDuration` is overstated by the same amount, so the breakdown misattributes input delay as processing time. The `longestScript.subpart` bucket boundaries (lines 353-358) are derived from `interactionTime + inputDelay`, so script attribution is affected too.

Note the three subparts still sum exactly to `metric.value` — they telescope — so the values are individually wrong rather than internally inconsistent. That's likely why this hasn't surfaced as an obvious contradiction.

**Steps to reproduce**

The required shape is a long non-interaction event and an interaction presented in the same frame, where the interaction's hardware `startTime` falls after the non-interaction event's `processingStart`:

| name | startTime | duration | processingStart | processingEnd | renderTime | interactionId |
|---|---|---|---|---|---|---|
| `pointerover` | 1000 | 208 | 1002 | 1160 | 1208 | **0** |
| `pointerdown` | 1100 | 104 | 1160 | 1175 | 1204 | 4001 |

`|1208 - 1204| = 4 ≤ 8`, so they group. `group.processingStart` becomes `min(1002, 1160) = 1002`, while `firstEntry.startTime` is `1100`.

Driving the shipped `dist/modules/attribution/onINP.js` with those two entries and stubbed browser globals — the entry values are synthetic, but all grouping and attribution logic is the real built code:

**Expected:** `inputDelay: 60`, `processingDuration: 15`, `presentationDelay: 29`.

**Actual (v6.2.0):**

```
metric.value        : 104
interactionTime     : 1100
nextPaintTime       : 1204
inputDelay          : -98      <-- negative
processingDuration  : 173      <-- inflated by exactly 98
presentationDelay   : 29
subparts sum        : 104
```

I also confirmed the premise in real Chrome (151), with a button whose `pointerover` handler blocks the main thread. Event Timing emits:

```
pointerover  start=16438.3  dur=224  pStart=16438.8  pEnd=16659.2  render=16662.3  iid=0
pointerdown  start=16438.3  dur=232  pStart=16659.5  pEnd=16668.6  render=16670.3  iid=1368
```

So `pointerover` really is exposed with `interactionId: 0`, really does carry a long duration, and really does fall inside the 8ms render window of the following `pointerdown`.

One caveat worth flagging if you try to automate this: in that capture the negative value did **not** appear, because WebDriver/CDP dispatches a whole click with a single hardware timestamp, so `pointerover` and `pointerdown` shared `startTime: 16438.3` and `inputDelay` came out at `+0.5`. The negative case needs the two events to have genuinely distinct hardware timestamps — i.e. a real user clicking part-way through the blocking handler — which synthesized input can't express. A unit-level test that feeds the entries directly (as above) is probably the more reliable regression test.

**Suggested fix**

Clamp the source value, so the subparts keep telescoping to `metric.value`:

```js
// `group.processingStart` is the earliest processing start across *all* events
// presented in this frame, which can predate this interaction's `startTime`
// (e.g. a long `pointerover` handler still running when the user clicked, or an
// earlier interaction in the same frame). Clamp so `inputDelay` is never negative.
const processingStart = Math.max(group.processingStart, firstEntry.startTime);
```

With that applied to a copy of the built module, the same run gives `inputDelay: 0`, `processingDuration: 75`, `presentationDelay: 29` — sum still `104`.

I also checked the case #748 was written to fix: two entries of the *same* interaction where the later one carries an earlier `processingStart`. Patched and unpatched are identical there (`inputDelay: 30`, `processingDuration: 55`, `presentationDelay: 25`), so the clamp is a no-op for it — every entry of a single interaction has `processingStart >= startTime`, so the `Math.max` only bites when a foreign entry pulled the group minimum below the interaction's own start.

Two approaches I'd avoid: clamping `attribution.inputDelay` after the fact breaks the `inputDelay + processingDuration + presentationDelay === metric.value` identity; and restricting `processingDuration` to only the interaction's own entries would undo the deliberate frame-wide processing attribution from #388 / #442.

Happy to open a PR with a regression test.

**Environment:** web-vitals 6.2.0 (present since 6.0.0), attribution build. Real-Chrome corroboration on Chrome 151.0.7922.138, macOS.

suhailopensource:
Fixes #788.

Clamps `group.processingStart` to the interaction's own `startTime` in `attributeINP()`, as discussed in the issue:

```js
const processingStart = Math.max(group.processingStart, firstEntry.startTime);
```

This keeps events with `interactionId === 0` counted in `processingDuration`, per the subpart definitions you set out — input delay is non-handler code, processing is handler code including zero-`interactionId` events. I verified that directly: with an interaction starting at 1000 and an `interactionId: 0` handler running 1050–1090, `processingStart` still resolves to 1050, so that handler's time lands in `processingDuration` and the entry stays in `processedEventEntries`. The clamp only discards zero-`interactionId` time that predates the interaction's own `startTime`, which was never part of its latency.

In the issue's repro this gives `inputDelay: 0`, `processingDuration: 75`, `presentationDelay: 29` (was `-98 / 173 / 29`).

Worth noting the clamp can only raise `processingStart` as far as `firstEntry.startTime`, which is `<= startTime + duration`, so `nextPaintTime` cannot move. `metric.value` and `presentationDelay` are therefore unchanged, and only `inputDelay` and `processingDuration` shift by equal and opposite amounts — the three subparts still sum to `metric.value`.

### first-input cleanup

Also drops the now-redundant `|| entry.entryType === 'first-input'` clause. I confirmed in Chrome 151 that `first-input` exposes a non-zero `interactionId` (measured `7282` and `4643` in separate runs), and that it is still delivered when `durationThreshold` would filter the corresponding `event` entry — so the fallback for fast-only interactions is preserved.

Two notes on this part:

- The same pattern appears in **two** places, `lib/InteractionManager.ts:118` and `attribution/onINP.ts:213`. I simplified both for consistency; happy to drop the second if you'd rather keep this focused.
- I could only verify the non-zero `interactionId` in Chrome. If Firefox or Safari still report `first-input` with `interactionId: 0`, the simplification would stop INP being reported at all on pages whose only interactions fall under `durationThreshold`. CI covers both browsers, so it should surface here, but flagging it since the failure mode is a missing metric rather than a wrong value.

### Tests

`test/unit/attribution-onINP-test.js` (new) covers the subparts:

- the negative-`inputDelay` scenario — fails on `main` with `expected: 0, actual: -98`, passes with the clamp
- the same-interaction out-of-order case from #748, unchanged at `30 / 55 / 25`

`test/unit/InteractionManager-test.js` gains two cases pinning the guard, since there was no existing coverage of the `first-input` path (`grep first-input test/e2e/onINP-test.js` returns nothing): a `first-input` entry with an `interactionId` is still treated as an INP candidate, and entries without one are skipped. Both pass under the old and new guard by design — they pin behaviour that must hold either way.

These are unit tests rather than e2e deliberately. The bug needs `pointerdown.startTime` to fall after another event's `processingStart`, and WebDriver dispatches a whole click with a single hardware timestamp — I tried, and the synthesized entries all share `startTime`, so an e2e test can't express it.

### One thing I have not measured

The `longestScript.subpart` boundaries reduce to `processingStart` and `processingEnd`, so the clamp shifts them: a script starting before the interaction now falls in the `input-delay` bucket rather than `processing-duration`, in a bucket whose reported width is 0. I derived that from reading lines 351-358 rather than measuring it, since it needs LoAF entries. Flagging rather than asserting, in case you want it handled differently.

### Verification

`build`, `lint` and `format:check` clean. `test:unit` 26/26. Full `--browsers=chrome` e2e: 5/5 spec files, 153 passing, 4 skipped. Reverting just the clamp leaves the new test failing and everything else green.


tunetheweb (approved):
LGTM - thanks!

